PR #42, opened 2023-09-02
Tim/pol 86 create webserver

tg339:


tg339:
Webserver has been created and can run when using the dev command. This webserver only runs hello world at this point. We need to add dynamic routing and a sync between the watcher and the webserver

callicles (approved):

PR #45, opened 2023-09-04
Tim/pol 99 create route from file

tg339:


apps/igloo-kit-cli/src/cli/watcher.rs
```diff
@@ -0,0 +1,80 @@
...
+use notify::{RecommendedWatcher, Config, RecursiveMode, Watcher, event::ModifyKind};
+
+use crate::{framework::directories::get_app_directory, cli::user_messages::show_message};
+
+use super::{CommandTerminal, user_messages::{MessageType, Message}};
+
+fn process_event(project_dir: PathBuf, event: notify::Event, route_table:  Arc<Mutex<HashSet<PathBuf>>>) {
+    match event.kind {
+        notify::EventKind::Create(_) => {
+            let route = event.paths[0].clone();
```
callicles, line 12: When can this be greater than length 1? Should we do something about it?

apps/igloo-kit-cli/src/cli/watcher.rs
```diff
@@ -0,0 +1,80 @@
...
+            let clean_route = route.strip_prefix(project_dir).unwrap().to_path_buf();
+            let mut route_table = route_table.lock().unwrap();
+            route_table.insert(clean_route);
+        },
+        notify::EventKind::Modify(mk) => {
+            match mk {
+                ModifyKind::Name(_) => {
+                    // If the file is renamed, remove the old route and add the new one
+                    let route = event.paths[0].clone();
```
callicles, line 21: Same here

apps/igloo-kit-cli/src/cli/webserver.rs
```diff
@@ -0,0 +1,68 @@
...
+use std::sync::Arc;
+use super::CommandTerminal;
+
+
+async fn handler(req: Request<Body>, route_table: Arc<Mutex<HashSet<PathBuf>>>) -> Result<Response<String>, Infallible> {
+    let route_prefix = PathBuf::from("/");
+    let route = PathBuf::from(req.uri().path()).strip_prefix(route_prefix).unwrap().to_path_buf();
+
+    // Check if route is in the route table
+    if route_table.lock().unwrap().contains(&route) {
```
callicles, line 24: You have to lock the implementation to read? Does it mean you can't write concurrently?

apps/igloo-kit-cli/src/cli/watcher.rs
tg339: I haven't encountered it yet and need to do more testing. I'll create a ticket for this one.

apps/igloo-kit-cli/src/cli/webserver.rs
tg339: With a mutex, that is correct. You have to lock to read. We could try a RWLock to see if we can get concurrent readers. I had concerns about the ability to pass the RWLock to the thread which is why I didn't use it. I'll add a comment to investigate the RWLock but we can probably differ that to future releases

PR #48, opened 2023-09-21
Tim/pol 92 create table with file name

tg339:
Massive lift here. We had to pivot from just storing JSON to creating tables with the right types. We're parsing prisma and creating all the underlying infra based on the data model that the user defines. 

Massive kudos to Nic for helping me find an efficient path forward 🎉


apps/igloo-kit-cli/src/cli/mod.rs
```diff
@@ -81,20 +81,33 @@ impl CommandTerminal {
 }
 
 async fn top_command_handler(term: &mut CommandTerminal, config: Config, commands: &Option<Commands>, debug: bool) {
-
-    if (!config.features.coming_soon_wall) {
+    let clickhouse_config = ClickhouseConfig {
```
callicles, line 84: Feels like we will need to eventually make those parameters?

apps/igloo-kit-cli/src/cli/mod.rs
```diff
@@ -81,20 +81,33 @@ impl CommandTerminal {
...
+            routines::start_containers(term, clickhouse_config.clone());
             infrastructure::setup::validate::validate_red_panda_cluster(term, debug);
-            routines::start_development_mode(term).await;      
+            routines::start_development_mode(term, clickhouse_config.clone()).await;      
         }
         Some(Commands::Update{}) => {
             // This command may not be needed if we have incredible automation
-            todo!("Will update the project's underlying infrascructure based on any added objects")
+            // todo!("Will update the project's underlying infrascructure based on any added objects")
+            let path = "/Users/timdelisle/Dev/igloo-stack/apps/igloo-kit-cli/tests/psl/simple.prisma";
```
callicles: Hum we should probably remove this now, no?

callicles (approved):
Great work on the clickhouse integration! Do we want to remove the hardcoded path now or later?

apps/igloo-kit-cli/src/cli/mod.rs
tg339: nah, bad form in rust and the linter was complaining

apps/igloo-kit-cli/src/cli/mod.rs
tg339: fixed

PR #49, opened 2023-09-26
Full demo flow achieved

tg339:


apps/igloo-kit-cli/src/cli/mod.rs
```diff
@@ -92,6 +92,12 @@ async fn top_command_handler(term: &mut CommandTerminal, config: Config, command
         cluster_network: PANDA_NETWORK.to_owned(),
     };
 
+    let redpanda_config: RedpandaConfig = RedpandaConfig {
+        broker: "localhost:19092",
```
callicles, line 96: We should make this configurable as well in a follow up PR.

We could use the config file I set up

apps/igloo-kit-cli/src/cli/webserver.rs
```diff
@@ -31,11 +38,30 @@ async fn handler(req: Request<Body>, route_table: Arc<Mutex<HashMap::<PathBuf, R
...
-                let body = String::from_utf8(bytes.to_vec()).unwrap();       
+                let body = String::from_utf8(bytes.to_vec()).unwrap();
 
-                return Ok(Response::builder()
-                    .status(StatusCode::FOUND)
-                    .body(body)?)
+                let guard = route_table.lock().await;
+                let topic_name = &guard.get(&route).unwrap().table_name;
+
+                let res = configured_producer.lock().await.producer.send(
```
callicles, line 46: Damn that's a lot of locks. I get why it needs to be this way for dev. For prod we should assume the routing is immutable. Not sure if that changes how you would write the code

callicles (approved):

apps/igloo-kit-cli/src/cli/webserver.rs
tg339: Good call. We should have two paths here. One for dev mode and one for prod.

PR #59, opened 2023-10-07
Rfd/0002 - Ecosystem Skeleton Architecture

callicles:
First RFD about the system as a whole.
This describes what aim to cover and how we are thinking about it.

### DIFF
--- a/rfd/0002/README.mdx
+++ b/rfd/0002/README.mdx
@@ -1,7 +1,24 @@
 —--
 authors: Nicolas Joseph <nicolas@nicolasjoseph.com>
-title: Framework Skeleton Architecture
+title: Ecosystem Skeleton Architecture
 state: prediscussion
 —--
 
-# Framework Skeleton Architecture
+## Framework Problem Statement:
+
+We are building an opinionated framework to enabled engineers that are building data infrastructure and pipeline to focus on the business logic and not the infrastructure. We aim at providing them with the same user experience as a front end engineer building a web application. We want to provide a framework that is easy to use, easy to maintain, easy to extend and most importantly, gives the user feedback on a code iteration immediately.
+
+## Principles
+
+- Local first: We optimize for the user experience first and we will figure out how to deploy to production later.
+- Easy to use: We want to provide a framework that is easy to use, easy to maintain, easy to extend and most importantly, gives the user feedback on a code iteration immediately.
+- Easy to test: We want to provide a framework that has built in unit testing from the get go.
+- Performance: We want to provide a framework that is fast and can scale to large data sets for real life use cases.
+- Extensible: We want to provide a framework that is easy to extend and that can be used to build a wide range of data applications.
+
+## Architecture
+
+### Overview
+
+- Data Ingest SDK: The data ingest SDK is a simple library that can be embedded in the user's application to send data to an igloo deployment. It is typed with the User's Business Objects to ensure that the data sent to igloo is valid.
+- Igloo Server:
--- a/rfd/0002/README.mdx
+++ b/rfd/0002/README.mdx
@@ -1,8 +1,8 @@
-—--
+---
 authors: Nicolas Joseph <nicolas@nicolasjoseph.com>
 title: Ecosystem Skeleton Architecture
 state: discussion
-—--
+---
 
 ## Framework Problem Statement:
 
--- a/rfd/0002/README.mdx
+++ b/rfd/0002/README.mdx
@@ -1,7 +1,7 @@
 —--
 authors: Nicolas Joseph <nicolas@nicolasjoseph.com>
 title: Ecosystem Skeleton Architecture
-state: prediscussion
+state: discussion
 —--
 
 ## Framework Problem Statement:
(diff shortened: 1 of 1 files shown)


rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+- **Easy to use**: We want to provide a framework that is easy to use, easy to maintain, easy to extend and most importantly, gives the user feedback on a code iteration immediately.
+- **Easy to test**: We want to provide a framework that has built in unit testing from the get go.
+- **Performance**: We want to provide a framework that is fast and can scale to large data sets and volumes for real life use cases.
+- **Extensible**: We want to provide a framework that is easy to extend and that can be used to build a wide range of data applications.
+
+## Architecture
+
+### Overview
+
+- Data Ingest SDK: The data ingest SDK is a simple library that can be embedded in the user's application to send data to an igloo deployment. It is typed with the User's Business Objects to ensure that the data sent to igloo is valid.
```
mcginleyr1, line 24: Are you thinking at all about extraction? The real HOW to get data out of like shitforce and start converting it to streaming.

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+- Igloo Server: This is a component that abstracts away several parts of the architecture. It contains a web server that will ingest the data, a streaming engine that will store the data in streams and an OLAP Database that will store the data in a columnar format. It also runs all the code that is supplied by the user to transform the data.
```
mcginleyr1: Have you all thought about storage layers? Ie cold storage for data I need to keep for legal reasons but that I'm pretty much never going to access.

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+- Igloo CLI: This is a command line interface that will allow the user to interact with the igloo server. It will allow the user to create a new project, run the project locally and deploy the project to a cloud provider.
+- Data Consumption SDK: This SDK is typed with the user's data and allows the user to query the data from the Igloo Server. It will also allow the user to subscribe to data changes and receive updates in real time.
```
mcginleyr1, line 27: Are you going to include governance in the consumption and or server?

rfd/0002/README.mdx
callicles: That's a good callout. No, I don't think we have been thinking too much about that.

Knee jerk reaction is: 2 use cases for extraction
* pull based: we could run a function the user would implement on a cron. The function would use the before mentioned SDK to produce the events that are ingested, reading the data from the target platform apis.
* push based: the platform to extract data from has a webhook feature, and we allow that to point to igloo server for ingest - not sure how to think about types in this case.

rfd/0002/README.mdx
callicles: Yeah - Clickhouse has the ability to run on multiple storage tiers. We were thinking of leveraging that.

Ref: https://clickhouse.com/docs/en/integrations/s3#storage-tiers

rfd/0002/README.mdx
callicles: 🤔 another good callout, I have not considered governance yet, and how it lays out. 

Since we can't trust the client I would say both. 
* The Security / Access aspect needs to be handled on the server side. 
* The SDK / Client should be type safe from a quality of life perspective

In terms of data quality, I am thinking we could check the data at ingestion against the data schema / contract.

rfd/0002/README.mdx
mcginleyr1: Yea, nexla/fivetran do it where they use REST as much as they can and read records then shit them over into events/single records. It gets slow as heck though when you want to process 2m records a day from say gcs with lots of small files...  Bunch to think through in just the data acquisition part, especially since the majority of decisions folks would want to make around data and do ml with lay outside their product engineering orgs.

tg339: Can be an interesting area to think about partnering with folks as well as opposed to building. We might want to look into CDC providers like debezium for the DB related extractions

mcginleyr1: @tg339 you're also just thinking about DB's which is significantly easier then business applications which want to horde data.  That's where you can really get analytics unlocks.  I think all the normal B2C, and general martech use cases are super easy compared to end to end business analytics.

tg339: @mcginleyr1 good point. We need to do some more digging here. Feels like this should be solvable with reasonable lift. If we need to go down the full blown connector path, that's where the lift might become to heavy to justify the use cases

mcginleyr1: It might be easier to put together toolkits so folks can build their own connectors really easily. since all they have to do is read some crappy endpoints or connections and spew the data into your server.

callicles: Yeah - I will put together another rfd to look at what this could look like. I was looking at airbyte and how they do connectors, we might be able to do a destination into igloo and leverage their ecosystem of connectors. As well as provide low level sdk if someone wanted to build the integration directly.

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
+—--
```
francescoperera: seems like the formatting is off a bit . Here is what I see when I look at the file
**
image
**

Aside from the red highlighting, its harder to interpret the raw mermaid diagram

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+state: discussion
+—--
+
+## Framework Problem Statement:
+
+We are building an opinionated framework to enabled engineers that are building data infrastructure and pipeline to focus on the business logic and not the infrastructure. We aim at providing them with the same user experience as a front end engineer building a web application. We want to provide a framework that is easy to use, easy to maintain, easy to extend and most importantly, gives the user feedback on a code iteration immediately.
+
+## Principles
+
+- **Local first**: We optimize for the user experience first and we will figure out how to deploy to production later.
```
francescoperera, line 13: I am assuming this principle is directly tied to enabling users to have smaller and quicker feedback loops ?

Typically, in data engineering aspects your code iteration is often tied to access to your data storage systems ( both source and destination). In some instances ( i.e db only accessible through remote servers on  the internal network), those systems might not be reachable in a local environment.

francescoperera: Also in my experience, "local first" tools   reduce the barrier to entry, make it easier learn the tool but don't necessarily enable you to always have fast feedback loops.

Spark and Airflow can both be run locally yet you still see pretty large feedback loops.

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+
+- Data Ingest SDK: The data ingest SDK is a simple library that can be embedded in the user's application to send data to an igloo deployment. It is typed with the User's Business Objects to ensure that the data sent to igloo is valid.
+- Igloo Server: This is a component that abstracts away several parts of the architecture. It contains a web server that will ingest the data, a streaming engine that will store the data in streams and an OLAP Database that will store the data in a columnar format. It also runs all the code that is supplied by the user to transform the data.
+- Igloo CLI: This is a command line interface that will allow the user to interact with the igloo server. It will allow the user to create a new project, run the project locally and deploy the project to a cloud provider.
+- Data Consumption SDK: This SDK is typed with the user's data and allows the user to query the data from the Igloo Server. It will also allow the user to subscribe to data changes and receive updates in real time.
+- Data Visualization SDK: This SDK plays well with the data Consumption SDK and contains web components of common data visualizations that can be used to build a dashboard.
```
francescoperera: where would these dashboards be hosted ?

rfd/0002/README.mdx
francescoperera, line 27: will the 3 sdks come as 3 different dependencies/installs or all bundled together ?

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+    HTTP --Queries-->View
+  end
+  subgraph User Apps
+    sdk1[Data Ingest SDK]--Pushes-->HTTP
+    sdk2[Data Consumption SDK]--Queries--> HTTP
+    sdk3[Data Vizualization SDK]--Uses-->sdk2
+  end
+```
+
+### HTTP Server
```
francescoperera, line 55: what is the runtime ? will it be packaged as a container

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+
+The HTTP Server is a high performance API that has 2 responsabilities:
+
+- Proxy any event it receives to the proper topic in the streaming engine
+- Enable an SDK to proxy queries to any of the materielized views in the OLAP Storage layer.
+
+In Local mode it can be the same process serving both, in a production setting, we could separate it in 2 components that are horizontally scalable independantly.
+
+### Lambda Ingestion Function
```
francescoperera, line 64: does this mean that igloo would require AWS  ? 

If so, how would this play in a local environment ?

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+
+This is a lambda function is that is provided as a hook for the user to be able to add arbitrary code to transform events that are coming from the ingest layer.
+
+The input of the function inputs and outputs will be typed so that the user can properly encode the mapping.
+
+The lambda function writes back to the streaming engine. To Increase throughput the instances of this function can be arbitrarily scaled horizontally.
+
+Random idea: we could have those be state machines. In state machine mode we would not have concurrency enabled. in standard mode they could be scaled horizontally.
```
francescoperera: What benefit are you seeing from having state machines ?

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+### CLI
+
+The igloo CLI is a tool to help the user get setup really quickly. It helps with
+
+- Initial project setup
+- Runs and configures locally the OLAP DB and the streaming engine
+- Orchestrate the ingest functions and runs them
+- Orchestrate the Table Writer
+- Runs the API server locally
+- Runs the developper console for the project
```
francescoperera: *developer

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+  style Table fill:#E7D182,color:#000
+  style OLAPI fill:#E7D182,color:#000
+  style View fill:#E7D182,color:#000
+  style HTTP fill:#E7D182,color:#000
+```
+
+### Open Questions
+
+- In Production how to we deploy everything?
+- Do we provide configurations for the OLAP cluster?
```
francescoperera, line 147: what do you mean with providing configurations ?  Enabling users to specify all the different configurable inputs to the cluster? Offer pre-baked configs ( i.e small vs medium vs large) ?

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+- Do we provide configurations for the streaming cluster?
+- Do we point the user to cloud offerings for storage to be plugged in?
```
francescoperera, line 149: is this a do or how question ? 
Existing data storage solutions are valued and expensive assets.  Asking users, especially in enterprises, to start using another system might be challenging.

Also, enabling it be plugged in fits in well with the extensibility principle

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+- How scalable do we need the first version of this to be?
+- Should we just bundle this all in one docker container?
+
+# Appendix
+
+## Technologies we chose
+
+### Rust
+
+As a high performant language we chose rust to be the glue we write that can be common across multiple user environments. Whether the user is leveraging a higher level language like Typescript or Javascript.
```
francescoperera: What does the "glue" refer to if we look at the components( http server, table writer etc...)  you listed above ?

francescoperera: what was the reason behind choosing Rust ?

adding the why behind this choice of using Rust  could help document valuable context.

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+
+It is also used to build the CLI which is cross compiled for linux and macOS
+
+### Clickhouse
+
+Clickhouse is a high performance data storage with a small footprint. Ideal for both being deployed on the developper machine when he or she is iterating and deployed in the cloud for massive scale
```
francescoperera: *developer

rfd/0002/README.mdx
```diff
@@ -0,0 +1,198 @@
...
+
+### Redpandas
+
+Red pandas is an implementation of Kafka written in C/C++. Highly performant and having a much simple operating cost than a full Kafka cluster with a Zookeeper.
+
+### Prisma
+
+We are using prisma files to describe tables and provision them in the olap system
```
francescoperera: just like with rust, documenting the reason behind this choice could be extremely valuable for future discussions.

rfd/0002/README.mdx
francescoperera, line 24: would the sdks be the only way to communicate with the platform or will users be able to interact through APIs?

SDKs definitely come with their benefit( retries, types etc..) , but need to be fleshed out(i.e user flows and actions are well defined)  to start delivering  good value.

rfd/0002/README.mdx
callicles: Yeah, we need to solve the data problem as well, along with the environment.

rfd/0002/README.mdx
callicles: The goal would be to host it in a "traditional" web app. ie could be a nextJs or some other react app

rfd/0002/README.mdx
callicles: Quick reaction is we would have them independent of one another. The optional one being the dashboarding one.

rfd/0002/README.mdx
callicles: Right now we have the HTTP server runtime as a rust HTTP server.

We are envisioning the open source to have a docker image that could be deployed to production with the stateless parts of the framework embedded.

rfd/0002/README.mdx
callicles: Hum, no - we are thinking about this as a "generic" function. Like a mapping function that the framework will call whether the framework runs in the cloud or locally.

rfd/0002/README.mdx
callicles: Doing things like windowed computation

rfd/0002/README.mdx
callicles: I am thinking people should be able to interact with the APIs directly

rfd/0002/README.mdx
callicles: I will add

rfd/0002/README.mdx
callicles: Fixed

rfd/0002/README.mdx
callicles: Governance probably requires its own Discussion. I added it to the open questions

rfd/0002/README.mdx
callicles: I added it in the open questions

rfd/0002/README.mdx
callicles: Added a comment to that effect

rfd/0002/README.mdx
callicles: Added a comment to that effect

rfd/0002/README.mdx
callicles: I meant production configuration. Like configure the docker image with the proper number of open file allowance, the number of allowed open connection, the k8s manifest that is setup properly for high availability. 

There would be a lot of work there

rfd/0002/README.mdx
callicles: Hum yeah there is multiple layers to this:
* The type of connection we propose, ie are we only compatible with clubhouse or can we be compatible with Databricks as well?
* Then where does this live, if we do only clubhouse and people have it already deployed, we should allow them to use that deployment

PR #65, opened 2023-10-25
Tim/pol 89 create view automatically from table creation

tg339:


packages/igloo-client/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
```
callicles: Why not reuse the base ts config from the package?

PR #66, opened 2023-10-31
initial templating working

tg339:


callicles:
Can you rebase on `main` you have some conflicts

packages/igloo-kit/src/client.ts
```diff
@@ -0,0 +1,3 @@
+interface Client {
```
callicles, line 1: Do you still need this package then?

callicles (approved):
Nice work! Can't wait to give it a spin :)

packages/igloo-kit/src/client.ts
tg339: Not for now. This thing is just a placeholder.

PR #85, opened 2023-11-16
Revised Igloo folders to match 514

asl677:
- removed extra folders and scripts as mentioned on call


asl677:
Ah seems removing those folders borked it. Will probably need help getting this part right.

callicles (approved):

PR #87, opened 2023-11-23
feat: add internal logger to help us debug

callicles:
Adds a logger for us to be able to track all the actions taken during a user's session and help debug if something goes wrong

### DIFF
--- a/apps/igloo-kit-cli/src/cli.rs
+++ b/apps/igloo-kit-cli/src/cli.rs
@@ -8,7 +8,7 @@ mod watcher;
 
 use std::sync::{Arc, RwLock};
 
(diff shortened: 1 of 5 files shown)


apps/igloo-kit-cli/src/cli.rs
tg339, line 1: We may want to remove references to igloo in the methods since we know we’ll likely need to change the name

tg339 (approved):
LGTM. You may want to remove explicit igloo references in case of name change

apps/igloo-kit-cli/src/cli.rs
callicles: Good point

PR #95, opened 2023-12-07
Tim/514 193 load directory state on dev startup

tg339:


apps/igloo-kit-cli/src/cli/local_webserver.rs
```diff
@@ -31,12 +33,22 @@ use std::sync::RwLock;
...
     pub port: u16,
 }
 
+impl LocalWebserverConfig {
+    pub fn new(host: String, port: u16) -> Self {
+        Self { host, port }
+    }
+
+    pub fn url(&self) -> String {
+        format!("http://{}:{}", self.host, self.port)
```
callicles, line 48: I love this - nice simple abstraction

apps/igloo-kit-cli/src/framework/controller.rs
```diff
@@ -0,0 +1,245 @@
...
+    let view = ClickhouseView::new(fo.table.db_name.clone(), view_name, fo.table.clone());
+    let create_view_query = view.create_materialized_view_query().map_err(|e| {
+        Error::new(
+            ErrorKind::Other,
+            format!("Failed to get clickhouse query: {:?}", e),
+        )
+    })?;
+
+    // Clickhouse doesn't support dropping a view if it doesn't exist so we need to drop it first in case the schema has changed
+    let drop_view_query = view.drop_materialized_view_query().map_err(|e| {
```
callicles, line 90: I don't get the algorithm probably something I am missing.

- Creating table
- Dropping the table 
- Creating the same table again

Why the 2 first steps? Or is I that if the table already exists it not going to do anything?

callicles (commented):
Small comment on the replace algorithms otherwise looks good.

apps/igloo-kit-cli/src/framework/controller.rs
tg339: Not sure if I'm seeing what you're seeing. The algorithm is:

1. Drop table if it exists
2. Create table if it does not exist

This should create the table only if it does not already exist and replace it if it does. 

Note that in the line you're highlighting, we're only creating the query, which we then run. Maybe you're seeing the query creation and running as the separate steps?

callicles: Can you point me to the `if exists`  logic? I guess that's what I am missing.

tg339: Line 103 in /infrastructure/olap/clickhouse/queries.rs

callicles: Ahh - I just realized one statement is to create the query while the other executes it.  🤦🏻 .

Thank you for bearing with me

callicles (approved):

PR #98, opened 2023-12-16
chore: claim moose namespaces

callicles:
* adds `moose` as a package name we claim for the CLI on npm
* adds `create-moose-app` as a package for the shortcut to init
* Updates workflows to automatically publish changes to moose alongside igloo for now


tg339 (approved):
LGTM

PR #103, opened 2023-12-21
created the standard for logging

tg339:
adding the standard for logging practices to the logger module in the form of module documentation.

### DIFF
--- a/apps/igloo-kit-cli/src/cli/logger.rs
+++ b/apps/igloo-kit-cli/src/cli/logger.rs
@@ -1,3 +1,39 @@
+//! # Logger Module
+//!
+//! This module provides logging functionality for the application.
+//!
+//! ## Components
+//!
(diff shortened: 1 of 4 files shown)


tg339:
@callicles no idea why we're getting the lockfile drift here. Tried the usual fix and it didn't fix the vercel build. If you can help look into this that would be awesome!

callicles:
If you rebase on `main` - you should be able to just accept the version from `main` for the pnpm file since you did not touch any of the `package.json` files

apps/igloo-kit-cli/src/cli/logger.rs
```diff
@@ -1,3 +1,39 @@
+//! # Logger Module
```
callicles, line 1: That's interesting - I would have put this documentation on the documentation website in the section for contributors.

Since we don't have that up - we can leave it here for now and revisit later.

callicles (approved):
Just needs a rebase from main and can be merged

PR #109, opened 2023-12-25
Tim/514 212 turn the display module show message into a macro

tg339:
further simplification of the business logic by turning show_message into a macro


callicles (approved):
LGTM

PR #111, opened 2023-12-30
Tim/514-228-fix-webserver-start-issue-when-starting-dev-mode-from

tg339:
fixing a state initialization bug when starting dev mode


apps/igloo-kit-cli/src/utilities/docker.rs
```diff
@@ -2,6 +2,66 @@ use std::{path::PathBuf, process::Command};
...
+    pub running_for: String,
+    #[serde(rename = "Size")]
+    pub size: String,
+    #[serde(rename = "State")]
+    pub state: String,
+    #[serde(rename = "Status")]
+    pub status: String,
+}
+
+pub fn list_containers() -> std::io::Result<Vec<ContainerRow>> {
```
tg339, line 40: how do I run just this function?

callicles (approved):
LGTM

PR #116, opened 2024-01-04
ensure docker running before running commands

phiSgr:
The failure message is this:

`Failed to run docker commands. Is docker running?`


tg339:
can we create a cleaner error message with show_message

`Failed to run docker commands. Is docker running?`

apps/igloo-kit-cli/src/cli/routines.rs
```diff
@@ -126,10 +127,10 @@ impl RoutineSuccess {
 pub struct RoutineFailure {
     message: Message,
     message_type: MessageType,
-    error: Error,
+    error: Option<Error>,
 }
 impl RoutineFailure {
-    pub fn new(message: Message, error: Error) -> Self {
+    pub fn new(message: Message, error: Option<Error>) -> Self {
```
phiSgr: This caused a big diff. Maybe I should create another function to create a `RoutineFailure` without an `Error`.

tg339: What we're essentially saying here is that there are cases where a routine fails to generate the expected result but didn't error out. Right?

Maybe we should rethink the return type of the routines? What comes to mind is  a result with three cases Success, Failure, Error

tg339: That wouldn't solve the diff issue thought 😛

tg339: I'm a bit curious as to what the use case for a failure that isn't an error. Maybe we need to rethink that?

tg339 (approved):

apps/igloo-kit-cli/src/cli/routines.rs
phiSgr: > a failure that isn't an error

to be precise, a failure that isn't **caused by** an **IO** error

PR #117, opened 2024-01-04
rename dataframes to datamodels

phiSgr:


apps/igloo-kit-cli/src/framework.rs
```diff
@@ -15,15 +15,15 @@ pub enum Insights {
 pub enum TopLevelObjects {
     Ingestion,
     Flow,
-    Dataframe,
+    DataModel,
```
phiSgr, line 18: The dictionary is happy with `Dataframe` as one single word, but not `Datamodel`. I don't feel strongly either way.

tg339: I don't feel strrongly either way either. I think this is fine.

tg339 (approved):
LGTM

PR #120, opened 2024-01-05
Tim/514 218 build clear on brand developer users of moosejs

tg339:


apps/framework-docs/.gitignore
```diff
@@ -35,3 +35,6 @@ yarn-error.log*
 
 # Sentry Auth Token
 .sentryclirc
+
+# Sentry Config File
+.sentryclirc
```
callicles: duplicate entry?

apps/framework-docs/src/pages/building.mdx
```diff
@@ -0,0 +1,29 @@
+import { Cards, Card } from 'nextra/components'
+
+# Building a Data Service
+MooseJS helps you build data intensive applications and service. In its current state, MooseJS can't be used to build a full fledged web application. However, it can be used to build a data service which you can use to power adhoc analytics use cases.
```
callicles: I am not sure we should talk about what it can't do yet. ie I would remove: `MooseJS can't be used to build a full fledged web application` It would make it more positive.

apps/framework-docs/src/pages/building.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+## What is a data service?
+With the MooseJS's current functionality a data service can be thought of as a microservice that's used to capture, store and serve data. It interfaces with best in-class data infrastructure and allows you to build a data warehouse wihtout any data engineering skills. A service is composed of:
+
+- A series of data models which define the structure of the data you want to caputre
+- An ingestion API to capture data from other applications or services
+- A storage layer that queues up data on ingestion and stores it for later processing
+
+We recommend that you use all the standard software development practices to build your service.
```
callicles: Hum - reading this as a software engineer, I could be offended - of course I want to use best practices. I would remove this sentence.

Instead, I would flip it and say something along the lines of:

`Thanks to 'moose', you get to use the same best practices you use for your application layer in your analytics and data warehousing layer.`

apps/framework-docs/src/pages/building.mdx
```diff
@@ -0,0 +1,29 @@
...
+- A storage layer that queues up data on ingestion and stores it for later processing
+
+We recommend that you use all the standard software development practices to build your service. 
+
+**We'd recommend at least the following:**
+- Using version control. We like managing our data services along side our main user applications in a mono-repo
+- Using a CI/CD pipeline to deploy your service to production ensuring that tests are run before deployment
+
+## What can data services be used for?
+Data services can be used to capture data from your user applications and store it in a data warehouse. That means you can replace dedicated data capture tooling such as Segment or Mixpanel with your own data service and generic BI solution. This greatly reduces the cost of running your data infrastructure and gives you more control over your data.
```
callicles, line 20: I would add all the use cases that people can build on:
* telemetry application
* Transaction Statistics
* Dasboards
etc...

apps/framework-docs/src/pages/building/capture-data.mdx
```diff
@@ -0,0 +1,55 @@
+import { Callout } from 'nextra/components'
+import { Tabs } from 'nextra/components'
+
+# Capture data
+MooseJS makes it easy to capture data from your user facing applications by providing typed SDKs which are automatically generated from your data models. 
+
+## From MooseJS SDKs
+<Callout type="info" emoji="ℹ️">
+  We currently support Typescript SDKs. Support for other languages is coming soon.
```
callicles, line 9: Would be great to have a link to a feature tracker here so that people can upvote a language they would want to see.

apps/framework-docs/src/pages/building/capture-data.mdx
```diff
@@ -0,0 +1,55 @@
...
+</Callout>
+
+You can develop your data service right along side your user facing applications. MooseJS will automatically generate a Typescript SDK as you build your service locally and make them available for you to link to your application. To link a data service's SDK to your application, simply run the following command:
+
```
callicles, line 13: That's after you have run `moose dev` right?

apps/framework-docs/src/pages/building/capture-data.mdx
```diff
@@ -0,0 +1,55 @@
...
+```txt filename="Terminal"
+// From your application's root directory or wherever you run npm install
+npm link <project-name>-sdk 
+```
+
+## Through Ingestion Points
```
callicles, line 19: For this example we should have the `prisma` file next to it that is used to validate and ingest the data

apps/framework-docs/src/pages/building/make-changes.mdx
```diff
@@ -0,0 +1,20 @@
+# Changing your data models
+Change management in your data stack is critical to ensure that your data is always accurate and up to date based on your end user's expectations. Broken dashboards, reports, or machine learning models can to unhappy customers and lost revenue. MooseJS makes it easy to change your data models and ensure that your data stack is always up to date. To do so, we leverage git workflows to ensure that changes are tracked and can be rolled back if necessary.
+
+## Adding a new data model
+Adding a new data source is as simple as adding a new model to an existing or new file in the `data models` directory. MooseJS will automatically detect the new model and add it to the data stack. 
+
+## Changing an existing data model
+
+### During development
+Whenever you save a file, Moose will automatically detect the change and update your data stack. Under the hood, MooseJS manages the state of your stack and creates temporary tables each time you save your file. This allows you to test your changes without worrying the effects on the existing data.
```
callicles:
```suggestion
Whenever you save a file, Moose will automatically detect the change and update your data stack locally. Under the hood, MooseJS manages the state of your stack and creates temporary tables each time you save your file. This allows you to test your changes without worrying the effects on the existing data. 
```

apps/framework-docs/src/pages/building/make-changes.mdx
```diff
@@ -0,0 +1,20 @@
...
+
+## Adding a new data model
+Adding a new data source is as simple as adding a new model to an existing or new file in the `data models` directory. MooseJS will automatically detect the new model and add it to the data stack. 
+
+## Changing an existing data model
+
+### During development
+Whenever you save a file, Moose will automatically detect the change and update your data stack. Under the hood, MooseJS manages the state of your stack and creates temporary tables each time you save your file. This allows you to test your changes without worrying the effects on the existing data. 
+
+Once you commit your changes, MooseJS will automatically create a new version of your data stack and turn on the migration mechanisms that ensure that historical data is present in your modified schema. The migration mechanisms assume that the initial state of your data starts at the initial state of your branch. This limits the number of active migrations that are running in your stack to optimize performance and costs.
```
callicles, line 12: Not sure what you mean by that. Why isn't the migration base state the previous commit?

The advantage of using a commit is that it is the same simple predictable process on the `main` branch or on feature branches

apps/framework-docs/src/pages/building/make-changes.mdx
```diff
@@ -0,0 +1,20 @@
...
+
+If you already have a version of the SDK instrumented, that version will remain live and be marked for deprecation. This allows you to test your changes in without needing to update all your exiting instrumentation.
```
callicles:
```suggestion
If you are already using a version of the sdk, that version will remain live and be marked for deprecation. This allows you to test your changes in without needing to update all your exiting instrumentation.
```

apps/framework-docs/src/pages/building/make-changes.mdx
```diff
@@ -0,0 +1,20 @@
...
+
+### In production
+Once you're ready to deploy your changes to production, you can merge your branch into the `main` branch. MooseJS will automatically detect the merge and create a new version of your data stack with all the migrations taken care of behind the scenes while ensuring that your SDKs historical SDKs continue to work.
+
+## Deleting a data model
+Deleting a data model is as simple as deleting the model from the file. MooseJS will automatically flag the model for deprecation and create a new version of your data stack. This allows you to remove data models that are no longer needed without worrying about breaking systems that depend on your data service. Once no more data is being pushed to your service or consumed, Moose will automatically clean up your infrastructure.
```
callicles, line 20: I am not sure this should be the behavior.

If I understand correctly, In this case we will end up with the code in the repo that doesn't reflect the production environment, where, for some time, the infra will have data flow through with no code or model in the code.

We could recommend here the strangler pattern in which people would remove the model once nothing is consuming the data and then delete the model in the code. When the model is deleted, then the infrastructure would be deprovioned immediately

apps/framework-docs/src/pages/getting-started/installation.mdx
```diff
@@ -0,0 +1,23 @@
...
+npx create-moosejs-app my-app
+```
+
+---
+## Manual Installation
+
+If you'd like more flexibility, you can also also install MooseJS globally using NPM. This will allow you to run the MooseJS CLI from anywhere on your machine and initialize new projects.
+
+```txt filename="Terminal" copy
+npm install -g @514-labs/moosejs
```
callicles:
```suggestion
npm install -g @514-labs/moose-cli
```

callicles (commented):
Great work putting this together, we desperately needed it.

Small comments on the prospective features and something that may be missing is a reference for the config files in both `~/.moose` and `project.toml`

apps/framework-docs/src/pages/building/make-changes.mdx
tg339: Because you'd need a new infra + migration mechanism for each step no?

apps/framework-docs/src/pages/building/make-changes.mdx
tg339: Ok, let's talk more about this and fold it into the PRD. I'm a bit worried about hard deletes here and the potential implications.

apps/framework-docs/src/pages/building/make-changes.mdx
callicles: Yeah - as long as we have a way to prevent it from getting too large on disk that should be ok no?

apps/framework-docs/src/pages/building/make-changes.mdx
callicles: That's a fair worry. If we go with the git based approach though - people will need to do reviews in order to merge a deletion - as such that will be reviewed before it gets executed in prod

Does that mitigate the worry?

apps/framework-docs/src/pages/building/capture-data.mdx
callicles: or maybe the slack so that they join to suggest changes

PR #135, opened 2024-01-18
Small fixes

asl677:
- Smaller weight images
- Correct footer logos


apps/web/src/app/components/LogoComponent.tsx
```diff
@@ -42,7 +42,7 @@ export const LogoComponent = () => {
           <Image 
             // className="invisible"
             ref={imgageRef}
-            src="/logo-moose-black.svg"
```
phiSgr, line 45: Do we want to delete the old svg?

asl677: We can do that.

tg339:
@asl677 you need check in a new commit to ensure that the vercel build passes

tg339 (approved):

PR #198, opened 2024-01-23
simplify RouteService

phiSgr:
`FutureProducer` in `ConfiguredProducer` "can be cheaply cloned" - it already has an `Arc` for its inner producer - so we do not need to wrap `Arc<Mutex<` around it.
`route_table` lasts as long as the application, so it can have a `'static` lifetime and simplify the type.


callicles (approved):
Love the simplification

PR #302, opened 2024-01-30
init command generates a node project

phiSgr:
and misc changes.

### DIFF
--- a/apps/igloo-kit-cli/src/cli/routines/start.rs
+++ b/apps/igloo-kit-cli/src/cli/routines/start.rs
@@ -8,9 +8,9 @@ use super::{
     Routine, RoutineFailure, RoutineSuccess, RunMode,
 };
 use crate::cli::display::with_spinner;
-use crate::cli::routines::initialize::CreateIglooTempDirectoryTree;
+use crate::cli::routines::initialize::{CreateDockerNetwork, CreateIglooTempDirectoryTree};
 use crate::cli::routines::util::ensure_docker_running;
-use crate::utilities::constants::CLI_PROJECT_INTERNAL_DIR;
+use crate::utilities::constants::{CLI_PROJECT_INTERNAL_DIR, PANDA_NETWORK};
 use crate::{
     cli::display::Message,
     project::Project,
@@ -48,6 +48,7 @@ impl Routine for RunLocalInfrastructure {
         CreateIglooTempDirectoryTree::new(RunMode::Explicit {}, self.project.clone())
             .run_explicit()?;
         ValidateMountVolumes::new(igloo_dir).run_explicit()?;
+        CreateDockerNetwork::new(PANDA_NETWORK).run_explicit()?;
         ValidatePandaHouseNetwork::new().run_explicit()?;
         RunRedPandaContainer::new(self.project.clone()).run_explicit()?;
         ValidateRedPandaRun::new().run_explicit()?;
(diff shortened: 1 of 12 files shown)


apps/igloo-kit-cli/src/cli/routines/start.rs
```diff
@@ -48,6 +48,7 @@ impl Routine for RunLocalInfrastructure {
         CreateIglooTempDirectoryTree::new(RunMode::Explicit {}, self.project.clone())
             .run_explicit()?;
         ValidateMountVolumes::new(igloo_dir).run_explicit()?;
+        CreateDockerNetwork::new(PANDA_NETWORK).run_explicit()?;
```
phiSgr, line 51: Reduce the need to rerun `init`

apps/igloo-kit-cli/src/infrastructure/olap/clickhouse.rs
```diff
@@ -237,7 +237,7 @@ pub async fn check_ready(
     crate::utilities::retry::retry(
         || run_query(&dummy_query, configured_client),
         |i, e| {
-            i < 5
+            i < 10
```
phiSgr, line 240: Older machines need more time.

apps/igloo-kit-cli/src/project.rs
```diff
@@ -20,17 +21,21 @@ use crate::infrastructure::olap::clickhouse::config::ClickhouseConfig;
...
 use log::debug;
 use serde::{Deserialize, Serialize};
 use std::path::Path;
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
-struct ProjectConfigFile {
+#[serde(rename_all = "camelCase")]
+struct PackageJsonFile {
     pub name: String,
-    pub language: SupportedLanguages,
```
phiSgr, line 33: Probably always ts when it's `package.json`

apps/igloo-kit-cli/src/project.rs
```diff
@@ -165,14 +170,18 @@ impl Project {
...
     pub fn write_to_file(&self) -> Result<(), std::io::Error> {
-        let config_file = ProjectConfigFile {
+        let config_file = PackageJsonFile {
             name: self.name.clone(),
-            language: self.language,
+            version: "0.0".to_string(),
+            scripts: HashMap::from([("dev".to_string(), "igloo-cli dev".to_string())]),
+            dependencies: HashMap::new(),
+            dev_dependencies: HashMap::from([(
+                "@514labs/moose-cli".to_string(),
```
phiSgr, line 183: The executable in `@514labs/moose-cli` is still called `igloo-cli`.

callicles:
Hey! Do we want to actually replace the project.toml or just add a package.json? In nextjs there is a configuration file for the project called next.config.js we probably need something similar for project specific config no?

phiSgr:
For now we don't have specific configuration beyond what is in `package.json`, which BTW can hold additional metadata. I think we can add the toml file when we do need it.

apps/igloo-kit-cli/src/project.rs
callicles: Could we leave the ProjectConfg file and add the `package.json`? 

ie read the `name` from the `package.json` and leave `language` in the project config.

I am working on `moose prod` and I will need to add in the TOML file what's needed to configure connections in production. 

ie use the standard fields from the `package.json` when available and the custom fields would come from the `project.toml` That way we have a clean delineation.

callicles (approved):
There are a lot of good things in here and I am approving in case you need to merge quickly.

I wanted to leverage the `toml` file for `moose prod` would be great if you could leave that or I can add it back in my next PR like we talked about.

phiSgr:
> use the standard fields from the package.json when available and the custom fields would come from the project.toml That way we have a clean delineation.

Yes that's a solution. I'll let you undo my changes in your branch

PR #317, opened 2024-02-02
Tim/514 288 stitch the backend and frontend together for the console

tg339:


apps/moose-console/src/components/query-interface.tsx
```diff
@@ -0,0 +1,284 @@
...
+                <ResizableHandle withHandle />
+                <ResizablePanel defaultSize={25}>
+                <div className="flex h-full px-4 w-full overflow-y-auto">
+                    <div className="grow">
+                        <div className="text-xs font-mono flex flex-row items-center">
+                            <span>SQL</span> <span className="flex-grow"/> 
+                            {
+                                sqlKeyWordCount < sqlKeyWords.length ? 
+                                <Button variant="ghost" onClick={() => setSqlKeyWordCount(sqlKeyWords.length)}>more</Button> : 
+                                <Button variant="ghost" onClick={() => setSqlKeyWordCount(5)}>less</Button>
```
phiSgr: A bit weird that clicking "more" then "less" doesn't get me back to the original state (i.e. showing 12 keywords).

Also I think I want to see tables before SQL keywords.

tg339: Good catch

apps/moose-console/src/app/infrastructure/databases/[databaseName]/tables/[tableId]/page.tsx
```diff
@@ -48,81 +51,60 @@ async function describeTable(databaseName: string, tableName: string): Promise<a
 async function getTable(databaseName: string, tableName: string): Promise<any> {
```
Comment (callicles) [inline apps/moose-console/src/app/infrastructure/databases/[databaseName]/tables/[tableId]/page.tsx:51]: Probably not worth doing here but this should be done in a follow up PR.

return type not being any and either a response or an error

apps/moose-console/src/components/query-interface.tsx
```diff
@@ -0,0 +1,285 @@
...
+        const handleKeyDown = (e: KeyboardEvent) => {
+            if (e.key === "Enter" && (e.ctrlKey || e.metaKey)) {
+                runQuery(value).then((results) => {
+                    e.preventDefault()
+                    setResults(results)
+                })
+            }
+        }
+
+        document.addEventListener("keydown", handleKeyDown)
```
callicles, line 198: Hum should you be adding event listener on the whole document? feels like that should only be on the text area no?

callicles (approved):
Added some comments to be solved in a follow up PR

apps/moose-console/src/components/query-interface.tsx
tg339: No it's for the run run shortcut. Whole doc is the way to go I believe. The removal happens when the component goes out of view

PR #323, opened 2024-02-02
docs: use the dev command in scripts in package.json

phiSgr:


apps/framework-docs/src/pages/getting-started/installation.mdx
```diff
@@ -21,6 +21,12 @@ MooseJS comes with a handy CLI for managing your MooseJS application. You can ru
 npx @514labs/moose-cli dev
 ```
 
+Because `@514labs/moose-cli` is in `devDependencies` of the generated project,
+you can also use `npm run` to invoke them after running `npm install`.
+```txt filename="Terminal" copy
+npm run dev
```
phiSgr, line 27: The `dev` here is in the `scripts` field of `package.json`. But mentioning this doesn't flow nicely in the paragraph.

PR #348, opened 2024-02-09
log cleanup and use of spinners

cjus:


apps/framework-cli/src/cli/routines.rs
```diff
@@ -263,29 +263,34 @@ async fn initialize_project_state(
...
-            Ok(())
-        }
-        Err(e) => {
-            debug!("Schema directory crawl failed");
-            debug!("Error: {:?}", e);
-            Err(e)
-        }
-    }
+
+    let _ = with_spinner("Processing schema file", || async {
```
phiSgr: If I understand async Rust correctly:

`with_spinner` runs the provided function, which immediately returns a `Future`, which does nothing unless `.await`ed. You then discarded the `Future` with `let _ =`. This block basically does nothing I think.

The problem here is that `with_spinner` works with sync functions, but not async functions.

We need another version for async functions. Or better yet, migrate the other sync functions to async functions. For subprocess commands, we can use [the tokio version](https://docs.rs/tokio/latest/tokio/process/struct.Command.html). Or just use https://docs.rs/docker-api/latest/docker_api/ for the docker commands

tg339: Yeah I think that's right. From what I understand, you either need to `await` the future or pass it to an executor like block_on or tokio spawn to have the async block run.

tg339: https://tokio.rs/tokio/tutorial/async & https://rust-lang.github.io/async-book/03_async_await/01_chapter.html

cjus: @phiSgr thanks for that catch!  Are you proposing that we:

- refactor all uses of process::Command to use the [tokio version](https://docs.rs/tokio/latest/tokio/process/struct.Command.html) 
- refactor [docker.rs](https://github.com/514-labs/moose/blob/main/apps/framework-cli/src/utilities/docker.rs) to use the [docker API for Rust](https://docs.rs/docker-api/latest/docker_api/) instead of converting its use of process::Command with tokio versions.

The second point above for rewriting the docker.rs module seems like a big lift for a logging cleanup ticket.

apps/framework-cli/src/cli/routines.rs
```diff
@@ -263,29 +263,35 @@ async fn initialize_project_state(
...
+                debug!("Schema directory crawl failed");
+                debug!("Error: {:?}", e);
+                Err(e)
+            }
         }
-    }
+    })
+    .await;
+
+    Ok(())
```
phiSgr: i think we should be returning the result of this `await`

apps/framework-cli/src/cli/watcher.rs
```diff
@@ -98,7 +98,30 @@ async fn create_framework_objects_from_schema_file_path(
...
+                        show_message!(MessageType::Error, {
+                            Message {
+                                action: "Schema".to_string(),
+                                details: format!("file failed to process: {}", e),
+                            }
+                        });
+                    }
+                }
+            })
+            .await;
```
phiSgr: I think we should use `?` here instead of discarding the result with `let _ =`

apps/framework-cli/src/cli/routines.rs
```diff
@@ -263,29 +263,34 @@ async fn initialize_project_state(
...
+            Err(e) => {
+                debug!("Schema directory crawl failed");
+                debug!("Error: {:?}", e);
+                Err(e)
+            }
         }
-    }
+    })
+    .await?;
+    Ok(())
```
phiSgr, line 293: looks like we can simplify this and remove ~~`; Ok(())`~~ `?; Ok(())`

apps/framework-cli/src/utilities/package_managers.rs
```diff
@@ -98,7 +112,39 @@ pub fn link_sdk(
...
     let output = command.output()?; // We should explore not using output here and instead using spawn.
-    println!("{}", String::from_utf8(output.stdout).unwrap());
+    match String::from_utf8(output.stdout) {
+        Ok(val) => {
+            debug!("{}", val);
+        }
+        Err(e) => {
+            error!("Error: {:?}", e);
+        }
+    }
```
phiSgr, line 122: I don't really like 8 lines of duplicated code.

phiSgr (approved):

apps/framework-cli/src/utilities/package_managers.rs
cjus: @phiSgr how are those lines duplicated code? The intent is to output status to the logger.

apps/framework-cli/src/cli/routines.rs
cjus: @phiSgr Removing the Ok(()) results in : `expected Result<(), Error>, found ()`
As it stands I can't simply remove the `Ok(())`

apps/framework-cli/src/utilities/package_managers.rs
phiSgr: The same lines are in `install_packages`,`run_build`, and `link_sdk`.

apps/framework-cli/src/cli/routines.rs
tg339: Removing the Ok(()) only would cause that expected error but if you also remove the ?; and the Ok(()) it should work. Minor detail though so don't let that be a blocker

tg339 (approved):

apps/framework-cli/src/utilities/package_managers.rs
cjus: Noted.  I image that's the case in other parts of the code base.

PR #349, opened 2024-02-09
use explicit data table instead of inner table in materialized view

phiSgr:
in preparation for the version sync trigger that writes to the data table

### DIFF
--- a/apps/framework-cli/src/framework/controller.rs
+++ b/apps/framework-cli/src/framework/controller.rs
@@ -27,7 +27,7 @@ use crate::infrastructure::olap;
 
 use std::io::ErrorKind;
 
-use crate::infrastructure::olap::clickhouse::ClickhouseView;
+use crate::infrastructure::olap::clickhouse::ClickhouseKafkaTrigger;
 
 use std::io::Error;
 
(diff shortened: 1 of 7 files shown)


apps/framework-cli/src/framework/controller.rs
```diff
@@ -27,16 +27,14 @@ use crate::infrastructure::olap;
...
+use crate::infrastructure::olap::clickhouse::ClickhouseKafkaTrigger;
 
 use std::io::Error;
 
 use crate::infrastructure::olap::clickhouse::ConfiguredDBClient;
 
 use super::schema::parse_schema_file;
 use super::schema::DataModel;
-use super::schema::MatViewOps;
-use super::schema::TableOps;
```
phiSgr, line 39: Not sure if these two interfaces are the right abstraction.

apps/framework-cli/src/framework/controller.rs
```diff
@@ -137,23 +140,17 @@ pub(crate) async fn create_or_replace_view(
...
-) -> Result<(), Error> {
+) -> anyhow::Result<()> {
     info!("Creating table: {:?}", fo.table.name);
-    let create_table_query = fo.table.create_table_query().map_err(|e| {
-        Error::new(
-            ErrorKind::Other,
-            format!("Failed to get clickhouse query: {:?}", e),
-        )
-    })?;
-    let drop_table_query = fo.table.drop_table_query().map_err(|e| {
```
phiSgr, line 151: The `map_err` calls add quite a lot of noise. We don't need to convert the error by using anyhow. If we do want to add extra context later, we can rethink a better abstraction.

callicles (approved):
LGTM

PR #390, opened 2024-02-15
clickhouse fixes

phiSgr:


apps/framework-cli/src/infrastructure/olap/clickhouse/queries.rs
```diff
@@ -187,7 +187,7 @@ impl DropMaterializedViewContext {
     ) -> Result<DropMaterializedViewContext, UnsupportedDataTypeError> {
         Ok(DropMaterializedViewContext {
             db_name: view.db_name,
-            view_name: format!("{}_view", view.name),
+            view_name: view.name,
```
phiSgr, line 190: Bug introduced in #349

callicles (approved):

apps/framework-cli/src/infrastructure/console.rs
```diff
@@ -49,6 +49,9 @@ pub async fn post_current_state_to_console(
         .map(|fo| fo.data_model.clone())
         .collect();
 
+    olap::clickhouse::check_ready(configured_db_client)
```
phiSgr, line 52: Occasionally `fetch_all_tables` fail on start-up because clickhouse is not yet ready.

PR #394, opened 2024-02-15
skip broken messages and best effort date parsing for Kafka

phiSgr:


callicles (approved):

phiSgr:
Included [`date_time_input_format = 'best_effort'`](https://clickhouse.com/docs/en/operations/settings/formats#date_time_input_format) as suggested by @georgevanderson, so that clickhouse can parse ISO 8601.
Maybe that should be globally set, but putting it in Kafka settings is easy enough.

PR #413, opened 2024-02-19
remove docker check in init

phiSgr:


apps/framework-cli/src/cli/routines/validate.rs
```diff
@@ -89,7 +89,7 @@ impl Routine for ValidateRedPandaCluster {
             )
         })?;
 
-        if output.contains(REDPANDA_CONTAINER_NAME) || output.contains("localhost") {
+        if output.contains("redpanda") || output.contains("localhost") {
```
callicles: Can we use a constant?

tg339 (approved):

PR #428, opened 2024-02-20
Restore project.toml inside the app directory

callicles:
This restores the project.toml in the app directory and puts a hirarchy in place to be able to add project files for multiple languages later on.

I tested manually:
* Initializing a project
* Changing the password 
* Running dev
* Logging clickhouse directly with the new password
* looking at the console to make sure the view works properly.

What's not done in this:
* When project file is changed - regen docker-compose file
* Looks like the Table view is broken in `main`


apps/moose-console/src/lib/clickhouse.ts
```diff
@@ -0,0 +1,28 @@
+import { createClient } from "@clickhouse/client-web";
+import { Project, getCliData } from "../app/db";
+import { unstable_noStore as noStore } from "next/cache";
+
+export function getClient(project?: Project) {
+    // This is to make sure the environment variables are read at runtime
+    // and not during build time
+    noStore();
+
+    console.log(project);
```
georgevanderson: leftover

apps/moose-console/src/app/page.tsx
```diff
@@ -53,6 +53,8 @@ export default async function OverviewPage(): Promise<JSX.Element> {
   noStore();
   const data = await getCliData();
 
+  console.log(data);
+
```
georgevanderson: leftover

apps/framework-cli/src/project.rs
```diff
@@ -11,111 +11,146 @@
...
+use crate::utilities::constants::PROJECT_CONFIG_FILE;
+use crate::utilities::constants::{APP_DIR, APP_DIR_LAYOUT, CLI_PROJECT_INTERNAL_DIR, SCHEMAS_DIR};
+
+// We have explored using a Generic associated Types as well as
+// Dynaimc Dispatch to handle the different types of projects
+// the approach with enums is the one that is the simplest to put into practice and
+// maintain. With Copilot - it also has the advaantage that the boiler plate is really fast to write
+#[derive(Debug)]
+pub enum Project {
+    Typescript(TypescriptProject),
```
phiSgr, line 58: The default design I'd go for would be to put only language specific stuff in the enum.

callicles: Do you have a suggestion on how you would do it?

phiSgr: 🤔

```
pub struct TypescriptProjectConfig {
    // we don't have language specific config right now
}

pub enum LanguageConfig {
    Typescript(TypescriptProjectConfig)
}

pub struct Project {
    pub name: String,
    pub language_config: LanguageConfig,
    ...
}
```

phiSgr: I guess the downside of that is the de-serialization can't be auto-generated.

But the current design needs a `match` in both loading and writing.

Maybe it won't be too verbose if we do this when we load the `Project` struct.

```rust
Config::builder()
    // name and version comes from package.json
    .add_source(File::from(package_json_location).required(true))
    // other config in toml
    .add_source(File::from(project_file).required(true))
```

apps/framework-cli/src/cli.rs
```diff
@@ -76,25 +78,29 @@ async fn top_command_handler(settings: Settings, commands: &Commands) {
                     exit(1);
                 }
 
-                let project = Project::from_dir(dir_path, name.clone(), *language);
+                let project = Project::new(dir_path, name.clone(), *language);
+                let project_arc = Arc::new(project);
```
phiSgr, line 82: I guess the change comes from that we are having too many `project.clone()`. I wonder if we can avoid cloning and reference counting by adding lifetime to the structs.

Not saying that I want that done in this PR.

callicles: Hum that's a good question. Yeah this is allowing the same instance to be passed around across threads without cloning.

apps/framework-cli/src/project.rs
callicles: Good suggestion!

I will try to bake that in

callicles: Update with something along those lines - please check it again

phiSgr (approved):

PR #451, opened 2024-02-20
Use merge tree engine for data table

phiSgr:


apps/framework-cli/src/infrastructure/olap/clickhouse/queries.rs
```diff
@@ -49,18 +49,23 @@ impl CreateTableQuery {
...
                 kafka_host, kafka_port, topic, KAFKA_SETTINGS,
             ),
+            true,
         )
     }
 
     pub fn build(
         table: ClickhouseTable,
         engine: String,
+        ignore_primary_key: bool,
```
callicles: it looks like this depends on the engine - wouldn't it be better to have an if Kafka then that else this?

Feels like a weird parameter to ask the caller to set

phiSgr: Yeah I don't really like that either, I changed that to an enum.

callicles (approved):
Nice Work!

PR #469, opened 2024-02-23
versioned tables, version sync, and rewritten watcher logic

phiSgr:
`tail -n +1 package.json .moose/versions/0.0/simple.prisma app/datamodels/simple.prisma app/flows/User_migrate__0_0__1_0.sql`

```
==> package.json <==
{
  "name": "super-new-data-app",
  "version": "1.0",
  "scripts": {
    "dev": "igloo-cli dev"
  },
  "dependencies": {},
  "devDependencies": {
    "@514labs/moose-cli": "latest"
  }
}

==> .moose/versions/0.0/simple.prisma <==
model User {
    id String @id
}

==> app/datamodels/simple.prisma <==
model User {
    id String @id
    name String
}

==> app/flows/User_migrate__0_0__1_0.sql <==
(id) -> (id, 'name unknown')
```

```bash
curl -vvvvv localhost:4000/ingest/User/0.0 -XPOST -d '{"id":123}'
curl -vvvvv localhost:4000/ingest/User/1.0 -XPOST -d '{"id":234, "name":"stuff"}'
```

We can then find in `User_1_0`

|id | name|
|---|---|
|123 | name unknown|
|234 | stuff|


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -210,9 +214,19 @@ async fn router(
     );
 
     let route_split = route.to_str().unwrap().split('/').collect::<Vec<&str>>();
-
+    println!("route table: {:?}", route_table.read().await);
     match (req.method(), &route_split[..]) {
         (&hyper::Method::POST, ["ingest", _]) => {
+            ingest_route(
+                req,
+                route.join(current_version),
```
callicles, line 222: Not sure I get this - can you explain where `current_version` is coming from?

apps/framework-cli/src/framework/controller.rs
```diff
@@ -64,71 +89,148 @@ pub struct RouteMeta {
     pub view_name: Option<String>,
 }
 
+pub fn get_all_version_syncs(
+    project: &Project,
+    framework_object_versions: &FrameworkObjectVersions,
+) -> anyhow::Result<Vec<VersionSync>> {
+    let flows_dir = project.app_dir().join("flows");
```
callicles: Suggestion to move this to the `Project` `struct` since it will probably be used a bunch.

apps/framework-cli/src/framework/schema.rs
```diff
@@ -327,29 +318,33 @@ pub fn ast_mapper(ast: SchemaAst) -> Result<Vec<DataModel>, ParsingError> {
...
-    debug!("All objects created, generating sdk...");
-    let sdk_location = generate_ts_sdk(project, compilable_objects)?;
-    let package_manager = package_managers::PackageManager::Npm;
-    package_managers::install_packages(&sdk_location, &package_manager)?;
-    package_managers::run_build(&sdk_location, &package_manager)?;
-    package_managers::link_sdk(&sdk_location, None, &package_manager)?;
-    Ok(())
-}
+//
+// pub async fn process_schema_file(
```
callicles: Is that all commented on purpose?

apps/framework-cli/src/framework/schema.rs
```diff
@@ -64,13 +53,54 @@ impl Display for UnsupportedDataTypeError {
...
 impl std::error::Error for UnsupportedDataTypeError {}
 
-/// A function that maps an input type to an output type
-type MapperFunc<I, O> = fn(i: I) -> O;
+#[derive(Debug, Clone)]
+pub struct DuplicateModelError {
+    pub model_name: String,
+    pub file_path: PathBuf,
+    pub other_file_path: PathBuf,
+}
```
callicles, line 61: Are you using this to detect if a model is duplicated in the directory?

apps/framework-cli/src/framework/controller.rs
```diff
@@ -64,71 +89,148 @@ pub struct RouteMeta {
...
+            if let Some(captures) =
+                VERSION_SYNC_REGEX.captures(entry.file_name().to_string_lossy().as_ref())
+            {
+                println!("Captures: {:?}", captures);
+                let from_table_name = captures.get(1).unwrap().as_str();
+                let to_table_name = captures.get(4).map_or(from_table_name, |m| m.as_str());
+
+                let from_version = captures.get(2).unwrap().as_str().replace('_', ".");
+                let to_version = captures.get(5).unwrap().as_str().replace('_', ".");
+
```
callicles, line 111: If I get this right - here you are using the names of the migrations in the folder as the history of all the versions.

right?

callicles (commented):
Nice progress,

There are a bunch of `println` left in there - not sure if that's on purpose of if that should be put in the `log` output.

Is current version focused on getting older versions from the local env? I don't see the part where we would look up in the git history to get it. Might be a follow up PR

apps/framework-cli/src/cli/local_webserver.rs
phiSgr: From `project`

apps/framework-cli/src/framework/controller.rs
phiSgr: Will be loading the list of versions from `project.toml`

apps/framework-cli/src/framework/schema.rs
phiSgr: This is the error that gets ~~`yeet`ed~~ ~~thrown~~ returned when the duplication is detected.

apps/framework-cli/src/framework/controller.rs
callicles: 🤦🏻‍♂️ - yep

apps/framework-cli/src/cli/local_webserver.rs
callicles: kk - and how does the route match work? ie `(&hyper::Method::POST, ["ingest", _])` vs `(&hyper::Method::POST, ["ingest", _, _])`

phiSgr: If current version is 1.0, `ingest/User` goes to `ingest/User/1.0`

tg339 (approved):

apps/moose-console/src/lib/utils.ts
```diff
@@ -16,8 +16,11 @@ export function getIngestionPointFromModel(
 }
 
 export function getModelFromRoute(route: Route, cliData: CliData): DataModel {
-  const routeTail = route.route_path.split("/").at(-1);
-  return cliData.models.find((model) => model.name === routeTail);
+  const routeTail = route.route_path.split("/").at(-2); // -1 is now version number
+  console.log(route.route_path, routeTail);
```
callicles: Leftover console.log?

callicles (approved):
Great work!

small leftover console.log

PR #471, opened 2024-02-24
fix generated package.json dev command

phiSgr:


tg339 (approved):
haha I wasn't sure what was going on. Thought I had the wrong version installed :p

PR #478, opened 2024-02-25
adding enum parsing and added to test

tg339:


apps/framework-cli/src/framework/schema.rs
```diff
@@ -92,6 +107,12 @@ pub struct DataModel {
     pub version: i8,
 }
 
+#[derive(Debug, Clone, Serialize)]
+pub struct ValueEnum {
```
phiSgr: I'm not sure what "value" in `ValueEnum` mean.

phiSgr (approved):
Is it allowed in prisma to have the enum and the model using it being in different files? If so do we want to support that?

phiSgr (commented):
BTW submodule from the other PR is left in this PR.

apps/framework-cli/src/framework/schema.rs
```diff
@@ -280,53 +311,75 @@ fn field_to_column(f: &Field) -> Result<Column, ParsingError> {
...
+        name: schema_name,
+        version,
+    })
 }
 
-pub fn ast_mapper(ast: SchemaAst) -> Result<Vec<DataModel>, ParsingError> {
-    ast.iter_tops()
-        .map(|(_id, t)| top_to_schema(t))
-        .collect::<Result<Vec<DataModel>, ParsingError>>()
+fn top_to_enum(e: &Enum) -> DataEnum {
```
callicles: Not a blocker - but please take the time to have fully spelled out names for variables. It makes it easier to read the code.

Suggestion here for the function signature:

```rust
fn convert_prisma_enum_to_moose_enum(prisma_enum: &Enum) -> DataEnum {
```

callicles (approved):
Love that you added some more unit tests in the code base. 

Good work!

apps/framework-cli/src/framework/schema.rs
phiSgr: I can live without the "convert" and the second "enum".

apps/moose-console/src/components/model-table.tsx
```diff
@@ -22,19 +31,15 @@ export default function ModelTable({ datamodel }: { datamodel: DataModel }) {
                 {field.name}
               </div>
               <div className="grow basis-1 text-muted-foreground">
-                {" "}
-                {field.data_type}
+                {processType(field.data_type)}
```
phiSgr, line 35: The type of `data_type` in `interface Column` should be also changed

apps/framework-cli/src/framework/schema.rs
```diff
@@ -293,28 +326,66 @@ fn field_to_column(f: &Field) -> Result<Column, ParsingError> {
...
         }
-        _ => Err(ParsingError::UnsupportedDataTypeError {
-            type_name: "we don't currently support anything other than models".to_string(),
-        }),
-    }
+        Top::Enum(e) => {
+            enums.push(primsa_to_moose_enum(e));
+        }
+        _ => {
+            "we currently only support models and enums".to_string();
```
phiSgr: Unused value.

PR #497, opened 2024-03-01
Carlos/514 517 package app for deployment

cjus:
This PR adds three new moose-cli commands: `moose docker init`, `moose docker build` and `moose prod`. The PR is a functioning work-in-progress that is required in order proceed with cloud configuration and testing.


apps/framework-cli/src/cli.rs
```diff
@@ -152,6 +161,37 @@ async fn top_command_handler(settings: Settings, commands: &Commands) {
                 controller.add_routine(Box::new(CleanProject::new(project_arc, run_mode)));
                 controller.run_routines(run_mode);
             }
+            Commands::Docker { sub_method } => match sub_method.as_str() {
```
tg339, line 164: I'm pretty sure there's a clean enum based approach here as opposed to raw string matching. Not a blocker but something you may want to look into as you keep moving this forward

apps/framework-cli/src/cli/commands.rs
```diff
@@ -38,10 +38,16 @@ pub enum Commands {
...
+    Prod {},
     // Updates the redpanda cluster and clickhouse database with the latest objects
     Update {},
     // Stops development infrastructure
     Stop {},
     // Clears all temporary data and stops development infrastructure
     Clean {},
+    /// Docker related commands
+    Docker {
+        sub_method: String,
```
tg339, line 51: Same as above. Could be an enum of the docker methods you want. Again, not a blocker just calling out

apps/framework-cli/src/cli/routines.rs
```diff
@@ -257,16 +258,87 @@ pub async fn start_development_mode(project: Arc<Project>) -> anyhow::Result<()>
...
+            action: "Starting".to_string(),
+            details: "production mode".to_string(),
+        }
+    );
+
+    let mut route_table = HashMap::<PathBuf, RouteMeta>::new();
+
+    info!("Initializing project state");
+    initialize_project_state(project.clone(), &mut route_table).await?;
+    let route_table: &'static RwLock<HashMap<PathBuf, RouteMeta>> =
```
tg339, line 275: Do we actually need the lock here? Might be something to look into as we explore the production server going forward

tg339 (approved):
LGTM. Some comments in here that we can address in future PRs if you don't want to block forward momentum.

apps/framework-cli/src/cli/routines.rs
phiSgr: Yeah I think this can be just a reference, but I don't mind merge this as is.

apps/framework-cli/src/cli.rs
```diff
@@ -152,6 +161,37 @@ async fn top_command_handler(settings: Settings, commands: &Commands) {
...
+                "init" => {
+                    let run_mode = RunMode::Explicit {};
+                    info!("Running docker init command");
+                    let project = Project::load_from_current_dir()
+                        .expect("No project found, please run `moose init` to create a project");
+                    let project_arc = Arc::new(project);
+                    let mut controller = RoutineController::new();
+                    controller.add_routine(Box::new(CreateDockerfile::new(project_arc)));
```
phiSgr, line 172: Is the user supposed to edit this file? If yes then maybe it shouldn't be in the internal directory, if no then we can write it every time before we `build`.

Maybe the answer is "not really but here's an escape hatch if the user really wants to edit the file."

phiSgr (approved):

apps/framework-cli/src/cli/commands.rs
```diff
@@ -38,10 +38,16 @@ pub enum Commands {
     // },
     /// Starts a local development environment to build your data-intensive app or service
     Dev {},
+    /// Start a remote environment for use in cloud deployments
+    Prod {},
```
phiSgr, line 42: `next` call this `start`, BTW.

apps/framework-cli/src/cli.rs
cjus: Definitely something we can keep an eye on, in case the use-case emerges.  But I do see your  point. Good catch!

apps/framework-cli/src/cli/commands.rs
cjus: hmmm. I think `start` would be misleading to developers unless the comment says start in production.

PR #503, opened 2024-03-02
Refactor code to pull environment variables using the Config crate

cjus:
Because I had trouble getting the Rust Config crate to merge environment variables into the variables loaded from a config file, I fell back to doing this manually in an earlier PR.  This PR addresses the core issue that prevented me from succeeding on my earlier attempt.

The core issue was that I needed to use the Config `.prefix_separator("_")` in order to specify that environment variables will begin with MOOSE_ and that hierarchy would be represented with double underscore "dunderscore" as in the following example: `MOOSE_CLICKHOUSE__DB_NAME`

```rust
.add_source(Environment::with_prefix("MOOSE")
                    .prefix_separator("_")
                    .separator("__"),
```


callicles (approved):
Nice work! Happy you were able to sort it out ☺️

PR #506, opened 2024-03-03
Scripts used for k8s cluster

cjus:
Collection of scripts used to configure moose `dogfooding` Kubernetes cluster.

### DIFF
--- a/apps/framework-cli/deploy/kubernetes/moose-deployment.yaml
+++ b/apps/framework-cli/deploy/kubernetes/moose-deployment.yaml
@@ -1,0 +1,66 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: webserver-deployment
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: mooseweb
+  template:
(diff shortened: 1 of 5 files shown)


apps/framework-cli/deploy/kubernetes/moose-deployment.yaml
```diff
@@ -0,0 +1,66 @@
...
+      app: mooseweb
+  template:
+    metadata:
+      labels:
+        app: mooseweb
+        ver: v1
+    spec:
+      containers:
+      - name: web
+        image: pnxtech/moose-deployment-x86_64-unknown-linux-gnu:latest
```
callicles, line 18: You don't have access to the 514 org?

callicles (approved):

apps/framework-cli/deploy/kubernetes/moose-deployment.yaml
cjus: No not yet.  I remember we discussed it - just don't remember the outcome. Will update when I gain access.

PR #526, opened 2024-03-05
Update docker repo and new redpanda credentials keys

cjus:


callicles (approved):

apps/framework-cli/deploy/kubernetes/moose-deployment.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
       - name: web
-        image: pnxtech/moose-deployment-x86_64-unknown-linux-gnu:latest
+        image: 514labs/moose-deployment-x86_64-unknown-linux-gnu:latest
```
callicles, line 18: This is a specific deployment image for us right? ie that should be a private image

cjus: yes, will be private by default.

PR #549, opened 2024-03-10
Tim/514 459 typescript instead of prisma

tg339:


apps/framework-cli/src/framework/schema.rs
```diff
@@ -376,10 +384,154 @@ pub fn ast_mapper(ast: SchemaAst) -> Result<FileObjects, ParsingError> {
...
+    let name = enum_decl.id.sym.to_string();
+    let values = enum_decl
+        .members
+        .iter()
+        .map(|member| match member {
+            TsEnumMember {
+                id: TsEnumMemberId::Ident(ident),
+                ..
+            } => ident.sym.to_string(),
+            _ => "unsupported".to_string(),
```
phiSgr, line 422:
```
.map(|member| match &member.id {
    TsEnumMemberId::Ident(ident) => ident.sym.to_string(),
    TsEnumMemberId::Str(str) => str.value.to_string()
})
```

phiSgr: If enum values are numbers, then we should use those numbers in clickhouse as well.

```
enum Direction {
  Up = 2,
  Down,
  Left,
  Right,
}

console.log(JSON.stringify({"direction": Direction.Left}))
```

gives `{"direction":4}`.

---

Also, in clickhouse [Numbers are assigned starting from 1 by default](https://clickhouse.com/docs/en/sql-reference/data-types/enum), but in ts they start from 0 by default.

apps/framework-cli/src/framework/schema.rs
```diff
@@ -376,10 +384,154 @@ pub fn ast_mapper(ast: SchemaAst) -> Result<FileObjects, ParsingError> {
...
+        .filter_map(|field| match field {
+            swc_ecma_ast::TsTypeElement::TsPropertySignature(prop) => Some({
+                // match the key's sym if it's an ident
+                let name = match *prop.key.clone() {
+                    Expr::Ident(ident) => ident.sym.to_string(),
+                    _ => return None,
+                };
+                // match the type of the value and return the right column type
+                let data_type = match *prop.type_ann.clone().expect("no type for property") {
+                    TsTypeAnn { type_ann, .. } => match *type_ann {
```
phiSgr, line 447: Handle `Date` type here.

Also, not being able to specify an integer field feels wrong. 😕

apps/framework-cli/src/framework/schema.rs
phiSgr: If enum values are strings, then we should use the values instead of the identifiers, otherwise we'll have to do some mapping in the send functions before calling `JSON.stringify`.

```
enum Direction {
  Up = "UP",
  Down = "DOWN",
  Left = "LEFT",
  Right = "RIGHT",
}
```

And if the enum values are other stuff, we should probably reject the schema.

PR #566, opened 2024-03-11
Add lazy static support for Project

cjus:


apps/framework-cli/src/project.rs
```diff
@@ -36,6 +38,19 @@ use crate::project::typescript_project::TypescriptProject;
 use crate::utilities::constants::{APP_DIR, APP_DIR_LAYOUT, CLI_PROJECT_INTERNAL_DIR, SCHEMAS_DIR};
 use crate::utilities::constants::{FLOWS_DIR, PROJECT_CONFIG_FILE};
 
+lazy_static! {
+    pub static ref PROJECT: Arc<Mutex<Project>> = Arc::new(Mutex::new(Project {
```
phiSgr: We don't need `Arc` here because it'll live forever as a global variable.

If we can initialize it (from config files and env vars) we won't need mutability, and we can drop the `Mutex` as well.

cjus: @phiSgr please see the discussion on the Linear ticket: https://linear.app/514/issue/514-601/refactor-project-to-use-lazy-static

phiSgr (approved):
I still believe that we can safely remove the mutex, but that can be done (way) later.

PR #582, opened 2024-03-14
Add flow service to docker

DatGuyJonathan:
Summary of initial version of the flow service:

- `cli init` command creates a .moose/deno/transform.ts script
- `cli dev` command creates a deno container with the moose project as a shared volume & runs the transform script
- Transform script...
  - walks through the flows dir to figure out source, destination, & flow functions
  - listens to source topics, runs the transform function on them, and produce the result to destination topic

Currently, it requires you to restart `dev` to pick up model changes. Next step is to improve the local dev workflow so deno can pick up model changes dynamically.


callicles (approved):

apps/framework-cli/src/project.rs
```diff
@@ -171,6 +172,17 @@ impl Project {
         Ok(())
     }
 
+    pub fn create_deno_files(&self) -> Result<(), std::io::Error> {
+        let deno_dir = self.internal_dir()?.join("deno");
+        let transform_file_path = deno_dir.join("transform.ts");
```
callicles: Those strings could be constants 😄

PR #587, opened 2024-03-14
Rust based event capture sdk

DatGuyJonathan:
Basic framework for capturing events. It's currently not sending the data anywhere. We can turn it on when there's a production moose app.

The capture macro is currently being called from top level CLI commands. We can add it to other places in the future. I also extracted out the session ID in the logger into a lazy static so other parts of the code can use the same session ID.

Screenshot 2024-03-14 at 2 56 21 PM

### DIFF
--- a/apps/framework-cli/src/utilities/capture.rs
+++ b/apps/framework-cli/src/utilities/capture.rs
@@ -49,6 +49,7 @@ pub struct UserActivity {
 
 macro_rules! capture {
     ($activity_type:expr, $sequence_id:expr, $project_name:expr) => {
+        use crate::project::PROJECT;
         use crate::utilities::capture::{ActivityType, UserActivity};
         use crate::utilities::constants;
         use chrono::Utc;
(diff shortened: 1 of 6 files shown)


apps/framework-cli/src/utilities/capture.rs
```diff
@@ -0,0 +1,93 @@
...
+            project: $project_name,
+            activity_type: $activity_type,
+            sequence_id: $sequence_id,
+            timestamp: Utc::now(),
+            cli_version: constants::CLI_VERSION.to_string(),
+        });
+
+        // Get the environment variables
+        // let moose_contributor = std::env::var("MOOSE_CONTRIBUTOR").unwrap_or("unknown".to_string());
+        let scheme = std::env::var("SCHEME").unwrap_or("http".to_string());
```
callicles: Should that be part of the CLI configuration? 

That way you could reuse the same loading structure

cjus: @DatGuyJonathan ping me when you're ready to look at this.  I've built up some related understanding recently.

tg339 (approved):

apps/framework-cli/src/utilities/capture.rs
DatGuyJonathan: Thanks for pointing that out @callicles. @cjus showed me how those CLI configs worked, and I've updated the PR to follow the same structure.

PR #589, opened 2024-03-15
Nicolas/514 614 refactor clickhouse kafka table into a separate process for

callicles:
This replaces the kafka table sync we had with Clickhouse and spawns different tasks for the syncing to happen between different topics and tables.

It also removes the tests running on mac as we can't use service containers there for some of the tests that were added here.

They will be refactored in the `tests` folder in a subsequent PR.


tg339 (approved):

cjus (approved):
This is a large PR with lots of refactoring.  My sense is the sooner we can merge and test this further the better.  The code looks clean to me!

PR #606, opened 2024-03-16
Updated messaging to display correctly in prod

cjus:
Existing code already handles binding to a non-localhost address via configuration file and environment variables.
For example:

```
export MOOSE_HTTP_SERVER_CONFIG__HOST=0.0.0.0
export MOOSE_HTTP_SERVER_CONFIG__PORT=4000
```

However, server output to console was hardcoded to say localhost rather than display the address passed in from the user.


cjus:
@callicles this PR is not a block as I will use the environment variables on Kubernetes.

callicles (approved):

PR #614, opened 2024-03-17
initial commit on Pulumi PoC and documentation

cjus:
This PR kickstarts an investigation into using Pulumi and provides a placeholder location to store related documentation and configuration files.


callicles (commented):
Nice experiment!

I am not sure we will want it that deeply nested in the tree.
Look at how it would integrate with turborepo, we would want to be able to depend on other packages in the repo.

cjus:
@callicles These files are super temporary - I imagine we'll move documentation to our documentation site.  For now I just wanted to store these files somewhere so we don't lose them.

callicles (approved):
Cool that works then

PR #615, opened 2024-03-18
Work in progress structure and content

cjus:


callicles (approved):
Awesome documentation! Reading through this I wonder if we could simplify the secret handling further with using 1 secret with multiple fields instead of several secrets.

PR #621, opened 2024-03-18
Scaffold sample flow for new project

DatGuyJonathan:
Screenshot 2024-03-20 at 1 19 44 PM


apps/framework-cli/src/framework/transform.ts
```diff
@@ -73,6 +73,17 @@ const getFlows = (): Map<string, Map<string, string>> => {
   return output;
 };
 
+const jsonDateReviver = (key: string, value: unknown): unknown => {
+  const iso8601Format =
```
DatGuyJonathan, line 77: Got this from https://stackoverflow.com/a/14322189 :pray:

phiSgr: 😱

phiSgr (approved):

apps/framework-cli/src/framework/transform.ts
DatGuyJonathan: Yea haha, we should revisit further down the line.

PR #622, opened 2024-03-18
new moose build command and update to docker build process.

cjus:


apps/framework-cli/src/cli.rs
```diff
@@ -168,16 +184,6 @@ async fn top_command_handler(settings: Settings, commands: &Commands) {
                 controller.run_routines(run_mode);
             }
             Commands::Docker { sub_method } => match sub_method.as_str() {
```
callicles: should we remove this all together?

callicles (commented):
Quick question about former code clean up

apps/framework-cli/src/cli.rs
cjus: No we still need `moose docker build` and I imagine we'll want `moose docker push` to push to repos later. I believe @03cranec has a linear ticket for that.

apps/framework-cli/src/cli.rs
```diff
@@ -200,51 +219,6 @@ async fn top_command_handler(settings: Settings, commands: &Commands) {
...
             }
-            Commands::Docker { sub_method } => match sub_method.as_str() {
-                "init" => {
-                    let run_mode = RunMode::Explicit {};
-                    info!("Running docker init command");
-                    let project = Project::load_from_current_dir()
-                        .expect("No project found, please run `moose init` to create a project");
-                    let project_arc = Arc::new(project);
-
-                    crate::utilities::capture::capture!(
```
callicles, line 211: Looks like we are loosing this with the change - we probably want to instrument the `build` command all the same

cjus: Added new capture points. I also refactored capture.rs ActivityType's to match this PRs changes.

callicles (approved):

PR #629, opened 2024-03-19
Fix Dockerfile expects versions folder to be present

cjus:
The Dockerfile build was failing because it expects a versions folder to be present - even if empty.

```
# Copy the application files to the container
COPY ./app ./app
COPY ./versions .moose/versions
```

### DIFF
--- a/apps/framework-cli/src/cli/routines/docker_packager.rs
+++ b/apps/framework-cli/src/cli/routines/docker_packager.rs
@@ -72,8 +72,8 @@ impl Routine for CreateDockerfile {
 
         ensure_docker_running()?;
 
-        let file_path = internal_dir.join("packager/versions/.gitkeep");
-        let file_path_display = file_path.clone();
+        let mut file_path = internal_dir.join("packager/versions/.gitkeep");
+        let mut file_path_display = file_path.clone();
 
         info!("Creating versions at: {:?}", file_path_display);
         fs::create_dir_all(file_path.parent().unwrap()).map_err(|err| {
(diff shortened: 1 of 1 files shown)


apps/framework-cli/src/cli/routines/docker_packager.rs
```diff
@@ -72,10 +72,24 @@ impl Routine for CreateDockerfile {
 
         ensure_docker_running()?;
 
-        let file_path = internal_dir.join("packager/Dockerfile");
+        let file_path = internal_dir.join("packager/versions/.gitkeep");
```
phiSgr: I don't like using shadowing when the two variables are not conceptually the same thing.

cjus: That makes sense! Nice catch.

apps/framework-cli/src/cli/routines/docker_packager.rs
```diff
@@ -72,10 +72,24 @@ impl Routine for CreateDockerfile {
...
+                    "Failed".to_string(),
+                    "to create directory for app versions".to_string(),
+                ),
+                err,
+            )
+        })?;
+
         info!("Creating Dockerfile at: {:?}", file_path_display);
+        let file_path = internal_dir.join("packager/Dockerfile");
+        let file_path_display = file_path.clone();
```
phiSgr: BTW, no need to `.clone`, just pass by reference instead of move.

`fs::write(&file_path`

cjus: clone() is required. I tried pulling it out but needed due to:

```rust
 fs::create_dir_all(file_path.parent().unwrap()).map_err(|err| {
 ```

phiSgr (approved):

PR #633, opened 2024-03-19
Add filewatcher to flows for local dev

DatGuyJonathan:
I tried [Deno.watchFs](https://examples.deno.land/watching-files), but it seemed unreliable with Docker volumes. Specifically, I would delete a flow file locally, see it deleted in the container, but Deno's filewatcher wasn't picking it up.

[Chokidar](https://www.npmjs.com/package/chokidar) was able to because of its polling feature:

> It is typically necessary to set this to true to successfully watch files over a network, and it may be necessary to successfully watch files in other non-standard situations.


callicles (approved):
Elegant way to do this!

PR #640, opened 2024-03-20
Initial Mixplank Template

georgevanderson:
Stashed next changes before this PR got too big. Converted from mixpanel data stubs to receiving data from script. 

Idea is script is in the `public` dir of your vercel template. So when you deploy mixpanel app, you also are deploying the script which you use wherever it is instrumented. That way we can avoid any npm registry. Next changes are the rest of the queries and UI to accompany it
templates/mixplank/next/public/script.js


templates/mixplank/next/public/script.js
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     * Track page hit
+     */
+    function _trackPageHit() {
+        _setSessionId();
+
+        // If local development environment
+        // if (/^localhost$|^127(\.[0-9]+){0,2}\.[0-9]+$|^\[::1?\]$/.test(location.hostname) || location.protocol === 'file:') return;
+        // If test environment
+        if (window.__nightmare || window.navigator.webdriver || window.Cypress) return;
```
callicles, line 128: What's `nightmare`?

templates/mixplank/next/public/script.js
```diff
@@ -0,0 +1,193 @@
...
+                method: 'POST',
+                mode: 'no-cors',
+                headers: {
+                    'Content-Type': 'application/json'
+                },
+                body: JSON.stringify(
+                    {
+                        eventId: _uuidv4(),
+                        timestamp: new Date(),
+                        version: '1',
```
callicles: What's this version for? should it be synced with the version of the model?

templates/mixplank/next/src/app/data.ts
```diff
@@ -1,18 +1,21 @@
 import { createClient } from "@clickhouse/client-web"
 
-const client = createClient({
+export const clickhouseClient = createClient({
     host: "http://localhost:18123",
     username: "panda",
     password: "pandapass",
     database: "local",
```
callicles, line 8: not sure if is not already on your radar but making this configurable through env will be needed.

templates/mixplank/next/src/app/data.ts
```diff
@@ -1,18 +1,21 @@
 import { createClient } from "@clickhouse/client-web"
```
callicles: is this running on the server or in the client? thinking about in prod where this might not have network access to the db from the client

callicles (commented):
Looks good, the code is really clean.

Some questions for context

templates/mixplank/next/public/script.js
georgevanderson: Took this from plausible script. Looks like its a fairly popular web automation library, AKA don't send tracking for tests

templates/mixplank/next/src/app/data.ts
georgevanderson: Good call, will use server side instead

templates/mixplank/next/public/script.js
georgevanderson: Removed since we know this from table

callicles (approved):

PR #642, opened 2024-03-21
fix: docker build --docker on some systems

callicles:


cjus (approved):
I haven't used `debian::bookwork-slim` but this PR looks good otherwise.

PR #644, opened 2024-03-21
Nicolas/514 646 enable user to configure debugging logs to stdout for

callicles:
Adds a configuration to be able to output logs to stdout in k8s


cjus (approved):
LGTM

PR #648, opened 2024-03-21
Ignore null/undefined return value from flow function

DatGuyJonathan:
This allows users to add a filter logic in their flow function so they don't have to transform every row of source data.


cjus (approved):
This looks good to me.

PR #654, opened 2024-03-21
create a sorted vec of DCM versions

cjus:


apps/framework-cli/src/utilities/system.rs
```diff
@@ -13,3 +13,24 @@ pub fn copy_directory(source: &PathBuf, destination: &PathBuf) -> Result<(), std
 
     Ok(())
 }
+
+pub fn read_directory<P: AsRef<Path>>(path: P) -> io::Result<Vec<String>> {
```
callicles, line 17: Not sure if you looked at the standard library, if so why not using it? 

if not:
https://doc.rust-lang.org/std/fs/fn.read_dir.html

Same for the copy command

callicles (commented):
Small comment

callicles (approved):
NVM I miss read the code, sorry

PR #658, opened 2024-03-22
Nicolas/514 651 kafka subscriber is really slow

callicles:
This adds batching on the clickhouse side to improve speed.

### DIFF
--- a/apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
+++ b/apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
@@ -56,7 +56,7 @@ async fn flush(
 
         // in case the buffer is too big, we slice it and then we flush it
         for chunk in buffer_owned.chunks(MAX_BATCH_SIZE) {
-            match client.insert(&table, &columns, &chunk).await {
+            match client.insert(&table, &columns, chunk).await {
                 Ok(_) => {
                     debug!("Inserted {} records", chunk.len());
                 }
--- a/apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
+++ b/apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
@@ -1,0 +1,73 @@
+use crate::infrastructure::olap::clickhouse::client::ClickHouseClient;
+use crate::infrastructure::olap::clickhouse::model::ClickHouseRecord;
+use std::time::Duration;
+
+use log::{debug, error};
+use std::sync::Arc;
+use tokio::sync::Mutex;
+use tokio::time;
+
+use super::config::ClickHouseConfig;
+
+const MAX_FLUSH_INTERVAL_SECONDS: u64 = 1;
+const MAX_BATCH_SIZE: usize = 1000;
+
+pub type BatchRecords = Arc<Mutex<Vec<ClickHouseRecord>>>;
+
+pub struct Inserter {
+    buffer: BatchRecords,
+}
+
+// TODO Add at least once delivery guarantees
+impl Inserter {
+    pub fn new(clickhouse_config: ClickHouseConfig, table: String, columns: Vec<String>) -> Self {
+        let buffer = Arc::new(Mutex::new(Vec::new()));
+
+        tokio::spawn(flush(clickhouse_config, buffer.clone(), table, columns));
+
+        Self { buffer }
+    }
+
+    pub async fn insert(&self, record: ClickHouseRecord) -> anyhow::Result<()> {
+        let mut buffer: tokio::sync::MutexGuard<'_, Vec<ClickHouseRecord>> =
+            self.buffer.lock().await;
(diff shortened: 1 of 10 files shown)


apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use log::{debug, error};
+use std::sync::Arc;
+use tokio::sync::Mutex;
+use tokio::time;
+
+use super::config::ClickHouseConfig;
+
+const MAX_FLUSH_INTERVAL_SECONDS: u64 = 1;
+const MAX_BATCH_SIZE: usize = 1000;
```
cjus, line 13: Should these values should be in the clickhouse connection config and available for population via environment variable overrides in cloud product?  Also we need to document the tradeoffs of tuning those values.

apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        // in case the buffer is too big, we slice it and then we flush it
+        for chunk in buffer_owned.chunks(MAX_BATCH_SIZE) {
+            match client.insert(&table, &columns, chunk).await {
+                Ok(_) => {
+                    debug!("Inserted {} records", chunk.len());
+                }
+                Err(e) => {
+                    error!("Error inserting records: {:?}", e);
```
cjus, line 64: What's the effect of the insert failures? 
Would it be useful to output a record id in the error message?

apps/framework-cli/src/infrastructure/stream/redpanda.rs
```diff
@@ -160,6 +160,8 @@ pub fn create_subscriber(config: &RedpandaConfig, group_id: &str, topic: &str) -
     client_config
         .set("session.timeout.ms", "6000")
         .set("enable.partition.eof", "false")
+        .set("enable.auto.commit", "true")
+        .set("auto.commit.interval.ms", "1000")
```
cjus, line 164: Do these values need to be set via environment variables for cloud use?

apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
```diff
@@ -0,0 +1,73 @@
...
+    let mut interval = time::interval(Duration::from_secs(MAX_FLUSH_INTERVAL_SECONDS));
+
+    let client = ClickHouseClient::new(&clickhouse_config).unwrap();
+
+    loop {
+        interval.tick().await;
+        let mut buffer_owned = buffer.lock().await;
+        if buffer_owned.is_empty() {
+            drop(buffer_owned);
+            continue;
```
cjus, line 54: Is this block safe? There seems to be a potential for an endless loop.

cjus (commented):
I asked a few questions. I'm still learning Rust so some of my question/points might not be valid.  Overall this looks like an amazing start!

apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
callicles: That's a good question, from reading the docs it looks like the parameters from a clickhouse perspective are constrained, they are saying you should not make an insert more than 1 / sec and that the batches should be rather big. 

We could make it parametrizable with sane defaults but I wonder if a user would potentially just shoot themselves in a foot. Since we are trying to make it dead simple, unless absolutely necessary, I feel like we should hide parameters as much as possible.

I will leave it like this for now and we can change if we need to

apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
callicles: 🤔 Good suggestion about the id, I will look into adding something else. 

Right now an insert failure means we loose the batch. We will need to add a retry mechanism and potentially a dead letter queue.

apps/framework-cli/src/infrastructure/stream/redpanda.rs
callicles: Hum no - I believe they would stay the same, the 1000 value is synced with the `flush` frequency on the clickhouse client side

apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
callicles: It is an endless loop 😄 

The idea here is that we flush the buffer every 1 second. 
We spawn the loop in a different task to make sure we don't block the current green thread. 

The flowing code will block the loop for about 1s before letting it go to the next statement.
```rust 
interval.tick().await;
```

Which means the loop is executed every second or so.

cjus (approved):
LGTM

PR #659, opened 2024-03-22
new landing page

tg339:


apps/framework-landing/README.md
```diff
@@ -0,0 +1,36 @@
...
+## Getting Started
+
+First, run the development server:
+
+```bash
+npm run dev
+# or
+yarn dev
+# or
+pnpm dev
```
callicles, line 8: Maybe a quick edit just to have `pnpm` here since that's what we are using

apps/framework-landing/app/sections/FooterSection.tsx
```diff
@@ -0,0 +1,83 @@
...
+};
+
+export const FooterDisclaimerContainer = () => {
+  return (
+    <div className="flex flex-row grow min-h-16">
+      <div className="bg-primary aspect-square h-full min-h-16"></div>
+      <div className="flex flex-col justify-center px-5 no-wrap">
+        <SmallText className="my-0">2024 All rights rserved</SmallText>
+        <SmallText className="my-0">
+          Moose.js is by the folks at fiveonefour
```
callicles, line 69: maybe worth a link to the fiveonefour page?

apps/framework-landing/app/sections/FooterSection.tsx
```diff
@@ -0,0 +1,83 @@
...
+}) => {
+  return (
+    <Link href={item.href} className={cn("text-foreground", className)}>
+      <SmallText> {children} </SmallText>
+    </Link>
+  );
+};
+
+export const FooterNav = () => {
+  const navigation = [
```
callicles, line 37: This looks like could be reused with the `Nav` components if we want to keep things consistent

apps/framework-landing/public/vercel.svg
callicles, line 1: Needed?

apps/framework-landing/public/next.svg
callicles, line 1: Needed?

callicles (approved):
Nice work - Love how clean the components are with the hierarchy, feels ripe for use with a CMS!

Some small comments that could be addressed later if necessary. 

Is `framework-landing` replacing web?

apps/framework-landing/app/sections/FooterSection.tsx
tg339: No, they are different navs

tg339: As in they behave very differently

apps/framework-landing/app/sections/FooterSection.tsx
tg339: will address in next PR

tg339:
Yes framework landing will replace web

tg339:
I'll sequence the deletion though. I need to test the new instrumentation and I wanted both live and then to do a cutoff

PR #689, opened 2024-03-24
improvement: change default docker image tag in the build script to be latest

callicles:
This will help with automation for deploying. We can now do:

```
docker tag moose-df-deployment-x86_64-unknown-linux-gnu: latest moose-df-deployment-x86_64-unknown-linux-gnu:<versionChosenByTheUser>
```

for all builds, while before we had to dynamically get the version that was used to run the build to specify the tag.


cjus (approved):
LGTM

PR #690, opened 2024-03-24
Add retention.ms as a config

DatGuyJonathan:
https://docs.redpanda.com/current/reference/topic-properties/#retentionms

Flow service consumes message from topic & runs that message through a flow function. If the flow function finishes after `retention.ms` has elapsed, we try to mark the offset but the broker doesn't like it:

> {"level":"ERROR","timestamp":"2024-03-21T22:20:59.191Z","logger":"kafkajs","message":"[Connection] Response Fetch(key: 1, version: 11)","broker":"redpanda:9092","clientId":"deno-consumer","error":"The requested offset is not within the range of offsets maintained by the server","correlationId":6,"size":82}

Users' flow functions could have different exec times, so this allows them to configure the topic setting.

### DIFF
--- a/apps/framework-cli/src/framework/transform.ts
+++ b/apps/framework-cli/src/framework/transform.ts
@@ -14,6 +14,7 @@ CompressionCodecs[CompressionTypes.Snappy] = SnappyCodec;
 const cwd = Deno.args[0] || Deno.cwd();
 const FLOWS_DIR_PATH = `${cwd}/app/flows`;
 const FLOW_FILE = "flow.ts";
+const CONSUMER_ID = "deno-group";
 
 const getVersion = (): string => {
   const version = JSON.parse(Deno.readTextFileSync(`${cwd}/package.json`))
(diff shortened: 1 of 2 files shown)


apps/framework-cli/src/infrastructure/stream/redpanda.rs
```diff
@@ -24,13 +24,14 @@ pub async fn create_topics(
     // Prepare the AdminOptions
     let options = AdminOptions::new().operation_timeout(Some(std::time::Duration::from_secs(5)));
 
+    let retention_ms = config.clone().retention_ms.to_string();
```
phiSgr: Probably can do without the `clone` here.

apps/framework-cli/src/framework/transform.ts
```diff
@@ -102,20 +107,27 @@ const handleMessage = async (
...
         console.log(`Sent transformed data to ${destination}`);
-      } else {
-        transaction.abort();
       }
+    }
 
-      resolveOffset(message.offset);
-    } catch (error) {
+    if (didTransform) {
+      await transaction.sendOffsets({
```
phiSgr, line 115: looks like now we have exactly-once processing? This is too great to not be in the title.

phiSgr (approved):

apps/framework-cli/src/infrastructure/stream/redpanda.rs
DatGuyJonathan: Yes! Will update.

apps/framework-cli/src/framework/transform.ts
DatGuyJonathan: Haha, I appreciate you adding the TODO there & providing guidance.

PR #711, opened 2024-03-26
don't drop table on restart

cjus:


phiSgr:
Do we want to handle the case when the existing table is different from the model definition?

cjus:
> Do we want to handle the case when the existing table is different from the model definition?
I think we will want to - but probably not at the moment. @callicles what do you think?

apps/framework-cli/src/framework/controller.rs
```diff
@@ -198,11 +198,6 @@ pub(crate) async fn create_or_replace_tables(
             )
         })?;
 
-    // Clickhouse doesn't support dropping a view if it doesn't exist, so we need to drop it first in case the schema has changed
-    if !PROJECT.lock().unwrap().is_production {
-        drop_tables(fo, configured_client).await?;
-    }
```
callicles, line 204: Hum you do need to drop the tables if the data models have changed in between restarts

callicles:
I believe we need to. 
Current behavior is not ideal but it is working.

if we remove the drop, and the user changes something I believe it will look buggy. 

The edge case we won't be able to go for without the state management is if the user deletes a model all together

cjus:
> if we remove the drop, and the user changes something I believe it will look buggy.

I tested a running Moose instance where I sent data to UserActivity_0_0
I then shutdown Moose and edited the data model, package.json and project.toml
When I restarted Moose I had a new UserActivity_0_1 and no data loss in UserActivity_0_0.
So nothing looked buggy at that point. To me it looked like we're ready for a data migration.

apps/framework-cli/src/framework/controller.rs
cjus: That didn't seem to be the case for me.  See my earlier comment.
https://github.com/514-labs/moose/pull/711#issuecomment-2021143183

callicles:
What about the scenario where:

1. You send data to `UserActivity_0_0`
2. shut down the instance
3. change the schema of  UserActivity
4. spin back up the instance
5. send data to `UserActivity_0_0`

?

PR #713, opened 2024-03-26
restart kafka syncing when columns changed

phiSgr:


callicles (approved):
Can you add a comment - took me a min to understand why

PR #716, opened 2024-03-26
chore: make sure the default git head is main

callicles:
I noticed through running the init on my system that it was defaulting to `master` instead of `main` the new standard.

This makes sure we don't mistakenly create with the old convention


cjus (approved):
LGTM

PR #731, opened 2024-03-27
CLI command to initialize a new flow

DatGuyJonathan:
Help messages & the command result:

Screenshot 2024-03-27 at 12 55 34 PM

Error cases:

Screenshot 2024-03-27 at 12 49 16 PM

### DIFF
--- a/apps/framework-cli/src/cli/commands.rs
+++ b/apps/framework-cli/src/cli/commands.rs
@@ -60,6 +60,31 @@ pub enum Commands {
     Stop {},
     // Clears all temporary data and stops development infrastructure
     Clean {},
+    /// TODO: Add description for flow command
+    Flow(FlowArgs),
+}
+
+#[derive(Debug, Args)]
+#[command(arg_required_else_help = true)]
+pub struct FlowArgs {
+    #[command(subcommand)]
+    pub command: Option<FlowCommands>,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum FlowCommands {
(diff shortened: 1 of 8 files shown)


apps/framework-cli/src/cli/commands.rs
```diff
@@ -65,3 +67,28 @@ pub struct GenerateArgs {
 pub enum GenerateCommand {
     Migrations {},
 }
+
+#[derive(Debug, Args)]
+#[command(arg_required_else_help = true)]
+pub struct FlowArgs {
+    #[command(subcommand)]
+    pub command: Option<FlowCommands>,
```
DatGuyJonathan, line 82: @okane16 & I are leaning towards `moose flow init` instead of `moose generate flow` because we could have more commands in the future, like `moose flow test/run`, so putting them all under `flow` means less context switching. We could be convinced otherwise though, so please let us know if anyone feels strongly for putting this under `generate`.

callicles: Yeah I guess what we need to settle on is the architecture of the cli for consistency. 

Do we do 

```
moose <action> <object>
```

or 

```
moose <object> <action>
```

kubectl went:
```
kubectl get pods
```

callicles: After a quick inspection it looks like we went action as the top level. We could change but then we should change all of it for consistency purposes

callicles (approved):
Looks like this is blocking for a demo. We can change later if we want.

PR #740, opened 2024-03-28
Update instrumentation url & data model

DatGuyJonathan:
This creates a new model instead of using UserActivity so folks can continue to send test data to UserActivity without adding noise to instrumentation. There will be a follow-up PR to actually enable instrumentation after I verify the new model gets deployed.

### DIFF
--- a/apps/framework-cli/src/utilities/capture.rs
+++ b/apps/framework-cli/src/utilities/capture.rs
@@ -36,7 +36,7 @@ pub enum ActivityType {
 }
 
 #[derive(Debug, Clone, Serialize)]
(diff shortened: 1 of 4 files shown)


apps/framework-cli/src/utilities/capture.rs
```diff
@@ -17,26 +17,26 @@ use uuid::Uuid;
 
 #[derive(Debug, Clone, Serialize)]
 pub enum ActivityType {
```
DatGuyJonathan, line 19: Just re-ordering alphabetically for better readability.

callicles (approved):

PR #753, opened 2024-03-29
init data load

phiSgr:


cjus (approved):
@phiSgr this looks good to me.

PR #770, opened 2024-03-30
Update docker build to not depend on moose init

DatGuyJonathan:
The internal moose app doesn't have the right prisma file, even though we've updated framework-internal-app. I suspect we copy over the moose project to the docker image, but then it runs `moose init`. The prisma file from `init` doesn't & shouldn't have the same models as the one in our internal moose app. And it overwrites the one from framework-internal-app.


cjus (approved):
LGTM

PR #785, opened 2024-04-01
Update internal moose to use latest cli

DatGuyJonathan:
There was a CLI fix in https://github.com/514-labs/moose/releases/tag/v0.3.201 that's supposed to build the internal moose app as-is. But the latest deploy shows a data model is still missing. I think we need to bump the CLI version that internal moose app builds with.

### DIFF
--- a/apps/framework-internal-app/package.json
+++ b/apps/framework-internal-app/package.json
@@ -3,10 +3,10 @@
   "version": "0.0",
   "scripts": {
     "dev": "moose-cli dev",
-    "build": "moose-cli build --docker"
+    "build": "moose-cli --version && moose-cli build --docker"
   },
   "dependencies": {},
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


phiSgr (approved):

apps/framework-internal-app/package.json
```diff
@@ -3,10 +3,10 @@
   "version": "0.0",
   "scripts": {
     "dev": "moose-cli dev",
-    "build": "moose-cli build --docker"
+    "build": "moose-cli --version && moose-cli build --docker"
   },
   "dependencies": {},
   "devDependencies": {
-    "@514labs/moose-cli": "0.3.199"
```
callicles, line 10: This is not going to bump the version unless you do a `pnpm install` the lock file has not changed in this PR.

That's why I set the version explicitly here is to emphasize that this needs to go through a regular update process to get updated

PR #791, opened 2024-04-01
Send instrumentation to MooseActivity

DatGuyJonathan:


apps/framework-cli/src/utilities/capture.rs
```diff
@@ -57,39 +59,38 @@ macro_rules! capture {
...
-            cli_version: constants::CLI_VERSION.to_string(),
-        });
-        let remote_url = {
-            let guard = PROJECT.lock().unwrap();
-            guard.instrumentation_config.url().clone()
-        };
-
-        // Sending this data can fail for a variety of reasons, so we don't want to
-        // block user & no need to handle the result
-        tokio::spawn(async move {
```
DatGuyJonathan, line 78: I originally wanted to send the data in a spawned thread, but the main thread can actually finish faster and kill the spawned thread. So I just added a quick timeout to the request. Let me know if there's a better way we can spray-and-pray (a.ka. send the data without blocking user & without checking on the result).

apps/framework-cli/deploy/kubernetes/config-files/moose-deployment.yaml
```diff
@@ -107,5 +107,7 @@ spec:
                   key: sk-moose-http-server-config--port
             - name: MOOSE_LOGGER_STDOUT
               value: "true"
+            - name: MOOSE_INTERNAL
+              value: "true"
```
DatGuyJonathan: I noticed a lot of `moose prod` commands when we were still sending to UserActivity. I think it's coming from our internal moose, so figured we can add an env variable to not collect internal activities. We can probably use the same env var on our local dev envs.

apps/framework-cli/src/utilities/capture.rs
```diff
@@ -57,39 +59,38 @@ macro_rules! capture {
...
+                timestamp: Utc::now(),
+                cli_version: constants::CLI_VERSION.to_string(),
             });
+            let remote_url = {
+                let guard = PROJECT.lock().unwrap();
+                guard.instrumentation_config.url().clone()
+            };
+
+            let instrumentation_url = format!("{}/ingest/MooseActivity", remote_url);
+            println!("Sending to: {}. Event: {}", &instrumentation_url, &event);
```
callicles: Leftover print? if not could use the `log` crate instead

apps/framework-cli/src/utilities/capture.rs
```diff
@@ -57,39 +59,38 @@ macro_rules! capture {
...
-            // TODO: Change this to MooseActivity after the table is verified
-            let instrumentation_url = format!("{}/ingest/UserActivity", remote_url);
-            // TODO: Delete this, use event from above instead
-            let fake_event = json!({
-                "eventId": "1234",
-                "userId": "123456",
-                "activity": $activity_type,
-                "timestamp": Utc::now(),
+        // Ignore our deployments & internal testing
+        let moose_internal = env::var("MOOSE_INTERNAL");
```
callicles: Not sure if you know about this:

We have the ability to add settings at the cli level. In `apps/framework-cli/src/cli/settings.rs` The settings loaded there are using a hierarchy of env > files so that we could configure locally either env variables or configuration.

As I am writing this I now remember a discussion where we talked about letting the user opt out. 

For those reasons, having it in the cli settings with something like `enable_telemetry` that defaults to `true` might be good.

DatGuyJonathan: Gotcha. There was a project config, but it makes more sense to move that to the cli level.

callicles (approved):
More functionality, less code 

Booyah!

PR #815, opened 2024-04-02
Nicolas/514 718 restore sentry in the console

callicles:
* Add conditional to only instrument cli in production
* Fix the dev version dependencies for the npm package so that we don't have `pnpm.lock` file updates all the time
* Remove Analytics that is only available on vercel
* Adds prettier to the pre-commit hook and an empty prettier config file as a reference for all tools


apps/moose-cli-npm/package.json
```diff
@@ -23,9 +23,9 @@
     "typescript": "^4.9.5"
   },
   "optionalDependencies": {
-    "@514labs/moose-cli-darwin-arm64": "latest",
-    "@514labs/moose-cli-darwin-x64": "latest",
-    "@514labs/moose-cli-linux-arm64": "latest",
-    "@514labs/moose-cli-linux-x64": "latest"
+    "@514labs/moose-cli-darwin-arm64": "0.3.205",
```
cjus, line 26: Is the version set elsewhere? I expect this file and values would be updated by the CI/CD process?

apps/framework-cli/src/main.rs
```diff
@@ -14,14 +14,16 @@ fn main() {
...
-        ("https://83941f36fe439ebe1ffd4d35b834ed7a@o4505851966128128.ingest.sentry.io/4505851967963136", 
-        sentry::ClientOptions {
-        release: sentry::release_name!(),
-        traces_sample_rate: 1.0,
-        environment: Some(envionment.into()),
-        ..Default::default()
-      }));
+    if envionment == "production" {
+        let _guard = sentry::init(
+            ("https://83941f36fe439ebe1ffd4d35b834ed7a@o4505851966128128.ingest.sentry.io/4505851967963136",
```
cjus, line 19: Is exposing this URL a security issue? Could a bad actor mess with our metric?

cjus (commented):
Some questions but generally LGTM

apps/moose-cli-npm/package.json
callicles: It is! this is mostly for the dev setup.

The script that bumps those versions during the release process can be found here: https://github.com/514-labs/moose/blob/main/apps/moose-cli-npm/scripts/release-cli.sh

apps/framework-cli/src/main.rs
callicles: Yeah - I have been thinking about that. Unfortunatly this is software that runs anonymously on people's laptop. As such we need to be able to get metrics anonymously. We could make it a bit harder by embedding it at build time. But if someone is motivated, they will be able to get that back from the binary anyway.

cjus (approved):
Thanks for the replies. LGTM

PR #826, opened 2024-04-02
docs for generate migrations and bump-version

phiSgr:


phiSgr:
Got the branch name wrong, see #827.

PR #852, opened 2024-04-03
Use env vars for flow's redpanda conn settings

DatGuyJonathan:
Next step is to update `moose prod` command to run deno.

### DIFF
--- a/apps/framework-cli/src/cli/routines/docker_packager.rs
+++ b/apps/framework-cli/src/cli/routines/docker_packager.rs
@@ -10,7 +10,7 @@ use std::sync::Arc;
 static DOCKER_FILE: &str = r#"
 # Created from docker_packager routine
 
-ARG DENO_VERSION=1.42.1
+ARG DENO_VERSION=1.41.2
 FROM denoland/deno:bin-$DENO_VERSION AS deno
 
 FROM debian:bookworm-slim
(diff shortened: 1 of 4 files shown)


apps/framework-cli/src/cli/routines/docker_packager.rs
```diff
@@ -8,11 +8,17 @@ use std::fs;
...
 # Created from docker_packager routine
 
+ARG DENO_VERSION=1.41.2
+FROM denoland/deno:bin-$DENO_VERSION AS deno
+
+FROM debian:bookworm-slim
+
 ARG DEBIAN_FRONTEND=noninteractive
 
+COPY --from=deno /deno /usr/local/bin/deno
```
DatGuyJonathan, line 20: Deno will run in the same container as moose.

apps/framework-cli/src/framework/transform.ts
```diff
@@ -26,9 +44,28 @@ const scrubVersionFromTopic = (topic: string): string => {
...
+      return undefined;
+  }
+};
+
 const kafka = new Kafka({
   clientId: "deno-consumer",
-  brokers: ["redpanda:9092"],
+  brokers: [BROKER],
+  ssl: SECURITY_PROTOCOL === "SASL_SSL",
+  sasl: getSaslConfig(),
```
DatGuyJonathan, line 68: Tested this locally without sasl. Also remotely by connecting to our internal Redpanda cluster.

callicles (approved):
LGTM

PR #860, opened 2024-04-03
move create dir to CLI

phiSgr:


apps/framework-cli/src/cli.rs
```diff
@@ -101,13 +101,32 @@ async fn top_command_handler(settings: Settings, commands: &Commands) {
...
+                    "Running init command with name: {}, language: {}, location: {:?}, template: {:?}",
                     name, language, location, template
                 );
 
-                let dir_path = Path::new(location);
+                let dir_path =
+                    Path::new(
+                        location
+                            .as_deref()
+                            .unwrap_or(if *reinit { "." } else { name }),
```
phiSgr: If `--reinit`, defaults to current directory; otherwise defaults to a new directory of project name.

not sure if this condition is confusing

callicles (commented):
Honestly I don't think this is going to improve the experience significantly. 

The parsing / erroring should be done in the wrapper so that we don't see references to the CLI lower.

It would duplicate things but at the benefits of a much cleaner experience where the moose cli is not referenced

apps/framework-cli/src/cli/commands.rs
```diff
@@ -18,12 +18,16 @@ pub enum Commands {
         language: SupportedLanguages,
 
         /// Location of your app or service
-        #[arg(default_value = ".")]
-        location: String,
+        #[arg(short, long)]
```
phiSgr, line 21: The named parameter `--location` in the docs currently does not work.

```
error: unexpected argument '--location' found

  tip: to pass '--location' as a value, use '-- --location'
```

callicles: Good catch!

callicles (approved):

PR #873, opened 2024-04-05
wip

cjus:


apps/framework-cli/Cargo.toml
```diff
@@ -36,6 +36,7 @@ uuid = { version = "1.6", features = ["v4"] }
 serde_json = "1.0.108"
 serde_urlencoded = "0.7"
 base64 = "0.22.0"
+crypto-hash = "0.3.4"
```
callicles, line 39: It looks like this package might not be maintained, last release was 3 years ago.

I would advise this one instead: https://crates.io/crates/sha2

apps/framework-cli/src/framework/controller.rs
```diff
@@ -56,6 +111,8 @@ pub fn framework_object_mapper(
             s.name.as_str(),
         ),
         original_file_path: original_file_path.to_path_buf(),
+        version: version.to_string(),
+        schema_hash: String::new(),
```
callicles, line 115: I am a bit confused, if this is can be empty - should the type here of `schema_hash` be `optional`?

Since we have different ways to compute the hash should we even cache it?

PR #930, opened 2024-04-09
git actions should support repo root outside of moose

phiSgr:
I tried to run `moose bump-version` in `framework-internal-app` and failed.

### DIFF
--- a/apps/framework-cli/src/utilities/git.rs
+++ b/apps/framework-cli/src/utilities/git.rs
@@ -61,7 +61,7 @@ pub fn dump_old_version_schema(
     commit_hash: String,
     dest: &Path,
 ) -> Result<(), Error> {
-    let repo = Repository::open(project.project_location.clone())?;
+    let repo = Repository::discover(project.project_location.clone())?;
 
     let path = project.schemas_dir();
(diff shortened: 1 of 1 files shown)


callicles (approved):

apps/framework-cli/src/utilities/git.rs
```diff
@@ -61,18 +61,23 @@ pub fn dump_old_version_schema(
...
     dest: &Path,
 ) -> Result<(), Error> {
-    let repo = Repository::open(project.project_location.clone())?;
+    let repo = Repository::discover(project.project_location.clone())?;
 
-    let path = project.schemas_dir();
-    let path = path.strip_prefix(&project.project_location).unwrap();
+    let schema_dir = project.schemas_dir();
+    let schema_relative_path_from_repo_root = schema_dir
+        .strip_prefix(repo.path().parent().unwrap())
```
callicles, line 68: Are you confident with the `unwrap` here? looks like the method returns a `Result` so you could use `?` instead

phiSgr: `repo.path()` is the `.git`, so it will always have the parent.

PR #932, opened 2024-04-09
feat: add moose_developer field

callicles:
* Testing the migration in our dog food
* Adding a property to our instrumentation to distinguish between our devs and external users

### DIFF
--- a/apps/framework-cli/src/cli/settings.rs
+++ b/apps/framework-cli/src/cli/settings.rs
@@ -34,11 +34,21 @@ impl Default for Features {
 #[derive(Deserialize, Debug)]
 pub struct Telemetry {
     pub enabled: bool,
+
+    #[serde(default = "default_moose_developer")]
+    pub moose_developer: bool,
+}
+
+fn default_moose_developer() -> bool {
+    false
 }
 
 impl Default for Telemetry {
     fn default() -> Self {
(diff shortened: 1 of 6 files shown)


apps/framework-cli/src/cli/settings.rs
```diff
@@ -34,11 +34,21 @@ impl Default for Features {
 #[derive(Deserialize, Debug)]
 pub struct Telemetry {
     pub enabled: bool,
+
+    #[serde(default = "default_moose_developer")]
+    pub moose_developer: bool,
```
phiSgr: i prefer calling it `is_moose_developer`

apps/framework-cli/src/cli/settings.rs
```diff
@@ -34,11 +34,21 @@ impl Default for Features {
 #[derive(Deserialize, Debug)]
 pub struct Telemetry {
     pub enabled: bool,
+
+    #[serde(default = "default_moose_developer")]
```
phiSgr: The `Default::default` of `bool` is `false`, I don't think we need this func.

callicles: I will try

callicles: You are right!

phiSgr (approved):

PR #958, opened 2024-04-12
wip: typescript instead of prisma

DatGuyJonathan:
This is a copy-pasta from https://github.com/514-labs/moose/pull/549, minus doc changes & pr feedback changes. The goal is to get something merged to main in case we need to swarm on it.

I added an if-else to `parse_schema_file` so it can continue to work with current prisma schema. Next step is to iron out the wrinkles. Need to look into feedback from the original pr, and also what happens if we change `models.prisma` to `models.ts` with these interfaces.

```
interface UserActivity {
    eventId: string;
    timestamp: Date;
    userId: string;
    activity: string;
}

interface ParsedActivity {
    eventId: string;
    timestamp: Date;
    userId: string;
    activity: string;
}
```

### DIFF
--- a/apps/framework-cli/src/framework/schema.rs
+++ b/apps/framework-cli/src/framework/schema.rs
@@ -25,6 +25,7 @@ use std::{
 use crate::framework::controller::FrameworkObject;
 use diagnostics::Diagnostics;
 
+use log::debug;
 use schema_ast::ast::{Enum, Model};
 use schema_ast::{
     ast::{Attribute, Field, SchemaAst, Top, WithName},
@@ -34,6 +35,13 @@ use serde::Serialize;
 
 use crate::infrastructure::olap::clickhouse::mapper::arity_mapper;
 use crate::project::PROJECT;
+
+use swc_common::{self, sync::Lrc, SourceMap};
+use swc_ecma_ast::{
+    Decl, Expr, Module, ModuleItem, Stmt, TsEnumDecl, TsEnumMember, TsEnumMemberId,
+    TsInterfaceDecl, TsKeywordTypeKind, TsType, TsTypeAnn, TsTypeRef,
+};
+use swc_ecma_parser::{lexer::Lexer, Capturing, Parser, StringInput, Syntax};
 
 pub mod templates;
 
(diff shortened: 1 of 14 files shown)


apps/framework-cli/src/framework/schema.rs
```diff
@@ -378,10 +408,152 @@ pub fn ast_mapper(ast: SchemaAst) -> Result<FileObjects, ParsingError> {
...
+        ModuleItem::Stmt(Stmt::Decl(Decl::TsInterface(decl))) => {
+            ts_declarations.push(decl);
+        }
+        ModuleItem::Stmt(Stmt::Decl(Decl::TsEnum(decl))) => {
+            enums.push(ts_enum_to_data_enum(decl));
+        }
+        _ => {}
+    });
+
+    println!(
```
callicles: Can we replace by `info!`?

apps/framework-cli/src/framework/schema.rs
```diff
@@ -378,10 +408,152 @@ pub fn ast_mapper(ast: SchemaAst) -> Result<FileObjects, ParsingError> {
...
+        .filter_map(|field| match field {
+            swc_ecma_ast::TsTypeElement::TsPropertySignature(prop) => Some({
+                // match the key's sym if it's an ident
+                let name = match *prop.key.clone() {
+                    Expr::Ident(ident) => ident.sym.to_string(),
+                    _ => return None,
+                };
+                // match the type of the value and return the right column type
+                let TsTypeAnn { type_ann, .. } =
+                    *prop.type_ann.clone().expect("no type for property");
```
callicles, line 471: This is going to panic if the user fails to put a type. They will do that eventually, maybe we solidify this in a follow up PR with error handling.

callicles (approved):
Some small things - definitely some rough edges there but we should merge and iterate. nice work bringing this to the latest version

apps/framework-cli/src/framework/schema.rs
DatGuyJonathan: Nice catch, will update it.

apps/framework-cli/src/framework/schema.rs
DatGuyJonathan: Sounds good, I'll make a note of it.

PR #997, opened 2024-04-15
Nicolas/514 731 exit cli when docker is not running

callicles:
* Removes the coming soon banner, which is not used anymore anyway
* Refactors part of the CLI to not use Routines. Still, some to be refactored, but this at least helps move toward cleaner implementation. Now the cli will exit cleanly when docker is not running.


DatGuyJonathan (approved):
LGTM

PR #998, opened 2024-04-15
Show read-only list of flows in console

DatGuyJonathan:
Overview & detailed pages without flows:

console_noflows_overview

console_noflows_detailed

------------------------------------

Overview & detailed pages with flows: 

console_flows_overview

console_flows_detailed

### DIFF
--- a/apps/moose-console/src/components/overview-list.tsx
+++ b/apps/moose-console/src/components/overview-list.tsx
@@ -2,6 +2,15 @@ import OverviewCard from "components/overview-card";
 import { DataModel } from "app/types";
 
 export function PrimitivesOverviewList({ models }: { models: DataModel[] }) {
+  const flows = models
+    .filter((model) => model.flows && model.flows.length > 0)
+    .flatMap((model) =>
+      model.flows.map((flow) => ({
+        name: `${model.model.name} to ${flow}`,
+        link: "",
+      })),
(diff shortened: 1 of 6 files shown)


apps/moose-console/src/components/overview-list.tsx
```diff
@@ -20,9 +20,18 @@ export function PrimitivesOverviewList({ models }: { models: DataModel[] }) {
...
+              0,
+            )}
             link="/primitives/flows"
-            items={[]}
+            items={models
+              .filter((model) => model.flows && model.flows.length > 0)
+              .slice(0, 4)
+              .map((model) => ({
+                name: model.model.name,
+                link: "",
```
georgevanderson: Maybe here we make the name a concat of Source -> Destination instead of just the model

DatGuyJonathan: Ah, like this? What do you think @asl677 ?

Screenshot 2024-04-16 at 1 46 55 PM

DatGuyJonathan: Spoke with @asl677 and we landed on using the format {source} to {destination} like this:

Screenshot 2024-04-18 at 6 10 59 PM

georgevanderson (approved):

PR #1014, opened 2024-04-16
remove contains(SCHEMAS_DIR) check, as old versions don't have that dir

phiSgr:


callicles (approved):

apps/framework-cli/src/framework/schema.rs
```diff
@@ -106,11 +105,9 @@ impl Display for DuplicateModelError {
 impl std::error::Error for DuplicateModelError {}
 
 pub fn is_schema_file(path: &Path) -> bool {
-    path.to_string_lossy().contains(SCHEMAS_DIR)
```
DatGuyJonathan, line 109: IIRC, this would return true on flow files then eventually try to parse the flow.ts files. Do we need to exclude flow directory then?

phiSgr: Oh, I'll add this check in the watcher code then.

DatGuyJonathan: I think watcher is local only. Would it still try to parse flow.ts files in prod?

phiSgr: When initializing, the dir crawling has a correct base path to start.

PR #1030, opened 2024-04-17
Create README.md for product analytics instructions

okane16:


03cranec (commented):
Content looks good. My only concern is that when we start to duplicate documentation in different places, it gets really easy to forget to update all the places when things change. 

Given this is in the same repo as our docs, maybe there is a way to reference/pull the reference file from the docs to here? I think @callicles did something like this for another readme?

callicles:
Where is that duplicated from?

okane16:
> Where is that duplicated from?

RIght now its not duplicated from anywhere. It could be duplicated by the template detail page content on moosejs.com/templates/[template id] but currently that page does not use markdown

callicles:
Let's merge then

callicles (approved):

PR #1051, opened 2024-04-19
.ts migration files

phiSgr:


apps/framework-cli/src/framework/transform.ts
```diff
@@ -174,9 +208,21 @@ const handleMessage = async (
 
 const startConsumer = async (): Promise<void> => {
   const flows = getFlows();
-  const flowTopics = Array.from(flows.keys()).map(
-    (flow) => `${flow}_${version}`,
-  );
+  const migrations = getMigrations();
+
+  for (let [sourceTopic, vs] of migrations) {
```
DatGuyJonathan: This looks like it could be a `const` instead of `let` since they don't change.

DatGuyJonathan (approved):

apps/framework-cli/src/framework/transform.ts
phiSgr: good catch

PR #1066, opened 2024-04-19
lower case project name for docker compose

phiSgr:
```
invalid project name "MagicStuff": must consist only of lowercase alphanumeric characters, hyphens, and underscores as well as start with a letter or number
```


DatGuyJonathan (approved):

cjus (commented):
What if the user is expecting the project name to include uppercase letters as they entered them?

phiSgr:
This project name is only for the docker, the user usually don't have to see it.

PR #1095, opened 2024-04-23
Aggregations prototype

DatGuyJonathan:
Implemented the workflow defined here: https://www.figma.com/file/Eyi7IU5j9uYQAXD8H79UhI/Aggregations-and-Consumption?type=whiteboard&node-id=221-509&t=JzgHLuze4IKAamAP-4

Chatted with @okane16 and started a list of next steps: https://www.figma.com/file/Eyi7IU5j9uYQAXD8H79UhI/Aggregations-and-Consumption?type=whiteboard&node-id=274-3014&t=MHoNLOAZrhvK9cwJ-4. We'll turn them into linear tickets.

### DIFF
--- a/apps/framework-cli/src/framework/aggregations.ts
+++ b/apps/framework-cli/src/framework/aggregations.ts
@@ -11,8 +11,8 @@ interface ShowTablesResponse {
 }
 
 const cwd = Deno.args[0] || Deno.cwd();
-const AGGREGATES_DIR_PATH = `${cwd}/app/aggregates`;
-const AGGREGATE_FILE = "*.ts";
+const AGGREGATIONS_DIR_PATH = `${cwd}/app/aggregations`;
+const AGGREGATIONS_FILE = "*.ts";
 
 const CLICKHOUSE_DB =
   Deno.env.get("MOOSE_CLICKHOUSE_CONFIG__DB_NAME") || "local";
@@ -81,7 +81,7 @@ const getFileName = (filePath: string) => {
   return "";
 };
 
-const createAggregate = async (
+const createAggregation = async (
   chClient: ClickHouseClient,
   path: string,
   version: string,
@@ -92,7 +92,7 @@ const createAggregate = async (
     const sqlString = (await import(path)).default;
     if (typeof sqlString !== "string") {
       console.error(
-        `Not creating aggregate. Expected an export default SQL string from ${fileName}`,
+        `Not creating aggregation. Expected an export default SQL string from ${fileName}`,
       );
       return;
     }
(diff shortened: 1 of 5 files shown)


apps/framework-cli/src/framework/aggregations.ts
```diff
@@ -0,0 +1,164 @@
...
+  );
+  return createClient({
+    url: `${protocol}://${CLICKHOUSE_HOST}:${CLICKHOUSE_PORT}`,
+    username: CLICKHOUSE_USERNAME,
+    password: CLICKHOUSE_PASSWORD,
+    database: CLICKHOUSE_DB,
+  });
+};
+
+const waitForClickhouse = (chClient: ClickHouseClient) => {
```
DatGuyJonathan, line 44: Aggregations is basically materialized views that take data from a source table, aggregates stuff, and puts the result in a target table. Creating a materialized view depends on the source table to exist, so this aggregations service need to wait on something.

I tried ping & show tables, but they aren't 100% reliable. Looks like we already have a ticket for it: https://linear.app/514/issue/514-750/moose-started-faster-than-clickhouse-and-start-up-failed. FYI @phiSgr , lmk if I can help with anything.

apps/framework-cli/src/framework/aggregations.ts
```diff
@@ -0,0 +1,164 @@
...
+};
+
+const startFileWatcher = (chClient: ClickHouseClient) => {
+  const version = getVersion();
+  const pathToWatch = `${AGGREGATIONS_DIR_PATH}/**/${AGGREGATIONS_FILE}`;
+
+  watch(pathToWatch, { usePolling: true }).on(
+    "all",
+    async (event: string, path: string) => {
+      const antiCachePath = `${path}?num=${Math.random().toString()}&time=${Date.now()}`;
```
DatGuyJonathan, line 137: User could write an aggregation then we create a materialized view in clickhouse. If they change the sql string, we get a "change" event but it's going to import the same file path so it just uses what's already cached. Appending url params is a workaround: `https://github.com/denoland/deno/issues/13754#issuecomment-1908241494`

apps/framework-cli/src/utilities/docker-compose.yml
```diff
@@ -50,18 +50,43 @@ services:
...
-  deno:
+  flows:
     image: denoland/deno:1.41.2
     depends_on:
       - redpanda
     environment:
       - FILEWATCHER_ENABLED=${FILEWATCHER_ENABLED:-true}
       - MOOSE_REDPANDA_CONFIG__BROKER=${MOOSE_REDPANDA_CONFIG__BROKER:-redpanda:9092}
     volumes:
-      - ../:/home
```
DatGuyJonathan, line 61: Deno creates an `app/node_modules` folder in the container. The file paths in node_modules are only relevant for that container. Now that we have multiple deno containers, we don't want those node_modules volume to conflict so this change is just to be specific on what can be shared between deno containers.

callicles (approved):

PR #1118, opened 2024-04-26
new highlight message type for CLI init success message & console link fix

okane16:


tg339 (approved):
LGTM ship it!!!

apps/framework-cli/src/cli.rs
```diff
@@ -152,33 +152,48 @@ async fn top_command_handler(
...
                     }
 
-                    show_message!(
-                        MessageType::Success,
-                        Message {
-                            action: "Created".to_string(),
-                            details: "new project".to_string(),
-                        }
-                    );
+                    {
```
callicles, line 163: Why did you put it in a closure?

It doesn't feel like that should be necessary

callicles (commented):
Nice first `rust` PR! You could merge as is, curious about, what looks like extra closures around the `show_message!`

apps/framework-cli/src/cli.rs
okane16: @tg339 can better explain but it was basically the fastest workaround for avoiding duplicate TERM imports with the back to back !show_message macros

PR #1122, opened 2024-04-26
Clean up existing aggregations mv on startup

DatGuyJonathan:


apps/framework-cli/src/framework/aggregations.ts
```diff
@@ -81,6 +82,24 @@ const getFileName = (filePath: string) => {
...
 
+const cleanUpAggregations = async (chClient: ClickHouseClient) => {
+  try {
+    console.log("Preparing aggregations");
+    const query = `SHOW TABLES LIKE '%${AGGREGATIONS_MV_SUFFIX}'`;
+    const queryResponse = await chClient.query({ query });
+    const showTablesResponse =
+      (await queryResponse.json()) as ShowTablesResponse;
+
+    showTablesResponse.data.forEach(async (table) => {
```
DatGuyJonathan, line 93: Seems like there's wip to allow drops with a list `https://github.com/ClickHouse/ClickHouse/pull/62359`

callicles (approved):

callicles:
Is this only for dev mode?

DatGuyJonathan:
> Is this only for dev mode?

No, I was thinking this is needed for prod as well. Currently, versions are specified in these aggregation functions. If they want to update versions, we'd need to delete the mv and recreate them. They also need a way to delete an aggregation file locally, push that change, and deploy, and that mv should be deleted in prod.

PR #1126, opened 2024-04-27
Nicolas/514 624 support enum values

callicles:
* Adds Enum support from parsing in typescript to SDK generation
* Had to add handlebars as a dependency since tinytemplate was lacking some templating functionality. Will port over the over templates in a follow up PR
* Remove duplicate TS SDK generation
* Remove the ts_interfaces from the framework object
* Removes the postgres interface from the config we have not been using
* Exposes port 9000 on clickhouse for native drivers
* Increase batch size for the Kafka sync to 100K records


apps/framework-cli/src/framework/typescript/parser.rs
```diff
@@ -92,17 +92,37 @@ fn extract_data_models_from_ast(ast: Module) -> Result<FileObjects, TypescriptPa
...
+                    return Err(TypescriptParsingError::UnsupportedDataTypeError {
+                        type_name: format!("{:?}", init),
+                    })
+                }
+            },
+            None => {
+                return Err(TypescriptParsingError::UnsupportedDataTypeError {
+                    type_name: format!("Non mapped enums"),
+                })
+            }
```
phiSgr: Do we want to handle auto-increment. The following definition has values 2,3,4,5

```
enum Direction {
  Up = 2,
  Down,
  Left,
  Right,
}
```

Without `= 2`, it'll have values 0,1,2,3

callicles: I was struggling to have something reliable that works - but with some late changes might be easier.

callicles: the thing that's annoying is that a user could also do

```ts
enum Direction {
  Up,
  Down,
  Left = 10,
  Right,
}
```

Do you know what will be `Right`? 11 or 2?

phiSgr: It's 11 and I never doubted that. :D

We can have a counter variable that is incremented when there is no associated value, or is set when there is a value.

apps/framework-cli/src/cli/routines.rs
```diff
@@ -297,7 +297,7 @@ pub async fn start_production_mode(project: Arc<Project>) -> anyhow::Result<()>
     let mut route_table = HashMap::<PathBuf, RouteMeta>::new();
 
     info!("<DCM> Initializing project state");
-    let (framework_object_versions, version_syncs) =
+    let (framework_object_versions, versions_syncs) =
```
phiSgr: The extra plural "s" sounds very awkward to me.

apps/moose-console/src/app/types.ts
```diff
@@ -57,12 +57,18 @@ export interface DataModel {
...
+  values: MooseEnumMember[];
 }
 
+export interface MooseEnumMember {
+  name: string;
+  value?: MooseInt | MooseString;
+}
+
+export type MooseInt = { Int: number };
+export type MooseString = { String: string };
```
phiSgr, line 70: why the "tag" instead of simply using union type?

apps/moose-console/src/lib/utils.ts
```diff
@@ -56,6 +59,16 @@ export function tableIsQueryable(table: Table): boolean {
   return table.engine === "MergeTree";
 }
 
-export function is_enum(type: any): type is { Enum: any } {
-  return typeof type === "object" && type["Enum"] !== undefined;
+export function is_enum(type: string | MooseEnum): type is MooseEnum {
+  return typeof type === "object";
+}
+
+export function is_MooseInt(input: MooseInt | MooseString): input is MooseInt {
```
phiSgr: Maybe we should use camel case all the way.

apps/framework-cli/src/framework/typescript/parser.rs
```diff
@@ -92,17 +92,37 @@ fn extract_data_models_from_ast(ast: Module) -> Result<FileObjects, TypescriptPa
...
             TsEnumMemberId::Str(str) => str.value.to_string(),
-        })
-        .collect();
-    DataEnum { name, values }
+        };
+
+        let value = match member.init {
+            Some(init) => match *init {
+                Expr::Lit(Lit::Str(str)) => EnumValue::String(str.value.to_string()),
+                Expr::Lit(Lit::Num(num)) => EnumValue::Int(num.value as u8),
```
phiSgr: We should disallow mixing string enum value and int enum value.

apps/framework-cli/src/framework/typescript/parser.rs
```diff
@@ -92,17 +92,37 @@ fn extract_data_models_from_ast(ast: Module) -> Result<FileObjects, TypescriptPa
...
+                }
+            },
+            None => {
+                return Err(TypescriptParsingError::UnsupportedDataTypeError {
+                    type_name: "Non mapped enums".to_string(),
+                })
+            }
+        };
+
+        values.push(EnumMember { name, value });
```
phiSgr, line 152: If the enum values are strings, I don't think the `name` is meaningful anywhere outside of JS/TS. E.g. the name is irrelevant in JSON.

apps/moose-console/src/lib/utils.ts
callicles: Ha yeah I just followed what was above

apps/framework-cli/src/framework/typescript/parser.rs
callicles: Clickhouse uses it to map the values. Ref: https://clickhouse.com/docs/en/sql-reference/data-types/enum

apps/framework-cli/src/framework/typescript/parser.rs
callicles: Cool - I will add that

apps/framework-cli/src/cli/routines.rs
callicles: I don't have a strong feeling, will change back

apps/moose-console/src/app/types.ts
callicles: It feels like it results in more readable function signatures like this

apps/framework-cli/src/framework/typescript/parser.rs
phiSgr:
```
enum Direction {
  Up = "UP",
  Down = "DOWN",
  Left = "LEFT",
  Right = "RIGHT",
}
```

`Right` is just only in js/ts. In JSON it's `RIGHT`, so in clickhouse it should use the string values as well.

callicles: Yeah but in the case for numbers 

```
enum Direction {
  Up = 0,
  Down = 1,
  Left = 2,
  Right = 3,
}
```

it is better to have clickhouse with `Enum('Up' = 0, 'Down' = 1...)` that way the user has human readable labels

phiSgr (approved):

PR #1131, opened 2024-04-29
fix rfd script

phiSgr:


scripts/rfd/new.js
```diff
@@ -64,12 +64,12 @@ execSync(`git checkout -b "rfd/${rfdNumber}"`);
 // We read the template and hydrate it with the RFD number, title and author
 const template = fs.readFileSync(__dirname + "/template.mdx", "utf8");
 const hydratedTemplate = template
-  .replace("{{title}}", title)
-  .replace("{{authors}}", `${author} <${authorEmail}>`);
+  .replace("{ { title } }", title)
+  .replace("{ { authors } }", `${author} <${authorEmail}>`);
```
phiSgr, line 68: Alternatively we can prevent `lint-staged` from changing the braces, but changing these two lines is easier.

callicles (approved):

PR #1139, opened 2024-04-29
Pipe docker logs to cli logs

DatGuyJonathan:
Hopefully makes it easier to just check one place for logs. Did it for flows, aggregations, & consumption.

Screenshot 2024-04-29 at 2 09 24 PM


callicles (approved):
That looks awesome and really clean - could we use the same for the other containers?

DatGuyJonathan:
> That looks awesome and really clean - could we use the same for the other containers?

@callicles , I'll check it out. On first glance, there doesn't seem to be useful info in the console & clickhouse container logs, and redpanda container logs seem spammy, but I'll keep monitoring.

PR #1140, opened 2024-04-30
Tim/514 918 python data models

tg339:


apps/framework-cli/src/framework/python/parser.rs
```diff
@@ -0,0 +1,465 @@
...
+        _ => {
+            println!("failed parsing {:?} ", body_node);
+            Err(PythonParserError::UnsupportedDataTypeError {
+                type_name: "Unsupported data type".to_string(),
+            })
+        }
+    }
+}
+
+pub fn extract_data_model_from_file(path: &PathBuf) -> Result<(), PythonParserError> {
```
callicles: This is the only method that needs to public in this file right?

apps/framework-cli/src/framework/python/parser.rs
```diff
@@ -0,0 +1,465 @@
...
+            "int" => Ok(ColumnType::Int),
+            "float" => Ok(ColumnType::Float),
+            "bool" => Ok(ColumnType::Boolean),
+            "datetime" => Ok(ColumnType::DateTime),
+            _ => Err(PythonParserError::UnsupportedDataTypeError {
+                type_name: name.id.to_string(),
+            }),
+        },
+        _ => Err(PythonParserError::UnsupportedDataTypeError {
+            type_name: "Unsupported data type".to_string(),
```
callicles: Here you should put the name of the type that's missing. The sentence `Unsupported data Type` will be provided by the error string that you templated above.

apps/framework-cli/src/framework/python/parser.rs
```diff
@@ -0,0 +1,465 @@
...
+};
+
+#[derive(Debug, Clone, thiserror::Error)]
+#[error("Failed to parse the python file")]
+#[non_exhaustive]
+pub enum PythonParserError {
+    FileNotFound {
+        path: PathBuf,
+    },
+    #[error("Python Parser - Unsupported data type: {type_name}")]
```
callicles, line 36: Here you will print `Python Parser - Unsupported data type: Unsupported data type` if you don't specify the data type properly at the Error building site

apps/framework-cli/src/framework/python/parser.rs
callicles: Maybe here - the error should be a different one?

callicles (approved):
Small comments around error handling - otherwise looks good to me

apps/framework-cli/tests/python/simple.py
```diff
@@ -0,0 +1,26 @@
+from dataclasses import dataclass, field
+from enum import Enum
+from typing import List, Optional
+
+
+type Key[T: (str, int)] = T
```
phiSgr: I fed it to `mypy` and got `simple.py:6: error: invalid syntax  [syntax]`

phiSgr: This works

```python
from typing import List, Optional, TypeVar, Union
from typing_extensions import TypeAlias

T = TypeVar('T', bound= Union[str, int])
Key: TypeAlias = T
```

Or `bound=str | int` since python 3.10

apps/framework-cli/tests/python/simple.py
```diff
@@ -0,0 +1,26 @@
...
+    ERROR = "error"
+
+
+@dataclass 
+class MyModel:
+    name: Key[str]
+    age: int
+    flag: bool
+    status: Status
+    test_key: str = field(alias="test-key")
```
phiSgr:
```
simple.py:24: error: No overload variant of "field" matches argument type "str"  [call-overload]
simple.py:24: note: Possible overload variants:
simple.py:24: note:     def [_T] field(*, default: _T, init: bool = ..., repr: bool = ..., hash: Optional[bool] = ..., compare: bool = ..., metadata: Optional[Mapping[Any, Any]] = ...) -> _T
simple.py:24: note:     def [_T] field(*, default_factory: Callable[[], _T], init: bool = ..., repr: bool = ..., hash: Optional[bool] = ..., compare: bool = ..., metadata: Optional[Mapping[Any, Any]] = ...) -> _T
simple.py:24: note:     def field(*, init: bool = ..., repr: bool = ..., hash: Optional[bool] = ..., compare: bool = ..., metadata: Optional[Mapping[Any, Any]] = ...) -> Any
```

It looks like `alias` is not a thing yet?
https://github.com/python/cpython/issues/101192

apps/framework-cli/tests/python/simple.py
tg339: Yeah that's the old syntax. The above works in 3.12

tg339: We can add additional test cases and supported parsings. I'll keep 3.12 for now

apps/framework-cli/tests/python/simple.py
tg339: What python version are you using

apps/framework-cli/tests/python/simple.py
phiSgr: The python code runs, but mypy 1.10 on python 3.12 rejects it. Is there an alternative type checker?

apps/framework-cli/tests/python/simple.py
phiSgr:
```
python3.12 -m venv venv
source venv/bin/activate
pip install mypy
mypy simple.py
```

mypy version is 1.10.0

phiSgr:
```
python --version
Python 3.12.3
```

```
python simple.py
Traceback (most recent call last):
  File "/Users/george/stuff/simple.py", line 18, in <module>
    class MyModel:
  File "/Users/george/stuff/simple.py", line 23, in MyModel
    test_key: str = field(alias="test-key")
                    ^^^^^^^^^^^^^^^^^^^^^^^
TypeError: field() got an unexpected keyword argument 'alias'
```

apps/framework-cli/src/framework/python/parser.rs
```diff
@@ -0,0 +1,29 @@
+use std::path::PathBuf;
+
+use anyhow::Error;
+use rustpython_parser::{ast, Parse};
+
+pub fn extract_data_model_from_file(path: &PathBuf) -> Result<(), Error> {
+    let schema_file = std::fs::read_to_string(path)?;
+    let ast = ast::Suite::parse(&schema_file, "<embedded>");
+
+    println!("{:#?}", ast);
```
cjus: Is this println for debugging purposes?

cjus (approved):
LGTM

phiSgr:
Since the types are retained in Python runtime (just ignored by the interpreter), maybe we should, instead of interring the semantics from the syntax, import the user's file then do `typing.get_type_hints`.

apps/framework-cli/src/framework/python/scripts/framework_object_serializer.py
```diff
@@ -0,0 +1,73 @@
...
+    'dataclasses': {}
+}
+
+def serialize_enum(enum):
+    return {'name': enum.name, 'value': enum.value}
+
+def serialize_type(type_):
+    type_str = str(type_)
+    class_match = re.match(r"<class '(.*)'>", type_str)
+    enum_match = re.match(r"<enum '(.*)'>", type_str)
```
phiSgr, line 40: `.__name__`?

apps/framework-cli/src/framework/data_model/schema.rs
```diff
@@ -130,6 +138,12 @@ impl Serialize for ColumnType {
                 state.serialize_field("elementType", inner)?;
                 state.end()
             }
+            ColumnType::Nested(nested) => {
+                let mut state = serializer.serialize_struct("Nested", 2)?;
```
phiSgr, line 142: need to change `fn visit_map` as well
maybe we should add a tag to the JSON object.

apps/framework-cli/src/framework/python/scripts/framework_object_serializer.py
tg339: Yeah stupid copilot stuff here. It worked so didn't bother looking into the private APIs

apps/framework-cli/src/framework/data_model/schema.rs
tg339: What does tag mean in this case? A parent? A unique ID?

apps/framework-cli/src/cli.rs
```diff
@@ -512,3 +512,59 @@ pub async fn cli_run() {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn set_test_temp_dir() {
+        let test_dir = "tests/tmp";
+        std::fs::create_dir_all(test_dir).unwrap();
+        std::env::set_current_dir(test_dir).unwrap();
+    }
```
callicles, line 528: We have been using `assert_fs` to create temp dirs for io in tests.

You can check it out here: 
https://github.com/514-labs/moose/blob/99f8038247cb2fd1a2ecaae4aaeb109dbc6cd9d7/apps/framework-cli/tests/cli_init.rs#L20

apps/framework-cli/src/cli.rs
```diff
@@ -512,3 +512,59 @@ pub async fn cli_run() {
...
+        let data_model_path = project.app_dir().join("datamodels");
+
+        // Make sure all the data models are .py files
+        let data_model_files = std::fs::read_dir(data_model_path).unwrap();
+        for file in data_model_files {
+            let file = file.unwrap();
+            let file_name = file.file_name();
+            let file_name = file_name.to_str().unwrap();
+            assert!(file_name.ends_with(".py"));
+        }
```
callicles, line 588: Feels like we could move the test alongside the existing `init` test or bring the other one here. 

I would lean towards bringing this with the other one as they are both integration test (IO) and the setup and tear down on the integration side is currently shorter.

apps/framework-cli/src/framework/python/executor.rs
```diff
@@ -0,0 +1,53 @@
...
+pub enum PythonSerializers {
+    FrameworkObjectSerializer,
+    ProjectObjectSerializer,
+}
+
+impl PythonSerializers {
+    pub fn get_path(&self) -> &str {
+        match self {
+            PythonSerializers::FrameworkObjectSerializer => {
+                "src/framework/python/scripts/framework_object_serializer.py"
```
callicles, line 24: Hum this path is not going to be available at runtime in a user environment no?

What we currently do with ts is that we load the scripts at compile time in a static string and then write it on disk in the user's `.moose` directory when the CLI is run. 

We probably should do something similar here no? 

https://github.com/514-labs/moose/blob/99f8038247cb2fd1a2ecaae4aaeb109dbc6cd9d7/apps/framework-cli/src/framework/typescript/flow.rs#L10

apps/framework-cli/src/framework/python/parser.rs
```diff
@@ -0,0 +1,741 @@
+//! # Python Parser
```
callicles, line 1: Is it that if we use the rust parser here then we don't need `executor.rs` and `framework_object_seriliazer.py` as wells as `project_serializer.py`?

apps/framework-cli/src/project.rs
```diff
@@ -502,3 +536,55 @@ impl Project {
...
+
+    fn create_python_project() -> Project {
+        Project::new(
+            Path::new("tests/python/project"),
+            "test_project".to_string(),
+            SupportedLanguages::Python,
+        )
+    }
+
+    fn remove_python_project() {
```
callicles, line 553: Same as above - for those tests we should probably use `assert_fs` it will create isolated folders and gas all the setup and tear down logic in there

apps/framework-cli/tests/python/models/__pycache__/simple.cpython-312.pyc
callicles, line 1: This probably should not have been committed no?

callicles (commented):
I know this has been merged - but some small comments on how to improve the the code.

PR #1145, opened 2024-05-01
remove enums import when not needed

callicles:
* Removes the enums import in typescript generated code when the enums are not needed.
* Migrates the code generation for interfaces to handlebars


cjus (approved):
LGTM

PR #1150, opened 2024-05-02
RFD-0004: Diagram with regular flow and migration flow.

phiSgr:
Is the mermaid diagram worth it? IDK

### DIFF
--- a/rfd/0004/README.mdx
+++ b/rfd/0004/README.mdx
@@ -1,0 +1,144 @@
+---
+authors: George Leung <phisgr@gmail.com>
+title: Flows Input/Output Topics
+state: discussion
+---
+
+## Objective
+
+To unify the regular flows and migration flows.
+
+## Design Proposal
+
+A "flow" is an input topic, an output topic, and a function that transform the values.
+
+Note that the input topic is "owned" by the flow, not the ingestion topic (i.e. topic the ingestion point writes to).
+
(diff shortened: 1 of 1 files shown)


rfd/0004/README.mdx
```diff
@@ -0,0 +1,144 @@
+---
+authors: George Leung <phisgr@gmail.com>
+title: Flows Input/Output Topics
+state: discussion
+---
+
+## Objective
```
cjus, line 7: Perhaps this proposal is missing a WHY?  Why is this important - other than the potential to unify flows?

rfd/0004/README.mdx
```diff
@@ -0,0 +1,144 @@
...
+
+These extra hops between Kafka and Moose will introduce extra latency and consume more network bandwidth.
+
+### Engineering Impact
+
+With this proposed change the design is more uniform.
+We can probably simplify the implementation
+(at the cost of worse performance, as noted above).
+
+### User Impact
```
DatGuyJonathan, line 64: How does this change the `/flows` folder structure & file naming convention, if any?

Is there an example `flow.ts` for this new approach? I think I remember an import mapping PR and it looked clean. That could be a good way to show why this is easier to understand for users.

callicles:
@phiSgr Can you merge this?

callicles (approved):

PR #1152, opened 2024-05-02
moose logs command

DatGuyJonathan:
The default for `moose-cli logs` dumps the last few lines. Can add a `--filter <string>` or `--tail` or both. E.g:

Screenshot 2024-05-02 at 12 29 42 PM


callicles (approved):
Nice work!

PR #1157, opened 2024-05-03
Sql template string

georgevanderson:


apps/framework-cli/src/project.rs
```diff
@@ -235,6 +235,15 @@ impl Project {
             api_file.write_all(api_file_content.as_bytes())?;
         }
 
+        let deno_consumption_helper = deno_dir.join(CONSUMPTION_HELPERS);
+
+        if !deno_consumption_helper.exists() {
```
callicles: you may want to always override so that when we ship a new version you override the old one and update it

callicles (approved):
While not a big fan of the current solution, I understand why that's the first version. I wonder if we could do something where we write clickhouse type mappings from the cli into `.moose` where that could be picked up by the ts server to do the proper mapping.

apps/framework-cli/src/framework/consumption-helpers.ts
```diff
@@ -0,0 +1,141 @@
...
+  Date32: "string",
+  Date64: "string",
+  DateTime32: "string",
+  DateTime64: "string",
+  IPv4: "string",
+  IPv6: "string",
+};
+
+export const mapFromClickHouseType = (clickHouseType: string) => {
+  return typeMapping?.[clickHouseType] || "string";
```
DatGuyJonathan: Is the optional chaining needed? I'm not sure when typeMapping would be null.

apps/framework-cli/src/framework/consumption-helpers.ts
```diff
@@ -0,0 +1,141 @@
...
+export const mapToClickHouseType = (value: any) => {
+  if (typeof value === "number") {
+    // infer the float or int according to exist remainder or not
+    if (value % 1 !== 0) return "Float";
+    return "Int";
+  }
+  // When define column type or query result with parameterized query, The Bool or Boolean type both supported.
+  // But the column type of query result only return Bool, so we only support Bool type for safety.
+  if (typeof value === "boolean") return "Bool";
+  if (typeof value === "string") return "String";
```
DatGuyJonathan: This last `if` could probably be removed since it will default to `String`.

apps/framework-cli/src/framework/consumption-helpers.ts
```diff
@@ -0,0 +1,141 @@
...
+ * Only support to convert number to Int or Float, boolean to Bool, string to String, other types will convert to String.
+ * If exist complex type e.g: object, Array, null, undefined, Date, Record.. etc, just convert to string type by ClickHouse function in SQL.
+ * ClickHouse support converting string to other types function.
+ * Please see Each section of the https://clickhouse.com/docs/en/sql-reference/functions and https://clickhouse.com/docs/en/sql-reference/functions/type-conversion-functions
+ * @param value
+ * @returns 'FLoat', 'Int', 'Bool', 'String'
+ */
+
+export const mapToClickHouseType = (value: any) => {
+  if (typeof value === "number") {
```
DatGuyJonathan, line 47: Maybe constants or an enum could be useful for these strings and in `typeMapping` just to avoid any possible typos in the future.

DatGuyJonathan (approved):

PR #1160, opened 2024-05-03
include interface in an aggregation

DatGuyJonathan:


apps/framework-cli/src/framework/typescript/templates.rs
```diff
@@ -364,29 +364,39 @@ pub static BASE_AGGREGATION_SAMPLE_TEMPLATE: &str = r#"
...
         uniqState(userId) as dailyActiveUsers,
         toStartOfDay(timestamp) as date
     FROM ParsedActivity_0_0
     WHERE activity = 'Login' 
     GROUP BY toStartOfDay(timestamp)
     `,
-    orderBy: 'date',
-};
+  orderBy: "date",
+} as Aggregation;
```
phiSgr: Consider `satisfies T as T`
https://github.com/microsoft/TypeScript/issues/13626#issuecomment-1472348545

phiSgr (approved):

PR #1170, opened 2024-05-06
Update flow templates

DatGuyJonathan:
A couple things from https://linear.app/514/issue/514-934/flows-quick-wins:

- Timestamp as string in default data model
- Change flow arg name from event to source so it's more generic
- Also updated imports to use models.ts


callicles (approved):
LGTM

PR #1179, opened 2024-05-06
Verify datamodel name on flow init

DatGuyJonathan:
Screenshot 2024-05-06 at 1 47 42 PM

### DIFF
--- a/apps/framework-cli/src/cli/routines/flow.rs
+++ b/apps/framework-cli/src/cli/routines/flow.rs
@@ -117,8 +117,8 @@ pub fn create_flow_file(
     }
 
     Ok(RoutineSuccess::success(Message::new(
-        "".to_string(),
-        "".to_string(),
+        "Created".to_string(),
+        format!("flow {}", flow_file_path.display()),
     )))
 }
 
--- a/apps/framework-cli/src/cli/routines/flow.rs
+++ b/apps/framework-cli/src/cli/routines/flow.rs
@@ -3,7 +3,7 @@ use std::{fs, io::Write, process::Stdio};
 use tokio::io::{AsyncBufReadExt, BufReader};
 use tokio::process::Command;
 
-use crate::cli::display::Message;
+use crate::cli::display::{Message, MessageType};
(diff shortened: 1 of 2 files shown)


apps/framework-cli/src/cli/routines/flow.rs
```diff
@@ -76,12 +76,75 @@ pub fn create_flow_file(
...
+pub fn verify_datamodel(project: &Project, datamodel: String) -> anyhow::Result<()> {
+    let child = std::process::Command::new("grep")
+        .arg("-r")
+        .arg("--include=*.ts")
+        .arg("-E")
+        .arg(format!("export\\s+interface\\s+\\b{}\\b", &datamodel))
+        .arg(project.schemas_dir())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()?;
```
DatGuyJonathan, line 134: One option is to crawl the schema, get framework objects, and check against that. It seems heavy for just an init command.

Another option is to check the .moose folder, but that requires dev to have been run.

Grep seems lightweight enough and works on Mac/Linux. LMK what y'all think.

callicles: That's a creative solution!

Quick question - should we fail if the datamodel doesn't exist?

DatGuyJonathan: Their order of operation could vary. They could create their data models first, then the flow, in which case we wouldn't print next steps. They could create the flow file first, then their data models, in which case we're just reminding them to add their data models by printing these next steps. I think that's a valid workflow. Feels like the reminder is gentler & they won't have to come back & re-run the flow init command.

callicles: 🤔  that feels fair for the output but does it work for the input? ie can you write a flow without having defined the data you will take as input?

callicles: Your other PR should cover this

callicles (approved):

PR #1185, opened 2024-05-06
docs: update documentation for TS models

callicles:
* Updates documentation to remove the mention of prisma models
* Updates the navbar to be collapsed as a default

### DIFF
--- a/apps/framework-docs/src/pages/building/data-models/model-data.mdx
+++ b/apps/framework-docs/src/pages/building/data-models/model-data.mdx
@@ -23,6 +23,20 @@ MooseJS's data modeling is integrated with the data modeling capabilities of you
 For now that's Typescript Interfaces but Python is in the works, leveraging Data classes.
 We don't currently support all of Typescript schema modeling features, but we're working on it.
 
+### `Key<T>`
+
+The `Key<T>` type is a MooseJS-specific type that allows you to define a primary key for your data model.
+This is a required field for all data models. It can be added to multiple columns if you have a composite key.
+
+We are planning to add a moose helper package that will allow you to import the `Key` type from it. Right now,
+you have to defined it in your project. It doesn't matter what the type of the key is.
+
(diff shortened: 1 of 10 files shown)


apps/framework-docs/src/pages/building/data-models/model-data.mdx
```diff
@@ -2,54 +2,63 @@ import { Cards, Card, Tabs } from "nextra/components";
...
 }
 ```
 
 ## Data Model conventions
 
-As you can see, the data model is very simple. It's just a list of fields and their types. MooseJS's data modeling is built on top of Prisma's intuitive schema modeling language. You can read more about Prisma's schema modeling language [here](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema). We don't currently support all of Prisma's schema modeling features, but we're working on it; and we’ll also soon be adding support to define Data Models as Typescript interfaces.
+As you can see, the data model is very simple. It's just a list of fields and their types.
+MooseJS's data modeling is integrated with the data modeling capabilities of your language of choice.
+For now that's Typescript Interfaces but Python is in the works, leveraging Data classes.
+We don't currently support all of Typescript schema modeling features, but we're working on it.
```
03cranec, line 24: I would add an explanation about the Key functionality here

callicles: The `Key` as in `type Key<T extends string | number> = T;` or key functionality of ts data model in general?

03cranec: maybe both? i would explain why that specific line of code is needed / what it does, and then I would explain what marking a property as a Key does. and maybe any limitations on the convention (like, is it 1 and only 1 property must be marked as Key? can the Key only be a string or number type?)

callicles: Cool - I believe I addressed the feedback :) Thank you

03cranec (approved):

PR #1192, opened 2024-05-07
fix: properly error when there is a problem with initializing using the user home directory

callicles:
* Do not panic when user directory is protected, properly fail
* Refactor the initialize routine to not use `Routines`
* Check project name for special characters to prevent a user from calling their project with a dot in there
* Removes `models` folder (not used)
* Adds some errors with the new error pattern

### DIFF
--- a/apps/framework-cli/src/cli.rs
+++ b/apps/framework-cli/src/cli.rs
@@ -22,17 +22,16 @@ use logger::setup_logging;
 use settings::{read_settings, Settings};
 
 use crate::cli::routines::aggregation::create_aggregation_file;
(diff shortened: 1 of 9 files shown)


apps/framework-cli/src/cli.rs
```diff
@@ -104,6 +103,13 @@ async fn top_command_handler(
                 &settings
             );
 
+            if name.contains('.') {
```
phiSgr: If this is about docker compose project name, I suggest removing special characters there.

callicles: Good call - fixed it

phiSgr (approved):

PR #1195, opened 2024-05-08
Implement retry when creating aggregations

DatGuyJonathan:
For example, aggregation A could depend on B which depends on C which depends on the actual source table:

aggsDependencyChain

&nbsp;
The order in which they're created isn't guaranteed, so we retry later if they fail with a specific dependency error:
&nbsp;

createAggsRetry

### DIFF
--- a/apps/framework-cli/src/framework/aggregations.ts
+++ b/apps/framework-cli/src/framework/aggregations.ts
@@ -3,6 +3,9 @@ import {
   ClickHouseClient,
 } from "npm:@clickhouse/client-web@1.0.1";
 import { watch } from "npm:chokidar@3.6.0";
+import * as fastq from "npm:fastq@1.17.1";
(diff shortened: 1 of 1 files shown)


apps/framework-cli/src/framework/aggregations.ts
```diff
@@ -3,6 +3,9 @@ import {
   ClickHouseClient,
 } from "npm:@clickhouse/client-web@1.0.1";
 import { watch } from "npm:chokidar@3.6.0";
+import * as fastq from "npm:fastq@1.17.1";
```
DatGuyJonathan, line 6: https://www.npmjs.com/package/fastq

Seems lightweight, recently maintained, & has lots of downloads. LMK if you have another package or approach in mind.

callicles (approved):
LGTM

PR #1197, opened 2024-05-09
fix: allow nullable arrays in typescript

callicles:
* Allows nullable arrays from typescript and encode them to empty arrays in Clickhouse
* Moves all the queries from the clickhouse queries folder to handlebars
* Remove some unused commented code for clickhouse configuration
* Fix an issue with nullable types 
* Removes generation of ts files now that we are on Ts models
* Escapes string sequences on insert


apps/framework-cli/src/framework/typescript/export_collectors.rs
```diff
@@ -79,8 +80,9 @@ pub fn get_data_model_configs(
         Value::Object(map) => {
             let mut result = HashMap::new();
             for (key, value) in map {
-                let config: DataModelConfig = serde_json::from_value(value)?;
-                result.insert(key, config);
+                let _ = serde_json::from_value(value).map(|model_config| {
```
phiSgr: Imperative `if let` is better than `.map` here, as we are doing it for the effect.

apps/framework-cli/src/infrastructure/kafka_clickhouse_sync.rs
```diff
@@ -287,7 +287,20 @@ fn mapper_json_to_clickhouse_record(
                         };
                     }
                     None => {
-                        record.insert(key, ClickHouseValue::new_string("NULL".to_string()));
+                        // Clickhouse doesn't like NULLABLE arrays so we are inserting an empty array instead.
```
phiSgr, line 301: Does it not accept `null` in a non-nullable array field, or does it really not like `Nullable(Array(T))`?

phiSgr (commented):
Greate simplification after removing tinytemplate.

apps/framework-cli/src/framework/typescript/export_collectors.rs
callicles: Thx! will switch

apps/framework-cli/src/infrastructure/kafka_clickhouse_sync.rs
callicles: You can't create a table with the schema Array NULLABLE

phiSgr: Ah indeed. `Nested type Array(String) cannot be inside Nullable type. (ILLEGAL_TYPE_OF_ARGUMENT)`

cjus (commented):
@callicles where's the code that allows nullable arrays in TypeScript?

callicles:
It is a combination of:
* In the kafka sync, if the property that is an array is not found then insert `[]`
* In the schema builder in clickhouse, if a column is an array, don't set it to nullable even if it is

tg339 (approved):
we seem to be moving back to functions for everything as opposed to sstruct methods. Not sure if that's intentional but we should probably align on a design pattern. Not urgent or blocking

phiSgr:
> Not sure if that's intentional but we should probably align on a design pattern. Not urgent or blocking

I prefer the plain function approach (as a person who is easily thwarted by friction). But currently we're not displaying a lot of the progress. This makes the CLI look less responsive than it is.

PR #1198, opened 2024-05-09
dump table to migration input topic

phiSgr:


apps/framework-cli/src/infrastructure/olap/clickhouse_alt_client.rs
```diff
@@ -0,0 +1,194 @@
...
+use crate::infrastructure::olap::clickhouse::config::ClickHouseConfig;
+use crate::infrastructure::olap::clickhouse::model::{ClickHouseColumnType, ClickHouseTable};
+
+pub fn get_pool(click_house_config: &ClickHouseConfig) -> clickhouse_rs::Pool {
+    let address = format!(
+        "tcp://{}:{}",
+        click_house_config.host, click_house_config.native_port
+    );
+
+    clickhouse_rs::Pool::new(
```
callicles, line 27: Does this create a new pool every time? or does it get a pool based on the `address`?

callicles (approved):
Small question about the connection pooling - otherwise looks good to me - nice work

apps/framework-cli/src/infrastructure/olap/clickhouse_alt_client.rs
phiSgr: https://github.com/514-labs/moose/pull/1198/files#diff-d4af486162b78407e0eea7401a6a4376ac6e0578a5f8a70144a06be8549c0519R240

PR #1200, opened 2024-05-09
Add recommended vscode extensions

DatGuyJonathan:


apps/framework-cli/src/cli/routines/initialize.rs
```diff
@@ -31,6 +31,16 @@ pub fn initialize_project(project: &Project) -> Result<RoutineSuccess, RoutineFa
         )
     })?;
 
+    project.create_vscode_files().map_err(|err| {
+        RoutineFailure::new(
+            Message::new(
+                "Failed".to_string(),
+                "to create 'vscode' files, Check permissions or contact us".to_string(),
+            ),
```
cjus, line 39: Should we use the 514 email address for moosejs?

georgevanderson (approved):

apps/framework-cli/src/cli/routines/initialize.rs
DatGuyJonathan: That makes sense! We probably should pull all of these "contact us" strings to a const so the email is in one place. Will check it out in a follow-up.

PR #1206, opened 2024-05-09
feat: standardize to UTC in the clickhouse db

callicles:
Screenshot 2024-05-09 at 2 54 36 PM


phiSgr:
Is this a stop gap before we decide how best to handle time zones, or do we really want to use discard the timezone information?

callicles:
Hum I don't think this is discarding the timezone - this is converting everything to UTC, ie reading the timezone as input when parsing the date and then converting to UTC. 

It is standardizing storage.

phiSgr:
I think clickhouse can distinguish "3PM UTC-7" and "6PM UTC-4".

phiSgr (approved):

PR #1218, opened 2024-05-10
store the data model in clickhouse

phiSgr:


apps/framework-cli/src/cli/routines.rs
```diff
@@ -260,6 +261,16 @@ pub async fn start_development_mode(project: Arc<Project>) -> anyhow::Result<()>
     let (framework_object_versions, version_syncs) =
         initialize_project_state(project.clone(), &mut route_table).await?;
 
+    {
+        let mut client = get_pool(&project.clickhouse_config).get_handle().await?;
```
callicles, line 265: Is the client thread safe?

feels a weird location to have that in a closure

callicles (approved):
Looks good to me - small question about the client and instanciation but we could solve it in a subsequent PR

apps/framework-cli/src/cli/routines.rs
phiSgr: I think the pool is thread safe but the client is not.

PR #1221, opened 2024-05-10
Olivia/product update 1 v2

okane16:
Blog revisions


callicles (approved):

callicles:
Not sure if our rendering is compatible with it but would be amazing to have synthax highlighting on the code

okane16:
Ideally would like to have a component for the code blocks so we could get the syntax highlight, going to merge and investigate for this and future posts

PR #1225, opened 2024-05-13
Consumption in prod

georgevanderson:


apps/framework-cli/src/cli/routines/consumption.rs
```diff
@@ -0,0 +1,96 @@
...
+            )
+        })?;
+
+    Ok(RoutineSuccess::success(Message::new(
+        "Created".to_string(),
+        format!("consumption api {}", apis_file_path.display()),
+    )))
+}
+
+pub fn start_consumption_process(project: &Project) -> anyhow::Result<()> {
```
DatGuyJonathan, line 51: What's calling this function?

apps/framework-cli/src/framework/typescript/templates.rs
```diff
@@ -422,6 +422,34 @@ export default {
...
+  
+interface Aggregation {
+  select: string;
+  orderBy: string;
+}
+
+export default {
+  select: ``,
+  orderBy: "",
+} satisfies Aggregation as Aggregation;
```
DatGuyJonathan: Should these aggregation stuff be removed from the api file?

DatGuyJonathan (approved):

PR #1230, opened 2024-05-14
feat: allow data model to not have a landing table

callicles:
* Adds `latest` to the CLI output on start
* clean up table in the Route metadata
* Add configuration to not have necessarily a table as an output to the data model


cjus (approved):
Large PR but LGTM

PR #1248, opened 2024-05-15
check no version added in the middle

phiSgr:


callicles:
Do we want to close this?

phiSgr:
Yeah it makes sense to recreate this PR after we remove core v1.

PR #1250, opened 2024-05-16
Disallow matching data model & aggregation name

DatGuyJonathan:


apps/framework-cli/src/framework/controller.rs
```diff
@@ -134,17 +134,18 @@ pub fn get_all_framework_objects(
...
             let entry = entry?;
             let path = entry.path();
             if path.is_dir() {
                 debug!("<DCM> Processing directory: {:?}", path);
-                get_all_framework_objects(framework_objects, &path, version)?;
+                get_all_framework_objects(framework_objects, &path, version, aggregations)?;
             } else if is_schema_file(&path) {
                 debug!("<DCM> Processing file: {:?}", path);
-                let objects = get_framework_objects_from_schema_file(&path, version)?;
+                let objects = get_framework_objects_from_schema_file(&path, version, aggregations)?;
```
phiSgr, line 148: It should be fine if a data model in an old version have the same name as an aggregation.

DatGuyJonathan: How do they get to that point? Like, what would their repro steps be over time?

phiSgr: create new version
delete data model
create aggregation with the same name

this might be quite implausible, but if it's worth preventing the name clashing of the two types, i think it's equally worth preventing false positives.

DatGuyJonathan: Agreed, appreciate that mindset. I updated the PR so it disallows the name clashing for current version only. Let me know if there's anything else.

phiSgr (approved):

PR #1257, opened 2024-05-18
fix: fix project references for going to production

callicles:
* Removes the global mutable project reference
* Do some refactors removing the Routines for the bump version and generate migration commands
* Removes the db_name from the Table object, it is a global project configuration

Currently in production, we can see some logs that should not be there since we are in prod mode. 

They are showing up because the production environment was only set on the global env and not on the local object

```rust
    pub fn set_enviroment(&self, production: bool) {
        let mut proj = PROJECT.lock().unwrap();
        proj.is_production = production;
    }
```

I removed the global. Through this, I am becoming more convinced that explicit dependency on the project object is necessary and will yield a return on the investment of punching through when necessary. It makes the dependency explicit, makes functions more testable, and reveals smells of code when abstractions are leaking in the wrong parts of the code. 

ie by making it explicit here, I started punching through `project` in places that did not make sense. That made me ask, why am I doing this? is it needed? The `db_name` property is a good example of that.

### DIFF
--- a/apps/framework-cli/src/cli.rs
+++ b/apps/framework-cli/src/cli.rs
@@ -255,7 +255,7 @@ async fn top_command_handler(
             info!("Running dev command");
 
             let mut project = load_project()?;
-            project.set_enviroment(false);
+            project.set_is_production_env(false);
             let project_arc = Arc::new(project);
 
             crate::utilities::capture::capture!(
@@ -305,7 +305,7 @@ async fn top_command_handler(
             info!("Running prod command");
             let mut project = load_project()?;
 
-            project.set_enviroment(true);
+            project.set_is_production_env(true);
             let project_arc = Arc::new(project);
 
             crate::utilities::capture::capture!(
--- a/apps/framework-cli/src/cli.rs
+++ b/apps/framework-cli/src/cli.rs
@@ -29,8 +29,9 @@ use crate::cli::routines::dev::{copy_old_schema, create_deno_files};
 use crate::cli::routines::flow::{create_flow_directory, create_flow_file};
 use crate::cli::routines::initialize::initialize_project;
 use crate::cli::routines::logs::{follow_logs, show_logs};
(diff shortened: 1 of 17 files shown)


apps/framework-cli/src/project.rs
```diff
@@ -170,9 +154,8 @@ impl Project {
         }
     }
 
-    pub fn set_enviroment(&self, production: bool) {
-        let mut proj = PROJECT.lock().unwrap();
-        proj.is_production = production;
+    pub fn set_enviroment(&mut self, production: bool) {
```
cjus: Not sure about this naming.  set_environment doesn't seem to be specific enough - set_production_env or set_prod_env would result in the boolean argument value making sense while reading code.

apps/framework-cli/src/cli.rs
```diff
@@ -29,8 +29,9 @@ use crate::cli::routines::dev::{copy_old_schema, create_deno_files};
 use crate::cli::routines::flow::{create_flow_directory, create_flow_file};
 use crate::cli::routines::initialize::initialize_project;
 use crate::cli::routines::logs::{follow_logs, show_logs};
+use crate::cli::routines::migrate::generate_migration;
 use crate::cli::routines::templates;
-use crate::cli::routines::version::BumpVersion;
+use crate::cli::routines::version::bump_version;
```
cjus, line 34: How did the JS casing get past the Clippy?

cjus (commented):
@callicles really nice work!  These changes are going to go a long way towards code maintainability.  Just let two comments.

apps/framework-cli/src/project.rs
callicles: Good call! will change

apps/framework-cli/src/cli.rs
callicles: It was a `struct` before now it is a function

cjus (approved):
LGTM

PR #1280, opened 2024-05-22
Fix sample agg/consumption queries so it will aggregate new data

DatGuyJonathan:


apps/framework-cli/src/framework/typescript/templates.rs
```diff
@@ -370,7 +370,7 @@ interface Aggregation {
 export default {
   select: ` 
     SELECT 
-        count(distinct userId) as dailyActiveUsers,
+        uniqState(userId) as dailyActiveUsers,
```
DatGuyJonathan, line 373: `count distinct` operates on the entire data set at once. We probably saw it "work" because we already had data, then created the queries, which created the materialized view that looked properly aggregated.

But it doesn't aggregate new data because `count distinct` just doesn't support the incremental updates that a materialized view needs. Clickhouse needs an efficient way to track what's been aggregated and they do that through aggregation functions like `uniqState` & `uniqMerge`.

callicles (approved):

PR #1303, opened 2024-05-24
Nicolas/514 969 move flows to node instead of deno

callicles:
* Moves flows from Deno to node
* Removes the Flow container
* Enables flows to use packages
* Remove from the TS all the logic about loading flows so that it can be reused for Python
* Adds a Flows process manager
* Isolates all the flows in different processes so that if one crashes the others still work

QA Still wip


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -427,7 +427,7 @@ impl Webserver {
             MessageType::Success,
             Message {
                 action: "Started".to_string(),
-                details: "Development server.\n\n".to_string(),
+                details: "Webserver.\n\n".to_string(),
```
tg339, line 430: Why are we changing this? Development server seems clearer for locak

tg339 (approved):

apps/framework-cli/src/cli/local_webserver.rs
callicles: Because this is also the message that's run in production. ie it is the same server

PR #1326, opened 2024-05-29
cors-consumption

georgevanderson:


callicles:
What are you trying to fix here? it looks like you are adding cors headers to a response on the client

callicles (approved):

PR #1329, opened 2024-05-30
Move aggs to node

DatGuyJonathan:


apps/framework-cli/src/framework/aggregations/registry.rs
```diff
@@ -0,0 +1,43 @@
...
+use crate::infrastructure::olap::clickhouse::config::ClickHouseConfig;
+
+use super::model::{Aggregation, AggregationError};
+
+pub struct AggregationProcessRegistry {
+    registry: HashMap<String, Child>,
+    clickhouse_config: ClickHouseConfig,
+}
+
+impl AggregationProcessRegistry {
```
callicles, line 15: Why are you thinking that we need a registry here? 
Since the Aggregation process is small lived, we should not have to kill it? or are you thinking we put it for when we have a lot of aggregations?
Or did yo think about something else

callicles (commented):
LGTM - some question about the Registry

callicles:
Also will you update our current internal projects and templates with the new dependencies

apps/framework-cli/src/framework/aggregations/registry.rs
DatGuyJonathan: I was thinking the user could make multiple changes to the file & save quickly. Those changes would trigger multiple file watcher events, creating multiple aggregation processes, so I was concerned about getting into weird race conditions. I figured it was safer to have a registry that can kill the child process before starting new ones. The registry is also useful if we end up having a process for each aggregation, but for now, it's primarily to avoid race conditions.

callicles: That works

callicles (approved):

PR #1345, opened 2024-05-31
chore: simplify dependencies for the user

callicles:
* Use the new package for moose internal dependencies
* Update all internal projects to use the new moose lib
* Add a type helper for data model config

### DIFF
--- a/packages/ts-moose-lib/src/index.ts
+++ b/packages/ts-moose-lib/src/index.ts
@@ -23,12 +23,12 @@ export enum IngestionFormat {
   JSON_ARRAY = "JSON_ARRAY",
 }
 
-export type DataModelConfig = Partial<{
+export type DataModelConfig<T> = Partial<{
   ingestion: {
     format?: IngestionFormat;
   };
   storage: {
     enabled?: boolean;
-    order_by_fields?: string[];
+    order_by_fields?: (keyof T)[];
   };
 }>;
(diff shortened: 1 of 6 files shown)


packages/ts-moose-lib/src/index.ts
```diff
@@ -4,3 +4,18 @@ export interface Aggregation {
 }
 
 export type Key<T extends string | number | Date> = T;
+
+export enum IngestionFormat {
+  JSON = "JSON",
+  JSON_ARRAY = "JSON_ARRAY",
+}
+
+export type DataModelConfig = Partial<{
```
phiSgr: What about making it generic, and the order_by has to be field names of the type parameter?

callicles: Can you spell it out? I don't see what you are suggesting

phiSgr: `order_by_fields?: (keyof T)[];`

And if I mistype the field name:
```
app/datamodels/models.ts:47:23 - error TS2820: Type '"observedTimestam"' is not assignable to type 'keyof Log'. Did you mean '"observedTimestamp"'?

47     order_by_fields: ["observedTimestam"],
```

But then this is a bit complicated if we want to support nested keys.

callicles: We can start with that and then iterate

phiSgr (approved):

PR #1354, opened 2024-06-04
0.3 pointed to a version with data models different from deployed

phiSgr:


apps/framework-internal-app/project.toml
```diff
@@ -26,4 +26,4 @@ host_port = 3001
 "0.0" = "1fe7b86"
 "0.1" = "005108e"
 "0.2" = "969a0b7"
-"0.3" = "f8ebebb"
+"0.3" = "22d9fcb"
```
DatGuyJonathan, line 29: Do we need to add 'description' back to UserActivity model as well?

DatGuyJonathan (approved):

apps/framework-internal-app/project.toml
phiSgr: ah yes

PR #1361, opened 2024-06-05
Move Consumption to node

georgevanderson:
Additionally fixes for using pnpm and properly killing child processes of ts-node


packages/ts-moose-lib/src/commons.ts
```diff
@@ -0,0 +1,60 @@
...
+const [
+  ,
+  ,
+  CLICKHOUSE_DB,
+  CLICKHOUSE_HOST,
+  CLICKHOUSE_PORT,
+  CLICKHOUSE_USERNAME,
+  CLICKHOUSE_PASSWORD,
+  CLICKHOUSE_USE_SSL,
+] = process.argv;
```
callicles: That feels a bit risky to have in commons - that probably should say in the called script no?

packages/ts-moose-lib/src/commons.ts
```diff
@@ -0,0 +1,60 @@
...
+      ? "https"
+      : "http";
+  console.log(
+    `Connecting to Clickhouse at ${protocol}://${CLICKHOUSE_HOST}:${CLICKHOUSE_PORT}`,
+  );
+  return createClient({
+    url: `${protocol}://${CLICKHOUSE_HOST}:${CLICKHOUSE_PORT}`,
+    username: CLICKHOUSE_USERNAME,
+    password: CLICKHOUSE_PASSWORD,
+    database: CLICKHOUSE_DB,
```
callicles: and make those parameters?

callicles (approved):
Nice work!

apps/framework-internal-app/app/flows/UserActivity/ParsedActivity/flow.ts
```diff
@@ -9,7 +9,7 @@ export default function run(event: UserActivity): ParsedActivity {
   // Convert local timestamp to UTC and return new ParsedActivity object.
   return {
     eventId: event.eventId, // Retain original event ID.
-    userId: "puid" + event.userId, // Example: Prefix user ID.
+    userId: "puid2" + event.userId, // Example: Prefix user ID.
```
DatGuyJonathan: Why puid2? This affects my load test so ideally we don't change this if we don't have to.

PR #1366, opened 2024-06-06
Batch flows

DatGuyJonathan:
https://github.com/514-labs/moose/assets/7286437/be6e6afb-8906-4597-8d6b-f8ac75da1004

### DIFF
--- a/apps/framework-cli/src/framework/typescript/ts_scripts/flow.ts
+++ b/apps/framework-cli/src/framework/typescript/ts_scripts/flow.ts
@@ -112,13 +112,12 @@ const startProducer = async (): Promise<void> => {
 };
 
 const handleMessage = async (
-  targetTopic: string,
   flowFn: FlowFunction,
   message: KafkaMessage,
-): Promise<void> => {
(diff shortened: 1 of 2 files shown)


apps/framework-cli/src/framework/typescript/ts_scripts/flow.ts
```diff
@@ -159,10 +156,26 @@ const startConsumer = async (
...
+      const messages = await Promise.all(
+        batch.messages.map((message) => handleMessage(flowFunction, message)),
+      );
+
+      const filteredMessages = messages.filter((msg) => msg !== null);
+
+      if (filteredMessages.length > 0) {
+        await producer.send({
+          topic: targetTopic,
+          messages: filteredMessages as { value: string }[],
```
DatGuyJonathan, line 170: This is what I was doing wrong last time I tried batch. I was receiving a batch, but still sending one at a time. There's a noticeable difference now that both sides are batched.

phiSgr (approved):

PR #1367, opened 2024-06-06
Origin/tim/514 python aggs

tg339:


apps/framework-cli/src/framework/python/templates.rs
```diff
@@ -67,6 +67,29 @@ my_flow = Flow(
...
+pub static PTYHON_BASE_AGG_SAMPLE_TEMPLATE: &str = r#"
+from dataclasses import dataclass
+
+# Here is a sample aggregation query that calculates the number of daily active users
+# based on the number of unique users who complete a sign-in activity each day.
+
+@dataclass
+class Aggregation:
+    select: str
+    order_by: str
```
phiSgr, line 79: ideally this should be in a the library.

phiSgr (approved):
is there hot reload?

apps/framework-cli/src/framework/python/templates.rs
tg339: Yeah, that'll move in the final PR and with some of the work that @cjus is figuring out in terms of packaging

PR #1370, opened 2024-06-07
Nicolas/514 1026 add generate sdk command to the cli

callicles:
* Starts refactoring of the `controller` to isolated code state loading code
* adds `generate sdk` command


apps/framework-cli/src/cli/commands.rs
```diff
@@ -86,6 +88,17 @@ pub struct GenerateArgs {
 #[derive(Debug, Subcommand)]
 pub enum GenerateCommand {
     Migrations {},
+    Sdk {
+        /// Language of the SDK to be generated
+        #[arg(default_value_t = SupportedLanguages::Typescript, value_enum, short, long)]
+        language: SupportedLanguages,
+        /// Where the SDK files should be written to
+        #[arg(default_value = "./sdk", short, long)]
```
phiSgr, line 96: do we want to suggest the user to check the SDK into source control, if not should we add this to the default gitignore.

callicles: yeah I think that's the idea - in connectors for example you want to refer to SDK code directly

apps/framework-cli/src/framework/core.rs
```diff
@@ -0,0 +1,23 @@
...
+/// the language speicifc modules and build the representation of what the state of the infrastructure should be.
+/// It will then build a representation of the current state of the infrastructure and compare the two to
+/// determine what needs to be done.
+/// It will finally build a plan of what needs to be done and delegate execution
+/// to modules specific to the infrastructure provider.
+///
+///
+/// ┌──────────────┐                     ┌──────────────┐
+/// │              │                     │              │
+/// │  Python      ├──┐                ┌─►  Clikhouse   │
```
phiSgr: typo

phiSgr (approved):

apps/framework-cli/src/framework/core.rs
callicles: Thank you

PR #1384, opened 2024-06-11
Chunk flow batches

DatGuyJonathan:


apps/framework-cli/src/cli/settings.rs
```diff
@@ -57,6 +58,7 @@ impl Default for Telemetry {
 pub struct Settings {
     #[serde(default)]
     pub logger: LoggerSettings,
+    #[allow(dead_code)]
     #[serde(default)]
     pub features: Features,
```
DatGuyJonathan: Do we need this for anything?

phiSgr (approved):

apps/framework-cli/src/cli/settings.rs
callicles: Not anymore - we could reuse but it is not currently used

apps/framework-cli/src/cli/settings.rs
```diff
@@ -23,6 +23,7 @@ const ENVIRONMENT_VARIABLE_PREFIX: &str = "MOOSE";
 
 #[derive(Deserialize, Debug)]
 pub struct Features {
+    #[allow(dead_code)]
     pub coming_soon_wall: bool,
```
callicles: This we should remove

callicles (approved):
Love this - will you do the same for the python version

DatGuyJonathan:
> Love this - will you do the same for the python version

Sure, added a ticket for it.

PR #1385, opened 2024-06-12
skip processing of framework object if the data model did not change

phiSgr:


phiSgr:
Now redundant since we have new moose planning.

PR #1391, opened 2024-06-14
Handle flow fn returning an array

DatGuyJonathan:
Also updated docs to reflect this new ability.

### DIFF
--- a/apps/framework-docs/src/pages/building/flows/setup.mdx
+++ b/apps/framework-docs/src/pages/building/flows/setup.mdx
@@ -2,14 +2,19 @@ import { Callout, FileTree, Link } from "nextra/components";
 
 # Create Flow in CLI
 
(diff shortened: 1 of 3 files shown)


apps/framework-docs/src/pages/building/flows/setup.mdx
```diff
@@ -105,29 +118,37 @@ import { UserActivity } from "/path/to/UserActivity";
...
+    timestamp: new Date(source.timestamp), // Convert timestamp to UTC.
   };
 }
 ```
 
-Adjust function specifics according to your data transformation needs. This structured approach ensures clarity in your
-data processing tasks, enabling Moose to efficiently handle and automate data flows.
+Adjust function specifics according to your data transformation needs. The
+function's return type can be null (meaning no transformation will take place),
+a single object, or an array of objects. This structured approach ensures
```
DatGuyJonathan, line 136: Main doc change is this paragraph explaining the different return types, and the example on line 123 above. Rest are formatting changes.

callicles (approved):
LGTM

PR #1420, opened 2024-06-16
Nicolas/514 1041 refactor sdk generation in typescript for non hardcoding of

callicles:
* Finally removes TinyTemplate from the dependencies
* Add loader when generating the SDK
* Add an option in the SDK to only generate TS files and not the `package.json` or the `ts_config.json`
* Updates the docs for the generate SDK command
* Enables more options to be accepted in the logger settings for log levels

Tested in the internal deployment, in the default app and on a customer code base


apps/framework-cli/src/framework/typescript/generator.rs
```diff
@@ -305,32 +277,24 @@ pub fn std_table_to_typescript_interface(
...
+        ts_objects.push(TypescriptObjects::new(interface));
     }
 
     Ok(ts_objects)
 }
 
-fn collect_enums(framework_objects: &SchemaVersion) -> Vec<TSEnum> {
-    let mut enums: Vec<TSEnum> = Vec::new();
+fn collect_enums(framework_objects: &SchemaVersion) -> HashSet<TSEnum> {
+    let mut enums: HashSet<TSEnum> = HashSet::new();
```
tg339, line 294: Does removing order here end up impacting languages like python where order of declarations can make an impact on number of tries that it takes to accomplish a task? i.e. if you have nested structures and the nested structure isn't declared ahead of the other one, you can cause errors.

tg339 (approved):

apps/framework-cli/src/framework/typescript/generator.rs
callicles: The above order is only impacting order in the import statements. Not structural so it should be fine.

apps/framework-cli/src/cli/commands.rs
```diff
@@ -98,6 +98,9 @@ pub enum GenerateCommand {
         /// The location of the Moose project
         #[arg(default_value = ".", short, long)]
         project_location: PathBuf,
+        /// Whether or not to generate a full fledged package or just the source files in the language of choice
+        #[arg(default_value = "false", short = 'a', long)]
+        packaged: bool,
```
DatGuyJonathan: Generally I prefer short & long to start with the same letter so they're easier dx. Looks like "p" is already taken by project location. Couple more options could be `--full-package / -f` and `--complete / -c`.

callicles: Ha thank you for the suggestion - I also prefer it that way but did not think of other alternatives. Will switch it up!

PR #1439, opened 2024-06-18
Added a new quick references page and a FAQs page

camelCasedAditya:
The quick reference page is for users who already know the basics of how MooseJS works but need to refer to the docs quickly.

### DIFF
--- a/apps/framework-docs/src/pages/faqs.mdx
+++ b/apps/framework-docs/src/pages/faqs.mdx
@@ -1,0 +1,11 @@
+# FAQs
+
+---
+
+### Why is Docker is not working after making changes to some fields?
+
+Terminate the Docker containers and restart Docker manually. If you have problems restarting run `npm run dev` twice.
+
+### Why has a test run crashed?
+
+Making changes to the data model while a test run is going will cause it to crash because of the real time updates.
--- a/apps/framework-docs/src/pages/faqs.mdx
+++ b/apps/framework-docs/src/pages/faqs.mdx
@@ -6,6 +6,6 @@
 
 Terminate the Docker containers and restart Docker manually. If you have problems restarting run `npm run dev` twice.
 
-### Why has a test run crashed?
+### Can you change the data model while ingesting data?
 
-Making changes to the data model while a test run is going will cause it to crash because of the real time updates.
+Making changes to the data model while ingesting data will cause the your application to crash because MooseJS updates everything in realtime.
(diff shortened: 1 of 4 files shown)


apps/framework-docs/src/pages/quick-reference.mdx
```diff
@@ -0,0 +1,341 @@
...
+        and add a new Data Model definition.
+
+        ```ts filename="datamodels/models.ts" copy
+        import { Key } from "@514labs/moose-lib";
+
+        export interface UserActivity {
+        eventId: Key<string>;
+        timestamp: string;
+        userId: string;
+        activity: string;
```
callicles: Looks like you could add some indentation to those code blocks to make them render better

callicles (commented):
Small change :)

apps/framework-docs/src/pages/quick-reference.mdx
```diff
@@ -0,0 +1,341 @@
...
+        ```
+
+        ## Supported Types
+
+        Here is the status of MooseJS support for various data types in Data Models:
+        | Clickhouse | Typescript | Moose |
+        |:---------- |:---------- |:------- |
+        | String | String | ✅ |
+        | Boolean | Boolean | ✅ |
+        | Int64 | Number | ✅ |
```
phiSgr, line 35: I think it's misleading to say we support other number types as we map all ts number types to Float64

apps/framework-docs/src/pages/faqs.mdx
```diff
@@ -0,0 +1,11 @@
+# FAQs
+
+---
+
+### Why is Docker is not working after making changes to some fields?
+
+Terminate the Docker containers and restart Docker manually. If you have problems restarting run `npm run dev` twice.
+
+### Why has a test run crashed?
```
DatGuyJonathan: This gives the impression that there's a moose test framework they can run, which we don't yet have. Is this referring to sending data to ingestion point?

callicles (approved):

apps/framework-docs/src/pages/quick-reference.mdx
camelCasedAditya: I have just added a disclaimer to the quick reference page and the data models setup page in the latest Commit.

apps/framework-docs/src/pages/faqs.mdx
camelCasedAditya: I have rephrased this question for more clarity in the most recent commit. Let me know if it sounds better.

PR #1442, opened 2024-06-19
chore: add base infrastructure for the infrastructure map

callicles:
* Adds `InfrastructureMap` to have all the things to be provisioned
* Fix default app types for consumption api

### DIFF
--- a/apps/framework-cli/src/framework/core/infrastructure_map.rs
+++ b/apps/framework-cli/src/framework/core/infrastructure_map.rs
@@ -25,7 +25,7 @@ impl Topic {
         Topic {
             name: data_model.name.clone(),
             version: data_model.version.clone(),
-            // TODO configure this from data model
+            // TODO configure this from DataModelConfig
             retention_period: Duration::from_secs(60 * 60 * 24 * 7),
             source_primitive: PrimitiveSignature {
                 name: data_model.name.clone(),
--- a/apps/framework-cli/src/framework/core/infrastructure_map.rs
+++ b/apps/framework-cli/src/framework/core/infrastructure_map.rs
@@ -1,0 +1,213 @@
+use hyper::http::Method;
+use std::{collections::HashMap, time::Duration};
+
+use crate::framework::data_model::{
+    config::EndpointIngestionFormat,
+    schema::{DataModel, Table},
+};
+
+use super::primitive_map::PrimitiveMap;
+
+type TopicId = String;
+type APIId = String;
+
+#[derive(Debug)]
+pub struct Topic {
+    pub version: String,
+    pub name: TopicId,
+
+    pub retention_period: Duration,
+    pub source_primitive: PrimitiveSignature,
+}
+
+impl Topic {
+    pub fn from_data_model(data_model: &DataModel) -> Self {
+        Topic {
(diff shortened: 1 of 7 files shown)


apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub retention_period: Duration,
+    pub source_primitive: PrimitiveSignature,
+}
+
+impl Topic {
+    pub fn from_data_model(data_model: &DataModel) -> Self {
+        Topic {
+            name: data_model.name.clone(),
+            version: data_model.version.clone(),
+            // TODO configure this from data model
```
phiSgr: if we write `configure this from DataModelConfig` it might be more searchable

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -0,0 +1,213 @@
...
+
+use crate::framework::data_model::{
+    config::EndpointIngestionFormat,
+    schema::{DataModel, Table},
+};
+
+use super::primitive_map::PrimitiveMap;
+
+type TopicId = String;
+type APIId = String;
```
phiSgr, line 12: https://rust-lang.github.io/api-guidelines/naming.html

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID

BTW I've never found type alias for simple types aid readability.

apps/framework-cli/src/framework/core/primitive_map.rs
```diff
@@ -12,8 +15,8 @@ use crate::{
 #[derive(Debug)]
 pub struct PrimitiveMap {
     // This probably should not be a top level item and should be nested under the datamodels
-    enums: Vec<DataEnum>,
-    datamodels: HashMap<PathBuf, DataModel>,
+    enums: HashSet<DataEnum>,
+    datamodels: HashMap<PathBuf, Vec<DataModel>>,
```
phiSgr, line 19: I think it should clarify in a comment whether this is relative path or absolute path.
If we're storing this object in the DB, then relative path makes more sense as it'll be safe from moving the repo.

But then I don't see the benefit of using the file as key instead of the name of the data model.

callicles: Ha good comments, I was thinking about the diffs and how to be able to find data models quickly that needed update or deletion. The watcher is giving events by the file.

apps/framework-cli/src/framework/core/infrastructure_map.rs
callicles: Yeah I might just get rid of that next PR - with the ids that makes more sense

apps/framework-cli/src/framework/core/infrastructure_map.rs
callicles: Good call

apps/framework-cli/src/framework/core/primitive_map.rs
callicles: Storing that as relative paths from the root of the project makes sense

phiSgr (approved):

PR #1453, opened 2024-06-19
Nicolas/update pnpm setup action

callicles:
* updates `pnpm/action-setup`
* updates `actions/setup-node`

### DIFF
--- a/.github/workflows/test-framework-cli.yaml
+++ b/.github/workflows/test-framework-cli.yaml
@@ -57,7 +57,7 @@ jobs:
           version: 8
 
       - name: Install node
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: "20"
           cache: "pnpm"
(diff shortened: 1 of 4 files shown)


.github/workflows/test-framework-cli.yaml
```diff
@@ -91,12 +90,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - uses: pnpm/action-setup@v2
-        with:
-          version: 8
+      - uses: pnpm/action-setup@v4
```
phiSgr, line 93: what is the new version?
does that matter for the lockfile?

phiSgr (approved):

.github/workflows/test-framework-cli.yaml
callicles: It looks like the latest version of the runner is looking that the version in the lock file to install that one. It was complaining of having 2 sources of truth.

PR #1459, opened 2024-06-20
Fixed bug in data model parsing so now all objects are detected

camelCasedAditya:
Fixed the typescript so that all objects are detected when parsing the data models

### DIFF
--- a/packages/ts-moose-lib/src/typeConvert.ts
+++ b/packages/ts-moose-lib/src/typeConvert.ts
@@ -3,6 +3,7 @@ import ts, {
   isTypeReferenceNode,
   SymbolFlags,
   TypeChecker,
+  TypeFlags,
 } from "typescript";
 import { enumConvert, isEnum } from "./enumConvert";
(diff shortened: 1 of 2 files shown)


packages/ts-moose-lib/src/typeConvert.ts
```diff
@@ -53,7 +54,8 @@ const tsTypeToDataType = (
                     typeName,
                   )[1],
                 }
-              : nonNull.isClassOrInterface()
+              : nonNull.isClassOrInterface() ||
+                  (nonNull.flags & TypeFlags.Object) !== 0
```
callicles, line 58: How does `TypeFlags.Object` work?

phiSgr (commented):
maybe add a test?

packages/ts-moose-lib/src/typeConvert.ts
phiSgr: they're bitwise flags
the API provides those checking as a method for other type of types, but not the object type

phiSgr (approved):

PR #1460, opened 2024-06-20
chore: add algorithm to make the diff between 2 different infrastructure maps

callicles:
* Adds the algorithm to do the diff between 2 Infrastructure maps
* Remove types aliases for `TopicId` and `APIId`

### DIFF
--- a/apps/framework-cli/src/framework/core/infrastructure_map.rs
+++ b/apps/framework-cli/src/framework/core/infrastructure_map.rs
@@ -138,45 +138,18 @@ impl TopicToTableSyncProcess {
 }
 
 #[derive(Debug, Clone)]
-pub enum TopicChange {
-    Added(Topic),
-    Removed(Topic),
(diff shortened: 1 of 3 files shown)


apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -140,11 +137,53 @@ impl TopicToTableSyncProcess {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
+pub enum TopicChange {
+    Added(Topic),
+    Removed(Topic),
+    Updated { before: Topic, after: Topic },
+}
```
phiSgr: Looks like these enum can be generic.

callicles: Yes they can - thanks for the call out that was an easy change

phiSgr (approved):

PR #1463, opened 2024-06-20
ls command

DatGuyJonathan:
- Show data model, ingestion point, table, & view. Will look into topics next.
- Fix dcm bug where every version is a table

### DIFF
--- a/apps/framework-cli/src/framework/controller.rs
+++ b/apps/framework-cli/src/framework/controller.rs
@@ -304,7 +304,7 @@ pub async fn set_up_topic_and_tables_and_route(
         // In the case where the previous version of the data model and the new version of the data model are the same
         // we just need to use pointers to the old table and topic
         (Some(previous_fo), Some(current_table), Some(previous_table))
-            if previous_fo.data_model == fo.data_model =>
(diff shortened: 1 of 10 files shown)


apps/framework-cli/src/framework/controller.rs
```diff
@@ -304,7 +304,7 @@ pub async fn set_up_topic_and_tables_and_route(
         // In the case where the previous version of the data model and the new version of the data model are the same
         // we just need to use pointers to the old table and topic
         (Some(previous_fo), Some(current_table), Some(previous_table))
-            if previous_fo.data_model == fo.data_model =>
+            if previous_fo.data_model.columns == fo.data_model.columns =>
```
DatGuyJonathan, line 307: DataModel has file_path & version now: https://github.com/514-labs/moose/blob/main/apps/framework-cli/src/framework/data_model/schema.rs#L17 

So this equality check is always false, hence every version is a table. I think we're just trying to compare columns when trying to figure out whether it should be a view or not. Let me know if there're more to compare.

callicles (approved):

PR #1470, opened 2024-06-21
check for consumption API existence

phiSgr:


callicles (commented):
As talked in the review - let's change this to have the feedback in the dev logs - not in HTTP response

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -337,13 +363,16 @@ async fn ingest_route(
     }
 }
 
+#[allow(clippy::too_many_arguments)]
```
phiSgr: I still think making the `project` a global (immutable) variable is a good idea.

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -337,13 +363,16 @@ async fn ingest_route(
...
 async fn router(
     req: Request<hyper::body::Incoming>,
     current_version: String,
     route_table: &RwLock<HashMap<PathBuf, RouteMeta>>,
+    consumption_apis: &RwLock<HashSet<String>>,
     configured_producer: ConfiguredProducer,
     console_config: ConsoleConfig,
```
callicles: Did you rebase? The console config should be gone and that might enable you to not have to add the clippy argument

apps/framework-cli/src/cli/local_webserver.rs
callicles: Hum - This method doesn't hava `Project` in it - not sure that makes the point

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -116,9 +138,11 @@ async fn create_client(
 struct RouteService {
     host: String,
     route_table: &'static RwLock<HashMap<PathBuf, RouteMeta>>,
+    consumption_apis: &'static RwLock<HashSet<String>>,
```
callicles, line 140: I am not a big fan of that mutable state punched through. Not sure that there is a better solution for now though

callicles (commented):
Looks like a rebase is needed - as there are some merge conflicts as well

callicles (approved):

PR #1473, opened 2024-06-21
Add streaming topics to ls command

DatGuyJonathan:


apps/framework-cli/src/cli.rs
```diff
@@ -561,7 +565,11 @@ async fn top_command_handler(
                 &settings
             );
 
-            list_all(project_arc, version, limit).await
+            if *streaming {
+                list_streaming(project_arc, limit).await
+            } else {
+                list_db(project_arc, version, limit).await
+            }
```
phiSgr, line 598: What if the user wants to list "everything"?

apps/framework-cli/src/cli/routines/ls.rs
```diff
@@ -157,3 +187,47 @@ fn augment_output_table(
         }
     }
 }
+
+async fn get_topics(project: &Project) -> HashSet<String> {
+    let regex_filter = Regex::new(r"_\d+_\d+$").unwrap();
```
phiSgr: This feels very fragile.

Can we load the flows (including flow version syncs) and output that?

DatGuyJonathan: Normal flow topics are just the data model topics and they work fine. I haven't been able to query DCM flow topics though. Would you happen to know why?

apps/framework-cli/src/cli.rs
DatGuyJonathan: We could make --streaming flag additive instead of either-or, but would need to figure out what to show if they also add --version flag because right now we show all streaming topic versions per data model. Basically, need to iterate with product to see how this fits in their flow and if there are other ways we want to organize this information.

apps/framework-cli/src/cli/routines/ls.rs
DatGuyJonathan: This is the list from framework-internal-app before it's filtered. I supposed I could just remove __consume_offsets, but still not sure why DCM flow topics aren't showing.

PageViewEvent_0_0
UserActivity_0_0
ParsedActivity_0_0
MooseActivity_0_0
MooseActivity_0_1
MooseActivity_0_2
UserActivity_0_3
MooseActivityAugmented_0_4
MooseActivity_0_4
__consumer_offsets

callicles: Maybe we can start there and then use the infrastructure map?

phiSgr: They're named

`UserActivity_0_0_UserActivity_0_1_input`
`UserActivity_0_0_UserActivity_0_1_output`

DatGuyJonathan: @callicles , sounds good, would be great to converge on one source of truth.

@phiSgr , I changed it to a blacklist approach. LMK if that works

callicles (approved):

PR #1476, opened 2024-06-22
Generates the correctly formatted queries for clickhouse for data that contains nested objects or arrays 

camelCasedAditya:
Generates the correctly formatted queries for clickhouse for data that contains nested objects or arrays . The objects can be nested multiple times and the arrays can be of any dimension and the correct queries will be generated.


apps/framework-cli/src/infrastructure/olap/clickhouse/model.rs
```diff
@@ -109,125 +109,92 @@ pub enum ClickHouseRuntimeEnum {
...
+                "\'{}\'",
+                v.replace('\\', "\\\\").replace('\'', "\\\'").clone()
+            ),
+            ClickHouseValue::Boolean(v) => v.clone(),
+            ClickHouseValue::ClickhouseInt(v) => v.clone(),
+            ClickHouseValue::ClickhouseFloat(v) => v.clone(),
+            ClickHouseValue::Decimal => panic!("Decimals are not a supported data type"),
+            ClickHouseValue::DateTime(v) => v.clone(),
+            ClickHouseValue::Json => panic!("JSON is not a supported data type"),
+            ClickHouseValue::Bytes => panic!("Bytes are not a supported data type"),
```
callicles: We should not panic in this part of the code—if we do, the process will stop, and the following records will not be processed.

apps/framework-cli/src/infrastructure/kafka_clickhouse_sync.rs
```diff
@@ -580,29 +568,27 @@ fn map_json_value_to_clickhouse_value(
...
-                                std_column_to_clickhouse_column(col.clone()).unwrap(),
+                let mut values: Vec<ClickHouseValue> = Vec::new();
+                let _ = inner_nested.columns.iter().map(|col| {
+                    let col_name = &col.name;
+                    let val = obj.get(col_name);
+                    match val {
+                        Some(val) => (
+                            //std_column_to_clickhouse_column(col.clone()).unwrap(),
+                            values.push(
                                 map_json_value_to_clickhouse_value(&col.data_type, val).unwrap(),
```
callicles: We need to do something better than `unwrap` in this context. If there is an error we probably should log it and ignore. 

But `unwrap` will panic if the value is not the one intended. If this panics, the sync will stop and we don't want that.

It looks like we had some unwrap in the code before so this will make it more solid

apps/framework-cli/src/infrastructure/kafka_clickhouse_sync.rs
```diff
@@ -580,29 +568,27 @@ fn map_json_value_to_clickhouse_value(
...
-                        match val {
-                            Some(val) => (
-                                std_column_to_clickhouse_column(col.clone()).unwrap(),
+                let mut values: Vec<ClickHouseValue> = Vec::new();
+                let _ = inner_nested.columns.iter().map(|col| {
+                    let col_name = &col.name;
+                    let val = obj.get(col_name);
+                    match val {
+                        Some(val) => (
+                            //std_column_to_clickhouse_column(col.clone()).unwrap(),
```
callicles: You can remove the comment :)

apps/framework-cli/src/infrastructure/kafka_clickhouse_sync.rs
```diff
@@ -580,29 +568,27 @@ fn map_json_value_to_clickhouse_value(
...
-                                        .unwrap(),
-                                ),
-                            ),
-                        }
-                    })
-                    .collect();
+                            map_json_value_to_clickhouse_value(&col.data_type, val).unwrap(),
+                        ),
+                        None => (
+                            //std_column_to_clickhouse_column(col.clone()).unwrap(),
```
callicles: Same here

apps/framework-cli/src/infrastructure/olap/clickhouse/model.rs
```diff
@@ -109,125 +109,92 @@ pub enum ClickHouseRuntimeEnum {
...
-        }
+        ClickHouseValue::ClickhouseInt(format!("{}", value))
     }
 
     pub fn new_float_64(value: f64) -> ClickHouseValue {
-        ClickHouseValue {
-            value_type: ClickHouseColumnType::ClickhouseFloat(ClickHouseFloat::Float64),
-            value: format!("{}", value),
-        }
+        ClickHouseValue::ClickhouseInt(format!("{}", value))
```
callicles: This should be a `ClickhouseFloat` since f64 is a float

callicles (changes requested):

callicles (changes requested):
Some small changes to make - it is nearly there! Nice work.

Do you think you could add a small test along side this change to show some of the functionality working?

callicles:
Also it looks like the current tests are not passing

apps/framework-cli/src/infrastructure/olap/clickhouse/model.rs
```diff
@@ -109,125 +109,92 @@ pub enum ClickHouseRuntimeEnum {
...
+    ClickhouseInt(String),
+    ClickhouseFloat(String),
+    Decimal,
+    DateTime(String),
+    Json,
+    Bytes,
+    Array(Vec<ClickHouseValue>),
+    Enum(String),
+    Nested(Vec<ClickHouseValue>),
+    Null(String),
```
phiSgr: Is the enclosed string always `"NULL"`?

callicles (approved):
@phiSgr Can you double check this - this looks good to me

apps/framework-cli/src/infrastructure/olap/clickhouse/model.rs
camelCasedAditya: This is how it would look inside a query 
('T', [('A','B',[('a',[('d','e','f')],'c')],**NULL**)])

camelCasedAditya: I have removed the need to pass in a string and instead used the NULL constant

PR #1489, opened 2024-06-24
use the commit hash of the main branch when we do bump-version

phiSgr:


apps/framework-cli/src/project.rs
```diff
@@ -173,16 +169,13 @@ impl Project {
...
-            // TODO: consider putting the defaults into a source (e.g. include_str a toml file)
-            .set_default(
-                "clickhouse_config.native_port",
-                ClickHouseConfig::default().native_port,
-            )
-            .unwrap() // key in set_default is a static string, this never fails
+            .add_source(File::from_str(
+                include_str!("default_config.toml"),
+                FileFormat::Toml,
+            ))
```
phiSgr: I think this beats calling `.set_default().unwrap()` for every field we want to provide a default.
But then we provide the default values twice.

callicles: Yeah I don't know which one is more prone to errors

callicles: Do we actually need this all together? the default implementation should cover it - no?

phiSgr: Unfortunately the config loader is not that smart.

callicles: `Setting` is using the same config loader and we don't have to do this for default settings. example is the feature flags that are defaulted

phiSgr: Ah I remember, it can use the default if the whole struct is absent. But if only one field is provided it crashes.

callicles: Yeah you need to add the default pragma on the field also 

example:
https://github.com/514-labs/moose/blob/39324350287222a6887d9fffb912f4d6662dc3d3/apps/framework-cli/src/cli/settings.rs#L48-L49

phiSgr: Yeah I was hoping that the it can merge the fields in the default struct value. 🤦🏼‍♂️

callicles: With that in mind - can you remove `default_config.toml` then?

PR #1498, opened 2024-06-25
Nicolas/514 1088 hook up moose prod to the plan with infrastructure

callicles:
Sorry for the huge PR

* Added change interpreter for Redpanda
* Added change interpreter for Clickhouse
* Added change interpreter for the API
* Added change interpreter for topic to table sync processes
* Moved the `Table` object to the `core` package
* hooked up prod to have data models changes done through the new core when the feature flag is activated
* Removed unused property `file_path` in the `RouteMeta` object
* Adds Experiment on data lineage through infrastructure


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -405,6 +410,67 @@ impl Webserver {
...
+                        }
+                    }
+                }
+            }
+        });
+
+        tx
+    }
+
+    // TODO - when we retire the the old core, we should remove routeTable from the start method and using only
```
phiSgr, line 472: Do you mean reading from the channel inside the `tokio::select!` below?

DatGuyJonathan (approved):

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -247,8 +131,13 @@ impl InfrastructureMap {
     // changes several times. Something could be done to optimize this.
     // There is probably a way to make this also more generic so that we don't have very
     // similar code for each type of change.
-    pub fn diff(&self, target_map: &InfrastructureMap) -> Vec<InfraChange> {
-        let mut changes = vec![];
+    pub fn diff(&self, target_map: &InfrastructureMap) -> InfraChanges {
+        let mut changes = InfraChanges {
```
phiSgr: make `InfraChanges` `derive` `Default`

apps/framework-cli/src/cli/local_webserver.rs
callicles: There is a listener setup here: https://github.com/514-labs/moose/pull/1498/files#diff-fee871a911a583459e462af9d8c682e0f98be4006d2866ce74821b05ee304a9aR413

And we could just use that

phiSgr: i think we can do away with the lock if the listening is done in the `select` loop.

callicles: Hum if we can do that - it might be worth it too

PR #1502, opened 2024-06-25
chore: update page view event migration

callicles:
Changes the base of the migration for pageViewEvent to be 0_0


phiSgr (approved):
this is difficult to keep track of
ideally moose should do it for us

PR #1503, opened 2024-06-25
Log Collector

camelCasedAditya:
Takes in logs from a moose projects and the runs it through a flow to take the most important data and inputs it into a new table in a more readable format.


apps/framework-internal-app/app/flows/Logs/ParsedLogs/flow.ts
```diff
@@ -0,0 +1,39 @@
+import { ParsedLogs, Logs } from "../../../datamodels/logs";
+
+export default function run(source: Logs): ParsedLogs[] {
+  let returnArray: ParsedLogs[] = [];
+
+  for (let a = 0; a < source.resourceLogs.length; a++) {
+    for (let b = 0; b < source.resourceLogs[a].scopeLogs.length; b++) {
```
phiSgr: Use `for of` loop so that we don't have to worry about the indices

phiSgr: Or maybe `flatMap`

callicles (approved):

PR #1511, opened 2024-06-25
Blocks poc

DatGuyJonathan:


packages/ts-moose-lib/package.json
```diff
@@ -17,10 +17,10 @@
...
   },
   "dependencies": {
-    "@clickhouse/client-web": "^1.0.1",
-    "fastq": "^1.17.1",
-    "kafkajs": "^2.2.4",
-    "ts-patch": "^3.1.2"
+    "@clickhouse/client-web": "1.1.0",
+    "fastq": "1.17.1",
+    "kafkajs": "2.2.4",
+    "ts-patch": "~3.2.0"
```
callicles, line 23: This is probably fixing some of the failures we are seeing in the CI, right?
Thank you!

apps/framework-cli/src/framework/typescript/ts_scripts/blocks.ts
```diff
@@ -0,0 +1,109 @@
...
+
+const main = async () => {
+  const chClient = getClickhouseClient(clickhouseConfig);
+  console.log(`Connected`);
+
+  const blocksFiles = walkDir(BLOCKS_DIR_PATH, ".ts", []);
+  const numOfBlockFiles = blocksFiles.length;
+  console.log(`Found ${numOfBlockFiles} blocks files`);
+
+  const queue: queueAsPromised<BlocksQueueTask> = fastq.promise(asyncWorker, 1);
```
callicles, line 78: I wonder if this should be in TS or in the CLI.
Feels like in the CLI we could reuse some of this logic for python blocks.

You could merge as is as this is the first step. The above is about the direction

callicles (approved):

callicles:
Nice work!

apps/framework-cli/src/framework/typescript/ts_scripts/blocks.ts
DatGuyJonathan: Agreed, it would be great to not duplicate & maintain this over multiple languages. Would have to see if that means we have a process per file instead of one process to create all blocks.

packages/ts-moose-lib/package.json
DatGuyJonathan: I'm not sure about CI, but I had issues with npm link. It probably doesn't hurt to keep these the same versions as the ones in moose init.

PR #1515, opened 2024-06-27
fix: watcher hanging

callicles:
Looks like @tg339 and @phiSgr were right

This seems to fix our hanging tasks issue

### DIFF
--- a/apps/framework-cli/Cargo.lock
+++ b/apps/framework-cli/Cargo.lock
@@ -806,6 +806,15 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


tg339 (approved):

tg339 (approved):

phiSgr:
Still the explanation is not quite satisfactory. It's only blocking one thread, why can't the other threads run the other async functions?

phiSgr (approved):

phiSgr:
May as well revert my revert commit.

PR #1519, opened 2024-06-27
PY moose package skeleton

georgevanderson:


packages/py-moose-lib/requirements.txt
```diff
@@ -0,0 +1,25 @@
+certifi==2024.6.2
```
callicles, line 1: We need all of this??

callicles (approved):

packages/py-moose-lib/requirements.txt
georgevanderson: Dependencies of the build dependencies `setuptools wheel twine`

PR #1528, opened 2024-06-28
batch json array kafka sends

phiSgr:


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -316,9 +316,24 @@ async fn handle_json_array_body(
...
+        // but it does not report the error back
+        if count % 1024 == 1023 {
+            for future in temp_res {
+                res_arr.push(future.await)
+            }
+            temp_res = Vec::new();
+        }
+    }
+    for future in temp_res {
+        res_arr.push(future.await)
```
DatGuyJonathan: Appreciate the initiative @phiSgr! I tried some basic timers and there doesn't seem to be much difference. Is it because the test was local? Do you expect a difference in prod?

DatGuyJonathan:
```
diff --git a/apps/framework-cli/src/cli/local_webserver.rs b/apps/framework-cli/src/cli/local_webserver.rs
index 5d376592..50350094 100644
--- a/apps/framework-cli/src/cli/local_webserver.rs
+++ b/apps/framework-cli/src/cli/local_webserver.rs
@@ -43,6 +43,7 @@ use std::path::PathBuf;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Duration;
+use std::time::Instant;
 use tokio::net::TcpListener;
 use tokio::net::TcpStream;
 use tokio::sync::mpsc;
@@ -306,17 +307,23 @@ async fn handle_json_array_body(
     topic_name: &str,
     req: Request<Incoming>,
 ) -> Response<Full<Bytes>> {
+    let start_time = Instant::now();
+
     // TODO probably a refactor to be done here with the json but it doesn't seem to be
     // straightforward to do it in a generic way.
     let url = req.uri().to_string();
     let body = to_reader(req).await;
+    let json_parse_start_time = Instant::now();
     let parsed: Result<Vec<Value>, serde_json::Error> = serde_json::from_reader(body);
     if let Err(e) = parsed {
         return bad_json_response(e);
     }
+    let json_parse_duration = json_parse_start_time.elapsed();
+    println!("json_parse took: {} seconds", json_parse_duration.as_secs_f64());
 
     let mut res_arr = Vec::new();
     let mut temp_res = Vec::new();
+    let send_payload_start_time = Instant::now();
     for (count, payload) in parsed.ok().unwrap().into_iter().enumerate() {
         temp_res.push(send_payload_to_topic(
             configured_producer,
@@ -332,14 +339,20 @@ async fn handle_json_array_body(
             temp_res = Vec::new();
         }
     }
+    
     for future in temp_res {
         res_arr.push(future.await)
     }
+    let send_payload_duration = send_payload_start_time.elapsed();
+    println!("send_payload took: {} seconds", send_payload_duration.as_secs_f64());
 
     if res_arr.iter().any(|res| res.is_err()) {
         return internal_server_error_response();
     }
 
+    let duration = start_time.elapsed();
+    println!("handle_json_array_body took: {} seconds", duration.as_secs_f64());
+
     success_response(url)
 }
 
```

phiSgr: My bad, I was using `send_result` in `send_with_back_pressure` but not here.

DatGuyJonathan (approved):
Amazing!

callicles:
How much does this speed things up?

PR #1529, opened 2024-06-28
flat flows naming convention

phiSgr:


apps/framework-cli/src/cli/routines/flow.rs
```diff
@@ -29,9 +28,7 @@ impl FlowFileBuilder {
 
         let flow_file_path = project
             .flows_dir()
-            .join(source)
-            .join(destination)
-            .join(TS_FLOW_FILE);
+            .join(format!("{}__{}.ts", source, destination));
```
phiSgr, line 31: Double underscore just in case the user wants `_` in the data model name.
Though both TS interfaces and and Python classes are supposed to be PascalCase.

callicles:
This is a breaking change - should we make it as is or behind a feature flag?

Maybe that could be behind the `core_v2` feature flag as well?

callicles:
I should have a PR today that adds `flows` to `core_v2`

phiSgr:
not breaking. it still reads the flows in the old style directories

apps/framework-cli/src/framework/flows/loader.rs
```diff
@@ -51,27 +52,77 @@ async fn get_all_flows(
...
+                                "Fragments of file {:?} does not match the convention",
+                                source.path()
+                            );
+                        }
+                    }
+                    Some(caps) => {
+                        let flow = build_migration_flow(
+                            &source.file_name().to_string_lossy(),
+                            data_models,
+                            current_version,
```
phiSgr, line 132: Not sure what version should mean in a migration flow.

callicles: That's a good question - should we make a different type of flow that not versioned for migrations?

callicles: I don't think I have a good answer

callicles (approved):

callicles (approved):

PR #1533, opened 2024-06-29
Nicolas/514 1090 add functions to the plan

callicles:
* Adds `flows` to plan (prod and dev)
* Fixes dockerfile volume mount for redpanda
* rename `data_model::schema::DataModel` to `data_model::model::DataModel`

Tested with the demo for making sure that I don't break it when the feature flag is set to false


apps/framework-cli/src/framework/core/infrastructure/function_process.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    pub fn is_py_function_process(&self) -> bool {
+        self.executable.extension().unwrap().to_str().unwrap() == PYTHON_FILE_EXTENSION
+    }
+
+    pub fn id(&self) -> String {
+        format!(
+            "{}_{}_{}_{}",
+            self.name, self.source_topic, self.target_topic, self.version
```
DatGuyJonathan, line 110: Is this the same as Flow.id? Wondering if it's safer to have an id field and we save the value when creating a FunctionProcess because we have the Flow object at that time so both ids are in sync.

DatGuyJonathan (approved):

apps/framework-cli/src/framework/core/infrastructure/function_process.rs
callicles: As of now they are the same ids, but it is a good callout - it probably should be different since they are different objects.

apps/framework-cli/src/cli/watcher.rs
```diff
@@ -325,17 +326,11 @@ async fn watch(
     while let Some(res) = rx.recv().await {
         match res {
             Ok(debounced_events) => {
-                let events = debounced_events
-                    .into_iter()
-                    .filter(|event| !event.kind.is_access() && !event.kind.is_other())
```
phiSgr, line 330: So do we filter out `access` events in the new implementation?

callicles: It bundles events with at most 1 event a second - which included access events, etc...
That reduces the amounts of refreshes drastically.

phiSgr: I got this simply by opening a flow file with `vi`, but this behaviour is probably also present is previous versions.

```
⢺ Processing 2 Flow(s) changes from file watcher
⢺ Processing 1 Flow(s) changes from file watcher
⢺ Processing 1 Flow(s) changes from file watcher
```

Probably not be a big deal though.

phiSgr (approved):

PR #1535, opened 2024-07-01
Combine blocks funcs

DatGuyJonathan:


apps/framework-cli/src/framework/typescript/templates.rs
```diff
@@ -92,27 +89,31 @@ GROUP BY toStartOfDay(timestamp)
...
+        orderBy: "date",
+      },
+      materializedViewCreateOptions: {
+        name: MATERIALIZED_VIEW,
+        destinationTable: DESTINATION_TABLE,
+        select: SELECT_QUERY,
+      },
+      populateTableOptions: {
+        destinationTable: DESTINATION_TABLE,
+        select: SELECT_QUERY,
```
callicles: Looks like there is some repetition - for that `createAggregation` function could we consolidate some of those so that the user doesn't have to repeat?

callicles (approved):

PR #1537, opened 2024-07-01
Python blocks

DatGuyJonathan:


apps/framework-cli/src/framework/python/scripts/blocks_runner.py
```diff
@@ -0,0 +1,137 @@
...
+
+def walk_dir(dir, file_extension):
+    file_list = []
+
+    for root, dirs, files in os.walk(dir):
+        for file in files:
+            if file.endswith(file_extension):
+                file_list.append(os.path.join(root, file))
+
+    return file_list
```
callicles, line 74: Feels like some of this logic could be shared with TS right?

callicles (approved):
LGTM

apps/framework-cli/src/framework/python/scripts/blocks_runner.py
DatGuyJonathan: Yea I think so. Added a ticket for it: https://linear.app/514/issue/514-1107/investigate-one-blocks-process-per-file

PR #1540, opened 2024-07-02
Instrumenting moose and tracking number of http requests per route and its latency

camelCasedAditya:


apps/framework-cli/Cargo.toml
```diff
@@ -68,6 +68,9 @@ opentelemetry = "0.23.0"
 opentelemetry-appender-log ="0.4.0"
 walkdir = "2"
 comfy-table = "7.1.1"
+prometheus-client = "0.22.2"
+mini-redis = "0.4"
+prometheus = "0.13.4"
```
callicles: Do we need all those dependencies? Feels like `prometheus-client` should be enough, no?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -421,6 +522,19 @@ async fn router(
...
+        hyper::Method::POST => Method::POST,
+        hyper::Method::GET => Method::GET,
+        hyper::Method::PUT => Method::PUT,
+        hyper::Method::DELETE => Method::DELETE,
+        hyper::Method::HEAD => Method::HEAD,
+        hyper::Method::OPTIONS => Method::OPTIONS,
+        hyper::Method::CONNECT => Method::CONNECT,
+        hyper::Method::TRACE => Method::TRACE,
+        hyper::Method::PATCH => Method::PATCH,
+        _ => Method::OTHER,
```
callicles: that `enum` is not exhaustive? why do you need the `OTHER` branch?  

That probably would be worth putting in a function in the `metrics.rs` file

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -543,11 +735,9 @@ impl Webserver {
         route_table: &'static RwLock<HashMap<PathBuf, RouteMeta>>,
         consumption_apis: &'static RwLock<HashSet<String>>,
         project: Arc<Project>,
+        metrics: Arc<Metrics>,
     ) {
-        //! Starts the local webserver
```
callicles, line 547: Why did you remove those comments?

apps/framework-cli/src/cli/routines.rs
```diff
@@ -387,8 +389,10 @@ pub async fn start_development_mode(
     )?;
 
     info!("Starting web server...");
+    let server_config = project.http_server_config.clone();
+    let web_server = Webserver::new(server_config.host.clone(), server_config.port);
```
callicles: Those 2 lines are already higher in the file, you don't need to add them back.

That's probably an artefact of the merge conflict resolution you went through

apps/framework-cli/src/cli/routines.rs
```diff
@@ -479,9 +484,10 @@ pub async fn start_production_mode(
     }
 
     info!("Starting web server...");
-
+    let server_config = project.http_server_config.clone();
+    let web_server = Webserver::new(server_config.host.clone(), server_config.port);
```
callicles: same here

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -430,15 +544,44 @@ async fn router(
                 route.join(current_version),
                 configured_producer,
                 route_table,
+                metrics,
+                MetricsDetails {
+                    timer: now,
+                    metrics_method,
+                },
```
callicles: We should be able to remove metrics being passed in here right?

and have the measurement done at the end of this file.

Something like:
```rust
let res = match (req.method(), &route_split[..]) {
     (&hyper::Method::POST, ["ingest", _]) => {
    ....
}

metrics
        .send_data(Data::IngestHistogram((
            route.clone(),
            now.elapsed(),
            metrics_method,
        )))
        .await;

```

Once you done that you should be able to remove a bunch of code inside all of the `route` functions

apps/framework-cli/src/metrics.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub async fn receive_data(&self) -> String {
+        let (resp_tx, resp_rx) = tokio::sync::oneshot::channel::<String>();
+        self.tx.send(Data::Receive(resp_tx)).await.unwrap();
+
+        let res = resp_rx.await;
+
+        res.unwrap()
+    }
+
+    pub async fn controller(self: Arc<Metrics>, mut rx: tokio::sync::mpsc::Receiver<Data>) {
```
callicles: We could rename this to `start_listening_to_metrics` explicit names are always better than implicit or too general of names

apps/framework-cli/src/metrics.rs
```diff
@@ -0,0 +1,103 @@
...
+    DELETE,
+    HEAD,
+    OPTIONS,
+    CONNECT,
+    PATCH,
+    TRACE,
+    OTHER,
+}
+
+impl Metrics {
```
callicles, line 37: you could add a method in here that would be like so:

```rust
pub fn new() -> (Metrics, tokio::sync::mpsc::Receiver<Data>) {
        let (tx, rx) = tokio::sync::mpsc::channel(10);
        let metrics = Metrics { tx: tx };
        (metrics, rx)
}
```

That way in the code above, for creating a metric, you will be able to do:

```rust
let (metrics, rx) = Metrics::new()
metrics.start_listening_to_metrics(rx);
```

apps/framework-cli/src/metrics.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub async fn send_data(&self, data: Data) {
+        let _ = self.tx.send(data).await;
+    }
+
+    pub async fn receive_data(&self) -> String {
+        let (resp_tx, resp_rx) = tokio::sync::oneshot::channel::<String>();
+        self.tx.send(Data::Receive(resp_tx)).await.unwrap();
```
callicles: `unwrap` are a [code smell](https://www.opsera.io/blog/what-is-code-smell#:~:text=Code%20Smells%20are%20the%20traces,code%20quality%20and%20technical%20debt.) 

instead you should return and handle the `Result` properly in case there is an error

you can [read](https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html) about `?` as a way to easily handle it

apps/framework-cli/src/metrics.rs
```diff
@@ -0,0 +1,103 @@
...
+
+        let res = resp_rx.await;
+
+        res.unwrap()
```
callicles: same here

apps/framework-cli/src/metrics.rs
```diff
@@ -0,0 +1,103 @@
+use prometheus_client::metrics::family::Family;
+use prometheus_client::{
+    encoding::{text::encode, EncodeLabelSet, EncodeLabelValue},
+    metrics::histogram::Histogram,
+    registry::Registry,
+};
+use std::{path::PathBuf, sync::Arc, time::Duration};
+
+pub enum Data {
```
callicles: I would rename this to `MetricsMessage`, data is too generic, it could be anything

apps/framework-cli/src/metrics.rs
```diff
@@ -0,0 +1,103 @@
...
+    Receive(tokio::sync::oneshot::Sender<String>),
```
callicles: In that context this could be:

`GetMetricsRegistryAsString`

apps/framework-cli/src/metrics.rs
```diff
@@ -0,0 +1,103 @@
...
+    IngestHistogram((PathBuf, Duration, Method)),
```
callicles: and this one `HTTPLatency`

apps/framework-cli/src/metrics.rs
callicles: Also why does it need to be an Arc in `self` ?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -581,6 +771,7 @@ impl Webserver {
             current_version: project.cur_version().to_string(),
             configured_producer: producer,
             is_prod: project.is_production,
+            metrics: metrics.clone(),
```
callicles: Hum I see why you are cloning here, the service takes ownership of this when it probably should not. instead of passing the metrics object, you could try and pass in the `send_data` so instead so that you don't have to clone and create a duplicate of your internal metrics store

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -44,11 +46,22 @@ use std::path::PathBuf;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Duration;
+use std::time::Instant;
 use tokio::net::TcpListener;
 use tokio::net::TcpStream;
 use tokio::sync::mpsc;
 use tokio::sync::RwLock;
 
+pub struct MetricsDetails {
```
callicles: you should be able to remove this struct with some of the comments below

callicles (changes requested):
Nice work, I have a lot of feedback but that's normal, that's how you will learn!

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -162,18 +170,21 @@ impl Service<Request<Incoming>> for RouteService {
...
+            RouterRequest {
+                req,
+                route_table: self.route_table,
+            },
         ))
     }
 }
 
-fn options_route() -> Result<Response<Full<Bytes>>, hyper::http::Error> {
+async fn options_route() -> Result<Response<Full<Bytes>>, hyper::http::Error> {
```
callicles: you don't need the async change anymore here, do you?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -188,7 +199,7 @@ fn options_route() -> Result<Response<Full<Bytes>>, hyper::http::Error> {
     Ok(response)
 }
 
-fn health_route() -> Result<Response<Full<Bytes>>, hyper::http::Error> {
+async fn health_route() -> Result<Response<Full<Bytes>>, hyper::http::Error> {
```
callicles: same here

apps/framework-cli/src/cli/routines.rs
```diff
@@ -388,7 +390,7 @@ pub async fn start_development_mode(
 
     info!("Starting web server...");
     web_server
-        .start(route_table, consumption_apis, project)
+        .start(route_table, consumption_apis, project, metrics.clone())
```
callicles: Do you need the clone here? it looks like we are just passing down ownership?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -216,6 +227,20 @@ async fn log_route(req: Request<Incoming>) -> Response<Full<Bytes>> {
         .unwrap()
 }
 
+async fn metrics_route(metrics: Arc<Metrics>) -> Result<Response<Full<Bytes>>, hyper::http::Error> {
+    let response = Response::builder()
+        .status(StatusCode::OK)
+        .body(Full::new(Bytes::from(
+            match metrics.clone().receive_data().await {
```
callicles: you should not have to clone here I believe?

apps/framework-cli/src/metrics.rs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+impl Metrics {
+    pub fn new() -> (Metrics, tokio::sync::mpsc::Receiver<MetricsMessage>) {
+        let (tx, rx) = tokio::sync::mpsc::channel(32);
+        let metrics = Metrics { tx };
+        (metrics, rx)
+    }
+
+    pub async fn send_data(&self, data: MetricsMessage) {
```
callicles: maybe we rename this one `send_metric`

apps/framework-cli/src/metrics.rs
```diff
@@ -0,0 +1,106 @@
...
+        let _ = self.tx.send(data).await;
+    }
+
+    pub async fn receive_data(&self) -> Result<std::string::String, MetricsErrors> {
```
callicles: and this one `get_prometheus_metrics_string`

apps/framework-cli/src/cli/routines.rs
```diff
@@ -479,9 +482,8 @@ pub async fn start_production_mode(
     }
 
     info!("Starting web server...");
-
     web_server
-        .start(route_table, consumption_apis, project)
+        .start(route_table, consumption_apis, project, metrics.clone())
```
callicles: Same here

apps/framework-cli/src/cli.rs
```diff
@@ -625,7 +626,11 @@ pub async fn cli_run() {
 
     let cli = Cli::parse();
 
-    match top_command_handler(config, &cli.command).await {
+    let (metrics, rx) = Metrics::new();
+    let arc_metrics = Arc::new(metrics);
+    arc_metrics.clone().start_listening_to_metrics(rx).await;
```
callicles: Do you need the `clone` here? `start_listening_to_metrics` only needs a reference so it should not be taking ownership

callicles (commented):
It starts to look really clean!

Some more adjustments and I believe we will be over the finish line

callicles (approved):

PR #1544, opened 2024-07-04
Olivia/gtm 117 moose landing page

okane16:


apps/framework-landing/middleware.ts
```diff
@@ -22,37 +23,66 @@ export const config = {
...
+
+  //   if (!response.ok) {
+  //     throw new Error(
+  //       `Failed to send event: ${response.status}, ${response.url}`,
+  //     );
+  //   }
+  // } catch (error) {
+  //   console.error(error);
+  // }
+  // return NextResponse.next();
```
georgevanderson: Is this supposed to be commented? We aren't returning anything from this function

okane16: oh great catch. this was where we were sending PageViewEvents to mixpanel in parallel. Will fix

georgevanderson (approved):

PR #1545, opened 2024-07-05
Added a Prometheus histogram to track latency across all http endpoints

camelCasedAditya:
Runs a timer from when an http request is sent to when moose responds. Then prometheus histogram observes the time and places it into buckets. As long as the obeserver number is less then the buckets limit it is put into the bucket so one number can be put into multiple buckers. This PR adds this histogram tracking for all http endpoints but condensed into one Histogram.


callicles (approved):
LGTM!

PR #1548, opened 2024-07-05
feat: working consumption api with plan core

callicles:
Adds the consumption api to the core


tg339 (approved):

apps/framework-cli/src/cli/display.rs
```diff
@@ -340,6 +340,18 @@ pub fn show_changes(infra_plan: &InfraPlan) {
...
             }
+            ProcessChange::ConsumptionApiWebServer(Change::Added(_)) => {
+                infra_added("Starting Consumption WebServer...");
+            }
+            ProcessChange::ConsumptionApiWebServer(Change::Removed(_)) => {
+                infra_removed("Stoping Consumption WebServer...");
+            }
+            ProcessChange::ConsumptionApiWebServer(Change::Updated {
+                before: _,
+                after: _,
```
phiSgr: Just write `Change::Updated { .. }`?

apps/framework-cli/src/cli/watcher.rs
```diff
@@ -340,28 +339,41 @@ async fn watch(
...
+                                    Ok(())
+                                }
+                                Err(e) => {
+                                    show_message!(MessageType::Error, {
+                                        Message {
+                                            action: "\nFailed".to_string(),
+                                            details: format!("\n {}", e),
+                                        }
+                                    });
+                                    Ok(())
```
phiSgr: Both branches return `Ok(())`
what about moving that outside the pattern matching?

apps/framework-cli/src/cli/display.rs
callicles: Thanks did not know this one!

apps/framework-cli/src/cli/watcher.rs
callicles: yep - good catch

PR #1551, opened 2024-07-08
Add batch ingest to dogfood app

DatGuyJonathan:
For testing purposes


apps/framework-internal-app/package-lock.json
callicles, line 1: Should we remove this file? this is pnpm only right?

callicles (approved):

apps/framework-internal-app/package-lock.json
DatGuyJonathan: Yea, it probably got added by accident not too long ago. I used to remember not seeing a package-lock here.

PR #1552, opened 2024-07-09
rotate log per day

phiSgr:


callicles:
Should we add something for deleting logs files that are too old?

phiSgr:
fern doesn't have that out of the box

phiSgr:
but the user can specify a date pattern that updates once a month

callicles:
? what do you mean we know where it should be and it is in a dedicated `.moose` directory

callicles (approved):
thx!

PR #1561, opened 2024-07-11
feat: init git repo when creating project from a template

onelesd:
checks if new templated project is in a git repo, initializing one if not.


apps/framework-cli/src/cli.rs
```diff
@@ -156,7 +191,17 @@ async fn top_command_handler(
             // TODO: refactor this to be extracted in different functions
             match template {
                 Some(template) => {
-                    templates::generate_template(template, CLI_VERSION, dir_path).await
+                    let _ = templates::generate_template(template, CLI_VERSION, dir_path).await;
```
phiSgr: do not ignore the `Result` type. Use `?`
https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html

onelesd: how would `?` be used here? _rustc_ complains when i use `?` and suggests the `_` ignore syntax instead.

Screenshot 2024-07-11 at 12 07 50 PM

phiSgr: Right. There's a `RoutineSuccess` with the message `Run these commands to start Moose: cd {}/moose && npx @514labs/moose-cli@latest dev`.

If we don't want that message any more, change the return type of `generate_template` to `Result<(), RoutineFailure>`.

phiSgr (approved):

PR #1562, opened 2024-07-11
Metric console and moose metrics command

camelCasedAditya:
Created a ratatui terminal ui that displays average latency and the total number of requests. It also has a table that shows the path of each endpoint that a request has been sent to and its average latency and the number of requests sent to that http endpoint. The user can view this console by running moose metrics in a separate terminal window

Screenshot 2024-07-11 at 11 17 19 AM

### DIFF
--- a/apps/framework-cli/src/cli/commands.rs
+++ b/apps/framework-cli/src/cli/commands.rs
@@ -50,7 +50,9 @@ pub enum Commands {
     /// Generates missing migration files
     Generate(GenerateArgs),
     /// Bumps the version of the project
-    BumpVersion { new_version: Option<String> },
+    BumpVersion {
+        new_version: Option<String>,
+    },
     /// Clears all temporary data and stops development infrastructure
     Clean {},
     /// Transforms upstream data into materialized datasets for analysis
@@ -85,6 +87,8 @@ pub enum Commands {
         #[arg(short, long, default_value = "false")]
         streaming: bool,
     },
+
+    Metrics {},
 }
 
 #[derive(Debug, Args)]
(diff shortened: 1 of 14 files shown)


apps/framework-cli/src/cli/routines/metrics_console/run_console/handler.rs
```diff
@@ -0,0 +1,25 @@
+use crate::cli::routines::metrics_console::run_console::app::{App, AppResult};
+use crossterm::event::{KeyCode, KeyEvent, KeyModifiers};
+
+pub fn handle_key_events(key_event: KeyEvent, app: &mut App) -> AppResult<()> {
```
DatGuyJonathan, line 4: How does the user learn that these keys will exit out of the UI?

apps/framework-cli/src/cli/commands.rs
```diff
@@ -85,6 +87,8 @@ pub enum Commands {
         #[arg(short, long, default_value = "false")]
         streaming: bool,
     },
+
+    Metrics {},
```
DatGuyJonathan, line 90: We can add a quick description of this command using `/// <text>`. It will show up if they do `moose-cli help` or `moose-cli metrics help`

DatGuyJonathan (approved):

apps/framework-cli/src/cli/routines/metrics_console/run_console/app.rs
```diff
@@ -0,0 +1,62 @@
...
+pub type AppResult<T> = std::result::Result<T, Box<dyn error::Error>>;
+
+#[derive(Debug)]
+pub struct App {
+    pub running: bool,
+    pub average: f64,
+    pub total_requests: f64,
+    pub summary: Vec<(f64, f64, String)>,
+    pub starting_row: usize,
+    pub selected_row: usize,
```
phiSgr: is this always 0?

camelCasedAditya: Just removed that variable. Nice catch.

PR #1565, opened 2024-07-12
Added requests per second metric to console and adjusted metrics calculations

camelCasedAditya:
The metrics console now gets data once a second and calculates requests a second. Also I have removed the metrics endpoint from all calculations because it skews the numbers since the console is constantly making requests to it.

New UI Look:
image

### DIFF
--- a/apps/framework-cli/src/cli/routines/metrics_console/run_console.rs
+++ b/apps/framework-cli/src/cli/routines/metrics_console/run_console.rs
@@ -14,6 +14,13 @@ use app::App;
 use event::Event;
 use handler::handle_key_events;
 
+type Sender = tokio::sync::mpsc::Sender<(f64, f64, Vec<(f64, f64, String)>)>;
+type Receiver = tokio::sync::mpsc::Receiver<(f64, f64, Vec<(f64, f64, String)>)>;
+struct ConsoleChannel {
+    sender: Sender,
+    receiver: Receiver,
+}
+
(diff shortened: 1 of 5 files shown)


apps/framework-cli/src/cli/routines/metrics_console/run_console.rs
```diff
@@ -24,18 +25,35 @@ pub async fn run_console() -> app::AppResult<()> {
     let mut tui = tui::Tui::new(terminal, events);
     tui.init()?;
 
+    let (tx, mut rx) = tokio::sync::mpsc::channel::<(f64, f64, Vec<(f64, f64, String)>)>(10);
```
phiSgr: Probably better to use a struct, so that you can name the fields.

apps/framework-cli/src/cli/routines/metrics_console/run_console.rs
```diff
@@ -24,18 +25,35 @@ pub async fn run_console() -> app::AppResult<()> {
...
+
+    tokio::spawn(async move {
+        loop {
+            let (average, total_requests, summary) = client::client().await.unwrap();
```
phiSgr: Can we use a verb as the function name?

camelCasedAditya: Do you want me to wrap this into a function or rename client()?

phiSgr (approved):

PR #1569, opened 2024-07-12
Detailed metrics views for each http endpoint

camelCasedAditya:
Each http endpoint now has a detailed view which the user can access by navigating to the endpoint they want on the table and then clicking enter. In the detailed view there is metrics specific to that endpoint and a bar chart showing the number of requests under X latency in ms. There is also a sparkline graph chart that shows the number of requests per second over the past ___ seconds. The user can view a longer timeframe by expanding the console window horizontally. In the main view all endpoints currently receiving requests are colored purple, ingest and consumption endpoints colored green (not getting requests), and other endpoints colored cyan (not getting requests).

Detailed view UI:

In iterm code terminal:
image

In normal mac terminal:
image

Example of highlighted endpoints
image

### DIFF
--- a/apps/framework-cli/src/cli/local_webserver.rs
+++ b/apps/framework-cli/src/cli/local_webserver.rs
@@ -492,18 +492,13 @@ async fn router(
             .body(Full::new(Bytes::from("no match"))),
     };
 
-    if metrics_path
-        .clone()
-        .into_os_string()
-        .to_str()
-        .unwrap()
-        .contains("ingest/")
+    if metrics_path.to_str().unwrap().starts_with("ingest/")
         || metrics_path
             .clone()
             .into_os_string()
             .to_str()
             .unwrap()
-            .contains("consumption/")
+            .starts_with("consumption/")
     {
         metrics
             .send_metric(MetricsMessage::HTTPLatency((
--- a/apps/framework-cli/src/cli/local_webserver.rs
+++ b/apps/framework-cli/src/cli/local_webserver.rs
@@ -492,7 +492,19 @@ async fn router(
             .body(Full::new(Bytes::from("no match"))),
     };
 
-    if metrics_path.clone().into_os_string().to_str().unwrap() != "metrics" {
+    if metrics_path
+        .clone()
+        .into_os_string()
+        .to_str()
+        .unwrap()
+        .contains("ingest/")
+        || metrics_path
+            .clone()
+            .into_os_string()
+            .to_str()
+            .unwrap()
+            .contains("consumption/")
+    {
         metrics
             .send_metric(MetricsMessage::HTTPLatency((
                 metrics_path,
--- a/apps/framework-cli/src/cli/routines/metrics_console/run_console/app.rs
+++ b/apps/framework-cli/src/cli/routines/metrics_console/run_console/app.rs
@@ -75,6 +75,7 @@ impl App {
     pub fn req_per_sec(&mut self, new_total_requests: f64, path_metrics: &Vec<PathMetricsData>) {
         self.requests_per_sec = new_total_requests - self.total_requests;
 
+        // Initializes variables and vec for each path in summary to keep unwraps in ui.rs safe
         for path in &self.summary {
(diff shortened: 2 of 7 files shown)


apps/framework-cli/src/cli/routines/metrics_console/run_console/app.rs
```diff
@@ -1,26 +1,39 @@
...
+    pub path_detailed_data: Option<Vec<HistogramCount>>,
+    pub path_requests_per_sec: HashMap<String, f64>,
+    pub requests_per_sec_vec: HashMap<String, Vec<u64>>,
 }
 
 impl Default for App {
     fn default() -> Self {
         Self {
+            viewport: Rect::new(0, 0, 0, 0),
+            state: "main".to_string(),
```
callicles: What's this state? which panel this is on? if that's the case should this be an enum instead?

apps/framework-cli/src/cli/routines/metrics_console/run_console/ui.rs
```diff
@@ -8,116 +8,357 @@ use ratatui::{prelude::*, widgets::*};
 
 use crate::cli::routines::metrics_console::run_console::app::App;
 
-const INFO_TEXT: &str = "(q) quit | (↑) move up | (↓) move down";
+const INFO_TEXT: &str = "(q) quit | (↑) move up | (↓) move down | (enter) view endpoint details";
 
 /// Renders the user interface widgets.
 pub fn render(app: &mut App, frame: &mut Frame) {
```
callicles, line 15: This is a huge method - could you break it down in smaller ones?

apps/framework-cli/src/cli/routines/metrics_console/run_console/app.rs
```diff
@@ -53,7 +72,43 @@ impl App {
         }
     }
 
-    pub fn req_per_sec(&mut self, new_total_requests: f64) {
+    pub fn req_per_sec(&mut self, new_total_requests: f64, path_metrics: &Vec<PathMetricsData>) {
         self.requests_per_sec = new_total_requests - self.total_requests;
+
+        // Initializes variables and vec for each path in summary to keep unwraps in ui.rs safe
```
callicles, line 83: In general, convention across files will create bugs when someone else comes and changes things. People should be able to assume that they can change things locally to a function and that it won't break things in other places. Here this convention breaks that principle.

It would have been ok, in the same function to have something that assumes you can unwrap and makes sure that's safe. But that definitely should not be done cross files. That's what the types are here for.

callicles (changes requested):
Some small changes to make it good to merge!

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -492,7 +492,19 @@ async fn router(
             .body(Full::new(Bytes::from("no match"))),
     };
 
-    if metrics_path.clone().into_os_string().to_str().unwrap() != "metrics" {
+    if metrics_path
+        .clone()
+        .into_os_string()
```
phiSgr: seems like we don't need these two lines?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -492,7 +492,19 @@ async fn router(
...
+        .to_str()
+        .unwrap()
+        .contains("ingest/")
```
phiSgr: `starts_with` should be a more performant and robust check

apps/framework-cli/src/cli/routines/metrics_console/run_console/app.rs
```diff
@@ -53,7 +72,43 @@ impl App {
...
+            }
+        }
+
+        for path in &self.summary {
+            for item in path_metrics {
+                if item.path == path.path {
+                    self.path_requests_per_sec
+                        .insert(path.path.clone(), item.request_count - path.request_count);
+                }
+            }
```
phiSgr, line 100: this block seems unrelated to the outer loop.

callicles (approved):

PR #1575, opened 2024-07-15
Update manifesto-section.tsx

oatsandsugar:
Adding Joj as signatory

### DIFF
--- a/.github/workflows/format.yaml
+++ b/.github/workflows/format.yaml
@@ -9,8 +9,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: pnpm/action-setup@v4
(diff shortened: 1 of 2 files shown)


phiSgr:
Changed the CI script from `npx prettier` to `pnpm exec prettier`. The former uses the latest version, as nothing is installed. It might have different rules than the prettier in the lockfile.

From the logs of the failed run:

```
Run npx prettier --check './apps/**/*.{ts,tsx}'
npm warn exec The following package was not found and will be installed: prettier@3.3.3
Checking formatting...
[warn] apps/logs-frontend/src/components/ui/tree-view.tsx
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```

callicles (approved):

PR #1580, opened 2024-07-16
fix: TS consumption api not shutting down

callicles:
Adds SIGTERM listening

### DIFF
--- a/apps/framework-cli/src/framework/typescript/ts_scripts/consumption-api.ts
+++ b/apps/framework-cli/src/framework/typescript/ts_scripts/consumption-api.ts
@@ -88,6 +88,14 @@ const startApiService = async () => {
   console.log("Starting API service");
   const server = http.createServer(apiHandler);
 
+  process.on("SIGTERM", async () => {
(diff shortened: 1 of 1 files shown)


apps/framework-cli/src/framework/typescript/ts_scripts/consumption-api.ts
```diff
@@ -88,6 +88,14 @@ const startApiService = async () => {
   console.log("Starting API service");
   const server = http.createServer(apiHandler);
 
+  process.on("SIGTERM", async () => {
+    console.log("Received SIGTERM, shutting down...");
+    server.close(() => {
```
phiSgr, line 93: do we need a graceful shut down here?

BTW i just tested and it shut down just fine without this change
do you have the procedure to reproduce the zombie process?

callicles: I believe the step to reproduce is to change the consumption api file while dev is up

phiSgr (approved):

PR #1586, opened 2024-07-18
Bytes in and out metric tracking

camelCasedAditya:
Instruments moose using prometheus counter to track bytes in and out per sec. Bytes in and out per sec is displayed in metrics console. There is bytes in/out per sec displayed for the whole app and for each path in the detailed view.

Screenshots:

Screenshot 2024-07-18 at 11 59 26 AM

Screenshot 2024-07-18 at 11 59 39 AM

### DIFF
--- a/apps/framework-cli/src/cli/routines/metrics_console/run_console/app.rs
+++ b/apps/framework-cli/src/cli/routines/metrics_console/run_console/app.rs
@@ -1,7 +1,7 @@
 use prometheus_parse::HistogramCount;
 use ratatui::layout::Rect;
 
-use super::client::{parsing_histogram_data, ParsedMetricsData, PathMetricsData};
+use super::client::{parsing_histogram_data, ParsedMetricsData, PathBytesData, PathMetricsData};
 use std::{collections::HashMap, error};
 
 pub type AppResult<T> = std::result::Result<T, Box<dyn error::Error>>;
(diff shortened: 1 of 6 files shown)


apps/framework-cli/src/cli/routines/metrics_console/run_console/ui.rs
```diff
@@ -185,6 +194,45 @@ fn render_overview_metrics(app: &mut App, frame: &mut Frame, layout: &Rc<[Rect]>
...
+        .title(format!("Data Out: {}", {
+            let bytes = app.main_bytes_data.bytes_out_per_sec as f64;
+            if (bytes / 1000000000.0) > 1.0 {
+                format!("{:.3} GB/s", (bytes / 1000000000.0))
+            } else if (bytes / 1000000.0) > 1.0 {
+                format!("{:.3} MB/s", (bytes / 1000000.0))
+            } else if (bytes / 1000.0) > 1.0 {
+                format!("{:.3} KB/s", (bytes / 1000.0))
+            } else {
+                format!("{:3} B/s", bytes)
```
callicles: Love this! That probably can be refactored not to repeat yourself right?

apps/framework-cli/src/cli/routines/metrics_console/run_console/client.rs
```diff
@@ -76,6 +91,32 @@ pub async fn getting_metrics_data() -> Result<ParsedMetricsData> {
...
+                prometheus_parse::Value::Counter(v) => *v,
+                prometheus_parse::Value::Untyped(v) => *v,
+                _ => 0.0,
+            };
+
+            total_bytes_out += value as u64;
+
+            paths_bytes_out_vec.push(PathBytesData {
+                path: (metrics_vec[j].labels["path"]).to_string(),
+                bytes_data: value as u64,
```
callicles: In the 2 branches the logic seems to be the same but applied to a different vector?

apps/framework-cli/src/metrics.rs
```diff
@@ -16,6 +17,8 @@ pub enum MetricsErrors {
 pub enum MetricsMessage {
     GetMetricsRegistryAsString(tokio::sync::oneshot::Sender<String>),
     HTTPLatency((PathBuf, Duration, String)),
+    GetNumberOfBytesIn(PathBuf, u64),
+    GetNumberOfBytesOut(PathBuf, u64),
```
callicles: This is a `put` operation not a `get` one right?

apps/framework-cli/src/cli/routines/metrics_console/run_console/app.rs
```diff
@@ -109,6 +175,25 @@ impl App {
                     .remove(0);
             }
         }
+        for path in &self.parsed_bytes_data.previous_bytes_in {
+            for item in path_bytes_in {
```
callicles: You could have a `HashMap` instead of `Vec` here to simplify getting the rate and remove this loop

callicles (approved):

PR #1593, opened 2024-07-20
Kafka instrumentation and displaying metrics

camelCasedAditya:
Instruments moose to track the number of messages sent to kafka from each ingest point and the number of messages received from kafka for each data model. Lag and messages per second are also calculated and displayed to the user. The user can also toggle between tables and then use the arrow keys to scroll within the table they have selected

Screen Shot:

Screenshot 2024-07-19 at 11 26 53 PM

### DIFF
--- a/apps/framework-cli/src/cli/routines/metrics_console/run_console/ui.rs
+++ b/apps/framework-cli/src/cli/routines/metrics_console/run_console/ui.rs
@@ -325,7 +325,7 @@ fn render_clickhouse_sync_table(app: &mut App, frame: &mut Frame, layout: Rect) 
             table_state.select(Some(app.kafka_starting_row));
 
             let table = Table::new(rows, widths)
-                .widths(&widths)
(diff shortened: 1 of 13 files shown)


apps/framework-cli/src/cli/routines/metrics_console/run_console/ui.rs
```diff
@@ -104,64 +107,303 @@ pub fn render(app: &mut App, frame: &mut Frame) {
...
-                        "{}",
-                        (((x.request_count * 1000.0).round()) / 1000.0).to_string()
-                    ),
-                ])
-                .green()
-                .not_bold()
-            },
-        )
+    match &app.table_state {
+        TableState::Endpoint => {
```
callicles: This can probably be split into 2 functions right? one that generated the endpoint table and the other one the kafka table?

callicles (approved):

PR #1599, opened 2024-07-22
remove npm lock in framework-internal-app (we use pnpm)

phiSgr:


callicles (approved):

DatGuyJonathan:
I think this has happened accidentally a couple of times. Should we add this to .gitignore?

PR #1606, opened 2024-07-24
import command

phiSgr:


apps/framework-cli/src/cli/commands.rs
```diff
@@ -94,6 +94,15 @@ pub enum Commands {
 
     /// Opens metrics console for viewing live metrics from your moose app
     Metrics {},
+    Import {
```
callicles, line 97: maybe we need version as well?

apps/framework-cli/src/cli/commands.rs
```diff
@@ -94,6 +94,15 @@ pub enum Commands {
...
+        data_model_name: String,
+        #[arg(short, long)]
+        file: PathBuf,
+        #[arg(long)]
+        format: Option<String>,
```
callicles, line 102: Shouldn't formats be an enum here?

apps/framework-cli/src/cli.rs
```diff
@@ -631,6 +632,59 @@ async fn top_command_handler(
...
+                    .map_err(|e| {
+                        RoutineFailure::new(
+                            Message::new("Import".to_string(), "failed".to_string()),
+                            e,
+                        )
+                    })?;
+            } else {
+                return Err(RoutineFailure::error(Message::new(
+                    "Format".to_string(),
+                    "unknown".to_string(),
```
callicles: Maybe the error could be `Unknown Format, we are only compatible with 'csv'`

callicles (approved):
Could be merged as is - some comments about usability improvements

PR #1608, opened 2024-07-24
Measures and displays metrics from streaming functions kafka messages in and out

camelCasedAditya:
Tracks the number of messages in and out of each streaming function. Displays the data in a table in the ratatui ui. Also styled the ui based on the figma mockup with few changes.

Screenshot:

Overview Page:
Screenshot 2024-07-24 at 2 35 03 PM

Detailed View:
Screenshot 2024-07-24 at 2 37 19 PM

### DIFF
--- a/apps/framework-cli/src/cli/local_webserver.rs
+++ b/apps/framework-cli/src/cli/local_webserver.rs
@@ -236,6 +236,26 @@ async fn log_route(req: Request<Incoming>) -> Response<Full<Bytes>> {
         .unwrap()
 }
 
+async fn metrics_log_route(req: Request<Incoming>) -> Response<Full<Bytes>> {
+    let body = to_reader(req).await;
+    let parsed: Result<CliMessage, serde_json::Error> = serde_json::from_reader(body);
+    match parsed {
+        Ok(cli_message) => {
+            let message = Message {
+                action: cli_message.action,
+                details: cli_message.message,
+            };
+            print!("{:#?} --- {:#?}", cli_message.message_type, message);
+        }
+        Err(e) => println!("Received unknown message: {:?}", e),
+    }
+
+    Response::builder()
+        .status(StatusCode::OK)
+        .body(Full::new(Bytes::from("")))
+        .unwrap()
+}
+
 async fn metrics_route(metrics: Arc<Metrics>) -> Result<Response<Full<Bytes>>, hyper::http::Error> {
     let response = Response::builder()
         .status(StatusCode::OK)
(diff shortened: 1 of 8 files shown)


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -236,6 +243,45 @@ async fn log_route(req: Request<Incoming>) -> Response<Full<Bytes>> {
...
+    let parsed: Result<FlowMessages, serde_json::Error> = serde_json::from_reader(body);
+    match parsed {
+        Ok(cli_message) => {
+            let message = FlowMessages {
+                count: cli_message.count,
+                path: cli_message.path,
+                direction: cli_message.direction,
+            };
+            match message.direction.as_str() {
+                "in" => {
```
callicles: That probably should be an enum here 🙂

apps/framework-cli/src/cli/routines/metrics_console/run_console/client.rs
```diff
@@ -21,6 +21,8 @@ pub struct ParsedMetricsData {
     pub histogram_vec: Vec<Sample>,
     pub kafka_messages_in_total: HashMap<String, (String, f64)>,
     pub kafka_messages_out_total: Vec<(String, String, f64)>,
+    pub flows_messages_in: HashMap<String, f64>,
```
callicles: We probably should call those variables streaming_functions since that's they are now called in the documentation

apps/framework-cli/src/framework/typescript/ts_scripts/streaming-function.ts
```diff
@@ -13,6 +14,22 @@ const SASL_PASSWORD = process.argv[7];
...
+  direction: "in" | "out";
+};
+export const metricsLog: (log: CliLogData) => void = (log) => {
+  const req = http.request({
+    port: 4000,
+    method: "POST",
+    path: "/metrics-logs",
+  }); // no callback, fire and forget
+
+  req.write(JSON.stringify({ message_type: "Info", ...log }));
```
callicles: The message type here is not useful since you have a dedicated endpoint right?

apps/framework-cli/src/framework/python/scripts/streaming_function_runner.py
```diff
@@ -169,9 +189,20 @@ def parse_input(json_input):
...
+
     for item in output_data_list:
         # Ignore flow function returning null
         if item is not None:
             # send() is asynchronous. When called it adds the record to a buffer of pending record sends 
             # and immediately returns. This allows the producer to batch together individual records
             producer.send(target_topic, json.dumps(item, cls=EnhancedJSONEncoder).encode('utf-8'))
+            count_out+=1
+
+thread_running = False
```
callicles: is that a leftover?

callicles (approved):
Some small changes to do - you could do those and then merge

PR #1617, opened 2024-07-26
Generate Model CLI Command

georgevanderson:


apps/framework-cli/src/cli.rs
```diff
@@ -527,6 +529,25 @@ async fn top_command_handler(
                 }
             }
         }
+        Commands::DataModel(data_model_args) => {
+            info!("Running function command");
```
callicles: Wrong log?

apps/framework-cli/src/cli/commands.rs
```diff
@@ -190,3 +193,26 @@ pub enum ConsumptionCommands {
...
+#[derive(Debug, Args)]
+#[command(arg_required_else_help = true)]
+pub struct DataModelArgs {
+    #[command(subcommand)]
+    pub command: Option<DataModelCommands>,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum DataModelCommands {
+    /// Structures the project's directory & files for a new streaming function
```
callicles: Copy pasta?

apps/framework-cli/src/cli/routines/datamodel.rs
```diff
@@ -0,0 +1,235 @@
...
+    let mut schema = HashMap::new();
+
+    let mut i = 0;
+    for obj in stream {
+        if let Value::Array(array) = obj.unwrap() {
+            for obj in array {
+                let parsed_map = parse_json(&obj);
+                i += 1;
+                if i % 100 == 0 {
+                    println!("Processed {} records", i);
```
callicles: should we use show_message instead? that would format the output like the rest

apps/framework-cli/src/cli/routines/datamodel.rs
```diff
@@ -0,0 +1,235 @@
...
+                }
+                if i == 1000 {
+                    println!("Processed {} records", i);
```
callicles: Same here

callicles (changes requested):
Small copy pasta things that are left 😄

PR #1619, opened 2024-07-26
Refactored code to split huge struct

camelCasedAditya:


apps/framework-cli/src/cli/routines/metrics_console/run_console/app.rs
```diff
@@ -68,45 +55,88 @@ pub struct App {
...
+                path_detailed_data: None,
+                path_requests_per_sec: HashMap::new(),
+                requests_per_sec_vec: HashMap::new(),
+                parsed_bytes_data: PathBytesParsedData {
+                    previous_bytes: HashMap::new(),
+                    path_bytes_in_per_sec_vec: HashMap::new(),
+                    path_bytes_out_per_sec_vec: HashMap::new(),
+                },
+            },
+            kafka_metrics: AppKafkaMetrics {
```
callicles: Those are the sync to clickhouse metrics right?

if so - would be great to have that explicit

callicles (approved):

PR #1620, opened 2024-07-27
Nicolas/514 1093 add older versions to the plan

callicles:
* Remove some unused code for data model hash checks
* Adds older versions loading to core v2
* tidy up the core_v2 flags for when we will do the move-over. 
* Add Views to the plan
* adds retries for connecting to redpanda through RPK
* Only run docker info in dev and not in prod

Missing, to be added in a follow-up PR:
* Some optimizations for running loading data model configurations and parsing in parallel
* Initial data load for migrations.

Tested on the demo flow without the `core_v2` feature flag to make sure we don't have regressions

i.e., ran ingest / some changes in data model structure and visual visualization.


apps/framework-cli/src/framework/data_model/model.rs
```diff
@@ -34,8 +37,21 @@ impl DataModel {
         }
     }
 
+    /**
+     * This hash is used to determine if the data model has changed.
+     * It does not include the version in the hash and the abs_file_path is not included.
+     */
+    pub fn hash_without_version(&self) -> u64 {
```
phiSgr, line 44: Not convinced that hashing better than directly comparing the values.

apps/framework-cli/src/cli/watcher.rs
```diff
@@ -387,79 +387,82 @@ async fn watch(
                             format!("Processing error occurred: {}", e),
                         )
                     })?;
-                }
+                } else {
+                    let framework_object_versions = framework_object_versions.as_mut().unwrap();
```
phiSgr, line 392: ~~why is unwrap safe here?~~

maybe add a comment that `is None if using core_v2`

callicles: Because framework_object_versions is passed as Some when not core_v2.

This is temp and that code will be removed once we do the move over

apps/framework-cli/src/framework/data_model/model.rs
callicles: I wanted to make it explicit what was getting compared, I did not want to recursively compare values? maybe I over-engineered this.

We could change it back

phiSgr: no explicit recursion it think? just compare the 3 fields

phiSgr (approved):

apps/framework-cli/src/framework/data_model/model.rs
phiSgr: `Eq` does the "recursively compares the fields" for you, `Hash` does the "recursively hash the fields"

they're not that different i think

callicles: kk I will merge it for now and we can change later 😛

PR #1648, opened 2024-07-31
API Auth Ingest/Consumption

georgevanderson:
Set the hashed values as environment variables to turn on auth
MOOSE_INGEST_API_KEY
MOOSE_CONSUMPTION_API_KEY


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -530,6 +574,17 @@ async fn ingest_route(
...
 
+    let auth_header = req.headers().get(hyper::header::AUTHORIZATION);
+
+    if !check_authorization(auth_header, "MOOSE_INGEST_API_KEY").await {
+        return Response::builder()
+            .status(StatusCode::UNAUTHORIZED)
+            .body(Full::new(Bytes::from(
+                "Unauthorized: Invalid or missing token",
+            )))
+            .map_err(Into::into);
```
phiSgr: looks like we don't need `.map_err(Into::into);`

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -515,6 +530,35 @@ async fn handle_json_array_body(
...
+        },
+        Err(_) => {
+            // If API_KEY not SET
+            true
+        }
+    }
+}
+
+async fn check_authorization(auth_header: Option<&HeaderValue>, env_var: &str) -> bool {
+    let bearer_token = match auth_header {
```
phiSgr: maybe use `and_then`?

https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -515,6 +530,35 @@ async fn handle_json_array_body(
     success_response(url)
 }
 
+async fn validate_token(token: Option<&str>, env_var: &str) -> bool {
+    match env::var(env_var) {
```
phiSgr: we can probably use a `lazy_static` to cache the value

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -515,6 +530,35 @@ async fn handle_json_array_body(
...
+        Ok(ingest_api_key) => match token {
```
phiSgr: `token.is_some_and`

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -103,7 +107,18 @@ async fn create_client(
...
+    // Extract the Authorization header and check the bearer token
+    let auth_header = req.headers().get(hyper::header::AUTHORIZATION);
+
+    if !check_authorization(auth_header, "MOOSE_CONSUMPTION_API_KEY").await {
+        return Response::builder()
+            .status(StatusCode::UNAUTHORIZED)
+            .body(Full::new(Bytes::from(
+                "Unauthorized: Invalid or missing token",
+            )))
+            .map_err(Into::into);
```
phiSgr: i prefer `return Ok(...?);`

phiSgr (commented):
just some code style stuff

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -515,6 +530,35 @@ async fn handle_json_array_body(
     success_response(url)
 }
 
+async fn validate_token(token: Option<&str>, env_var: &str) -> bool {
+    match env::var(env_var) {
+        Ok(ingest_api_key) => match token {
+            Some(token) => hex_digest(Algorithm::SHA256, token.as_bytes()) == ingest_api_key,
```
callicles: We should not be using SHA-256 for this.

NIST recommends PBKDF2 - let's follow that recommendation with a constant time check on the output.

Ref: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pbkdf2

https://github.com/RustCrypto/password-hashes/tree/master/pbkdf2

callicles (changes requested):
Let's change the algorithm used for hashing

apps/framework-cli/src/cli/routines/auth.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    let key1_hex = hex::encode(key1);
+
+    // compare byte to byte to avoid timing attacks
+    let token_hash = hex::decode(private_pass).unwrap();
+    let key1_hash = hex::decode(key1_hex).unwrap();
+
+    for i in 0..token_hash.len() {
+        if token_hash[i] != key1_hash[i] {
+            return false;
```
callicles: If I am reading this correcly - you are doing an early return here - so not comparing byte. 

Doing constant time comparison is really hard - I would use a third-party to do that.

After some research I would use this: https://github.com/cesarb/constant_time_eq/tree/master

callicles (changes requested):
Sorry - last change and then we will be good to go 😄 Security is hard

callicles (approved):
Nice work!

PR #1651, opened 2024-08-01
Replace aggregation init with block init

DatGuyJonathan:


apps/framework-cli/src/cli/routines/block.rs
```diff
@@ -0,0 +1,42 @@
...
+};
+
+use super::{RoutineFailure, RoutineSuccess};
+
+pub async fn create_block_file(
+    project: &Project,
+    filename: String,
+) -> Result<RoutineSuccess, RoutineFailure> {
+    let blocks_dir = project.blocks_dir();
+    let block_file_path = blocks_dir.join(format!("{}.ts", filename));
```
phiSgr: probably worth adding the python version as well

phiSgr (approved):

PR #1674, opened 2024-08-07
Add path_prefix as env var

DatGuyJonathan:
On the hosting side: https://github.com/514-labs/commercial/pull/76

### DIFF
--- a/apps/framework-cli/src/cli/local_webserver.rs
+++ b/apps/framework-cli/src/cli/local_webserver.rs
@@ -93,7 +93,23 @@ impl LocalWebserverConfig {
     }
 
     pub fn url(&self) -> String {
-        format!("http://{}:{}", self.host, self.port)
+        let base_url = format!("http://{}:{}", self.host, self.port);
+        if let Some(prefix) = &self.path_prefix {
+            format!("{}/{}", base_url, prefix.trim_matches('/'))
+        } else {
+            base_url
+        }
+    }
+
+    pub fn normalized_path_prefix(&self) -> Option<String> {
+        self.path_prefix.as_ref().map(|prefix| {
+            let trimmed = prefix.trim_matches('/');
+            if trimmed.is_empty() {
+                prefix.to_string()
+            } else {
+                format!("/{}", trimmed)
+            }
+        })
     }
 }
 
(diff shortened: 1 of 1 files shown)


callicles (commented):
Did you check that this was not breaking metrics as well?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -805,11 +814,17 @@ impl Webserver {
         );
 
         if !project.is_production {
+            let path_prefix = project
+                .http_server_config
+                .path_prefix
+                .as_deref()
+                .map(|p| format!("/{}", p.trim_matches('/')))
+                .unwrap_or_default();
```
phiSgr: a bit weird that it's `.unwrap_or_default()` here `.unwrap_or("/".to_string())` several lines down
we can probably use the same variable.

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -627,12 +637,11 @@ async fn router(
...
 
-    let route_prefix = PathBuf::from("/");
-    let route = PathBuf::from(req.uri().path())
-        .strip_prefix(route_prefix)
-        .unwrap()
-        .to_path_buf()
-        .clone();
+    let original_path = PathBuf::from(req.uri().path());
+    let route = original_path
+        .strip_prefix(PathBuf::from(path_prefix))
```
phiSgr: do we have to validate whether the prefix ends with `/`

apps/framework-cli/src/cli/local_webserver.rs
DatGuyJonathan: Updated, lmk if there's anything else.

apps/framework-cli/src/cli/local_webserver.rs
DatGuyJonathan: Updated, lmk if there's anything else.

DatGuyJonathan:
https://github.com/user-attachments/assets/2a184243-9e85-42e3-aba2-4e7cd1fa4fc8

@callicles , metrics look good. As for sdk, I think we should be okay because the constructor takes a base url and I'd expect that to include the path prefix.

callicles (approved):

PR #1680, opened 2024-08-09
Python Auto Pub

georgevanderson:


callicles (changes requested):
we should have the python lib as part of the moose release flow. Could be In parallel but it needs to be in there to be able to benefit from the version number generation so that all packages are versioned the same

callicles (approved):

PR #1684, opened 2024-08-13
sync topic to topic in core V2

phiSgr:
for continuous migration

### DIFF
--- a/apps/framework-cli/src/framework/core/infrastructure/topic_sync_process.rs
+++ b/apps/framework-cli/src/framework/core/infrastructure/topic_sync_process.rs
@@ -81,7 +81,7 @@ impl TopicToTopicSyncProcess {
     }
 
     pub fn id(&self) -> String {
-        format!("{}_{}", self.source_topic_id, self.target_topic_id)
+        self.target_topic_id.to_string()
     }
 
     pub fn expanded_display(&self) -> String {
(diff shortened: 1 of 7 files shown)


callicles (approved):
Love how clean this is

apps/framework-cli/src/framework/core/infrastructure/topic_sync_process.rs
```diff
@@ -57,3 +65,36 @@ impl TopicToTableSyncProcess {
...
+            target_topic_id: source_for_func.id(),
+            source_primitive: PrimitiveSignature {
+                name: function.id(),
+                primitive_type: PrimitiveTypes::Function,
+            },
+        }
+    }
+
+    pub fn id(&self) -> String {
+        format!("{}_{}", self.source_topic_id, self.target_topic_id)
```
phiSgr: An example ID is `UserActivity_0_0_UserActivity_0_0_UserActivity_0_1_input`

I don't really like how this looks right now, `UserActivity_0_0` is duplicated, the `_` separation is not clear.

I prefer either the `target_topic_id` (`UserActivity_0_0_UserActivity_0_1_input`) or a different separator (`UserActivity_0_0->UserActivity_0_0_UserActivity_0_1_input`)

PR #1686, opened 2024-08-14
creates management server, adds infra-map route

phiSgr:


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -708,17 +739,11 @@ async fn router(
                 }
             }
         }
-        (&hyper::Method::POST, ["logs"]) if !is_prod => Ok(log_route(req).await),
-        (&hyper::Method::POST, ["metrics-logs"]) => {
-            Ok(metrics_log_route(req, metrics.clone()).await)
-        }
         (&hyper::Method::GET, ["health"]) => health_route(),
         (&hyper::Method::GET, ["metrics"]) => metrics_route(metrics.clone()).await,
```
callicles: We should move metrics as well

apps/framework-cli/src/cli/routines.rs
```diff
@@ -313,13 +319,17 @@ pub async fn start_development_mode(
         )
         .await?;
 
+        let infra_map: &'static RwLock<InfrastructureMap> =
```
callicles, line 322: I had not thought about - we need to keep this in memory. 

That works although I am not a big fan in that state being kept here. Not sure why though.

callicles (changes requested):
Missing update for CLI metrics client

callicles (approved):

PR #1691, opened 2024-08-14
handle duplicates in core V2

phiSgr:


apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -724,7 +724,10 @@ mod tests {
             abs_file_path: PathBuf::new(),
         };
         // Making some changes to the map
-        new_target_primitive_map.datamodels.add(new_data_model);
+        new_target_primitive_map
+            .datamodels
+            .add(new_data_model)
+            .unwrap();
```
callicles, line 730: Why do you need to add the `unwrap` here, is it safe?

phiSgr: it's a(n ignored) test

callicles (approved):

PR #1693, opened 2024-08-14
Patch to allow schema validation before completing build process

cjus:
The launch of the `moose dev` command performs schema validation via initialize_project_state()
However, the `moose build` command does not perform the same valuation.

Lines 262 call the initialize_project_state() before building the docker image.

### DIFF
--- a/apps/framework-cli/src/cli.rs
+++ b/apps/framework-cli/src/cli.rs
@@ -24,13 +24,15 @@ use routines::auth::generate_hash_token;
 use routines::datamodel::read_json_file;
 use routines::ls::{list_db, list_streaming};
 use routines::metrics_console::run_console;
-use routines::plan;
 use routines::ps::show_processes;
+use routines::{initialize_project_state, plan};
 use settings::{read_settings, Settings};
 use std::cmp::Ordering;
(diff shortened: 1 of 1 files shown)


apps/framework-cli/src/cli.rs
```diff
@@ -250,13 +253,24 @@ async fn top_command_handler(
...
+            let project_arc = Arc::new(load_project()?);
 
             check_project_name(&project_arc.name())?;
 
             let mut controller = RoutineController::new();
 
+            let route_table = HashMap::<PathBuf, RouteMeta>::new();
+            let route_table: &'static RwLock<HashMap<PathBuf, RouteMeta>> =
+                Box::leak(Box::new(RwLock::new(route_table)));
+            initialize_project_state(project_arc.clone(), &mut *route_table.write().await)
```
callicles: We are getting rid of `initialize_project_state` with corev2, we probably don't want to add a dependency on there. You could look at the Data models loader to do the same thing.

callicles (changes requested):

callicles (approved):
LGTM!

PR #1707, opened 2024-08-21
chore: harden docker images

callicles:
* Changes user pattern to not run as root user
* Use base node image instead of making our own
* Change order of instructions to maximize caching

Build is now 30s 

Screenshot 2024-08-21 at 4 32 38 PM


DatGuyJonathan (approved):
🚀

PR #1710, opened 2024-08-22
remove possibilty of default value and force user to choose language.

cjus:


callicles (approved):

callicles:
Just found this: https://github.com/514-labs/moose/blob/9e301f534b11d27b11de91774b1aef122f6dd908/apps/framework-cli/src/project.rs#L124-L128

You might want to remove it as well

PR #1711, opened 2024-08-22
Logic for overwrite flag

cjus:


apps/framework-cli/src/cli.rs
```diff
@@ -430,6 +431,23 @@ async fn top_command_handler(
...
+                    // Check if the directory contains any files or subdirectories
+                    let is_empty = std::fs::read_dir(&canonical_location)
+                        .map(|mut dir| dir.next().is_none())
+                        .unwrap_or(false);
+
+                    if !is_empty && !overwrite {
+                        return Err(RoutineFailure::error(Message {
+                            action: "Generate".to_string(),
+                            details: format!(
+                                "Directory '{}' is not empty, and overwrite flag is not set.",
```
callicles:
```suggestion
                                "Directory '{}' is not empty, and --overwrite flag is not set.",
```

callicles (approved):
LGTM

apps/framework-cli/src/cli.rs
cjus: Yes above would be a nice touch.  Adding now.

PR #1712, opened 2024-08-22
Nicolas/fix init python files

callicles:
Adds `__init__.py` files when using the empty flag

### DIFF
--- a/apps/framework-cli/src/project.rs
+++ b/apps/framework-cli/src/project.rs
@@ -235,9 +235,19 @@ impl Project {
         let app_dir = self.app_dir();
         std::fs::create_dir_all(&app_dir)?;
 
+        // TODO probably move this to the respective project language modules
+        if self.language == SupportedLanguages::Python {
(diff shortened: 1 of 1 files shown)


apps/framework-cli/src/project.rs
```diff
@@ -235,9 +235,19 @@ impl Project {
...
 
+        // TODO probably move this to the respective project language modules
+        if self.language == SupportedLanguages::Python {
+            std::fs::File::create(app_dir.join(PYTHON_INIT_FILE))?;
+        }
+
         for dir in APP_DIR_LAYOUT.iter() {
             let to_create = app_dir.join(dir);
-            std::fs::create_dir_all(to_create)?;
+            std::fs::create_dir_all(to_create.clone())?;
```
phiSgr: pass by reference and avoid the `clone`

callicles: done

phiSgr (approved):

PR #1713, opened 2024-08-22
remove Project language defaulting to TypeScript.  

cjus:
This involved adding a custom Deserializer.

### DIFF
--- a/apps/framework-cli/src/project.rs
+++ b/apps/framework-cli/src/project.rs
@@ -43,8 +43,8 @@ use crate::project::typescript_project::TypescriptProject;
 use config::{Config, ConfigError, Environment, File};
 use log::debug;
 use python_project::PythonProject;
-use serde::Deserialize;
 use serde::Serialize;
+use serde::{Deserialize, Deserializer};
 
 use crate::utilities::constants::CLI_DEV_CLICKHOUSE_VOLUME_DIR_CONFIG_SCRIPTS;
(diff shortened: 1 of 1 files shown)


apps/framework-cli/src/project.rs
```diff
@@ -110,6 +110,53 @@ pub struct Project {
...
+
+        let language_project_config = match helper.language {
+            SupportedLanguages::Typescript => {
+                // Here, you would load or initialize the TypeScript project configuration
+                LanguageProjectConfig::Typescript(TypescriptProject::new(
+                    "default_name".to_string(),
+                ))
+            }
+            SupportedLanguages::Python => {
+                // Here, you would load or initialize the Python project configuration
```
phiSgr, line 141: This looks like WIP comments. Do you mean to fix it up here or are the values populated elsewhere?

cjus: No I should remove the comments.

callicles:
I believe we can close this without merging. Here is the reasoning:

In the `load` method, on the `Project` object that loads a project from disk
* we first load the project using a default value for the config (Typescript). 
* Then we load the language specific config based on the language declared in the `moose.project.toml`
* Then we mutate what was previously loaded with the Typescript config to be the proper config

Ref: https://github.com/514-labs/moose/blob/e5cf777e4abaae5ab2b78894c28de4814c172c51/apps/framework-cli/src/project.rs#L177-L212

As such I don't think we need a custom deserializer.

cjus:
I see. Closing this PR.

PR #1716, opened 2024-08-23
ensure app exits on model error for build command

cjus:


apps/framework-cli/src/cli.rs
```diff
@@ -264,16 +264,22 @@ async fn top_command_handler(
...
-            let _result = get_all_framework_objects(
+            let result = get_all_framework_objects(
                 &project_arc.clone(),
                 &mut framework_objects,
                 &project_arc.clone().data_models_dir(),
-                // .join("separate_dir_to_test_get_all"),
                 version,
                 &aggregations,
             )
             .await;
```
phiSgr:
```
.map_err(|e| {
    RoutineFailure::new(Message { ... 
    }, e)
})?;
```

phiSgr (commented):
prefer the PR title (and thus commit message) to include the context that it's the build command

apps/framework-cli/src/cli.rs
cjus: Thanks!, I updated PR

phiSgr (approved):

PR #1721, opened 2024-09-04
fix: change docker build check for v2 method

callicles:
Siwtches method for v2 method for check


cjus (approved):
I see the changes.  No need to check cur_version and now we're using the PrimitiveMap::load.  LGTM.

phiSgr (approved):
LGTM

PR #1724, opened 2024-09-04
generate python functions

phiSgr:


apps/framework-cli/src/cli/routines/streaming.rs
```diff
@@ -1,36 +1,48 @@
...
 
 use super::{RoutineFailure, RoutineSuccess};
 
 pub struct StreamingFunctionFileBuilder {
+    language: SupportedLanguages,
     function_file_path: PathBuf,
     function_file_template: String,
 }
 
 impl StreamingFunctionFileBuilder {
```
callicles, line 24: Could you use this as an opportunity to remove the Object pattern for this routine?

callicles (approved):

apps/framework-cli/src/cli/routines/streaming.rs
phiSgr: next PR

PR #1731, opened 2024-09-09
fix: add ping to clickhouse conumption client to wake up idle cloud service

callicles:
* This should address the issue of consumption API getting hit while cloud services being idle

### DIFF
--- a/packages/ts-moose-lib/src/consumption-helpers.ts
+++ b/packages/ts-moose-lib/src/consumption-helpers.ts
@@ -8,7 +8,7 @@
  * @returns 'FLoat', 'Int', 'Bool', 'String'
  */
 
-import { ClickHouseClient } from "@clickhouse/client-web";
+import { ClickHouseClient } from "@clickhouse/client";
 
 export const mapToClickHouseType = (value: Value) => {
   if (typeof value === "number") {
--- a/packages/ts-moose-lib/src/commons.ts
+++ b/packages/ts-moose-lib/src/commons.ts
@@ -1,4 +1,4 @@
-import { createClient } from "@clickhouse/client-web";
+import { createClient } from "@clickhouse/client";
 import fs from "node:fs";
 import path from "node:path";
 import http from "http";


packages/ts-moose-lib/src/consumption-helpers.ts
```diff
@@ -151,6 +152,11 @@ export class MooseClient {
       {},
     );
 
+    // We are not using the result of the ping
+    // but this ensures that if the clickhouse cloud service is idle, we
+    // wake it up.
+    this.client.ping();
```
phiSgr, line 158: should we await the result, so that we know the service has woken up when we call `.query`?

callicles: We could, but then we will add some latency every time, and it won't guarantee that the service has woken up anyway. 

So I am thinking - this makes sure the signal is sent, maybe the next step would be to do some retries on the query?

phiSgr: i guess that depends on what happens when a request hit clickhouse when it's idle

callicles: Tell me more :) how does it depend?

phiSgr: So this code sends two requests one immediately after another (ping, then actual query).
This change makes sense if queries that reach clickhouse when it's "waking up" is handled properly, but the first query (that woke it up) is not.

callicles: Yeah - so as I understand it, the behavior is - the connection will be kept opened and help for the service to wake up and return with the response when it is woken up provided the client doesn't trigger a timeout in between.

phiSgr: so my doubt is, since we're sending two requests simultaneously, how much time do we buy to prevent the timeout?

callicles: The difference is that I am not sure the query itself wakes the service up. I *think* you have to send a ping to wake the service up. 

I am not sure about that though. 
I also can make an option in the Boreal UI to have idle time configurable. 
Which would make this a non problem. 

Ref: https://clickhouse.com/docs/knowledgebase/clickhouse_cloud_api_usage#question

phiSgr (approved):
i feel a bit uncertain, but let's just merge and see if that works

PR #1732, opened 2024-09-09
dcm doc fixes

DatGuyJonathan:


apps/framework-docs/src/pages/data-models/dcm.mdx
```diff
@@ -42,7 +42,7 @@ To ensure data consistency across versions, Moose implements continuous migratio
 
 - Data ingested into older versions is automatically migrated to the latest version.
 - This process runs in near real-time within the database.
-- Migrations are defined using SQL mapping files, which can be customized as needed.
+- Migrations are defined using TypeScript or SQL mapping files, which can be customized as needed.
```
phiSgr: we should probably stop mentioning SQL
and instead of TypeScript, maybe "in your programming language"?

phiSgr (approved):

PR #1733, opened 2024-09-10
fix api auth

georgevanderson:


callicles (approved):

callicles:
You have a lint rule that fails - otherwise looks good to me

PR #1735, opened 2024-09-10
Dynamic Logs/Metrics

georgevanderson:
Use configurable endpoint and added fields

### DIFF
--- a/apps/framework-cli/src/cli/logger.rs
+++ b/apps/framework-cli/src/cli/logger.rs
@@ -115,12 +115,6 @@ fn default_log_format() -> LogFormat {
     LogFormat::Text
 }
 
-fn default_export_to() -> Option<reqwest::Url> {
-    env::var("LOG_EXPORT_DEST")
-        .ok()
-        .and_then(|url| url.parse().ok())
-}
-
 impl Default for LoggerSettings {
     fn default() -> Self {
         LoggerSettings {
@@ -128,7 +122,7 @@ impl Default for LoggerSettings {
             level: default_log_level(),
             stdout: default_log_stdout(),
             format: default_log_format(),
-            export_to: default_export_to(),
+            export_to: None,
         }
     }
 }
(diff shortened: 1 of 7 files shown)


DatGuyJonathan (approved):

apps/framework-cli/src/cli/logger.rs
```diff
@@ -215,12 +219,30 @@ pub fn setup_logging(settings: &LoggerSettings, machine_id: &str) -> Result<(),
...
+                KeyValue::new("machine_id", String::from(machine_id)),
+            ];
+
+            let metric_labels = env::var("MOOSE_METRIC_LABELS")
+                .map(|encoded| general_purpose::STANDARD.decode(encoded))
+                .ok()
+                .and_then(|result| result.ok())
+                .and_then(|decoded| String::from_utf8(decoded).ok())
+                .and_then(|json_str| serde_json::from_str::<Value>(&json_str).ok())
+                .unwrap_or_else(|| serde_json::json!({}));
```
phiSgr:
```
            let metric_labels = env::var("MOOSE_METRIC_LABELS")
                .ok()
                .and_then(|encoded| general_purpose::STANDARD.decode(encoded).ok())
                .and_then(|json_str| serde_json::from_slice::<Value>(&json_str).ok())
                .unwrap_or_else(|| serde_json::json!({}));
```

and cache it

phiSgr: also we should probably fail instead of silently ignoring the env var if it is not correct base64

apps/framework-cli/src/metrics.rs
```diff
@@ -438,6 +446,15 @@ impl Metrics {
...
 
+        let metric_labels = cloned_metadata
+            .metric_labels
+            .clone()
+            .map(|encoded| general_purpose::STANDARD.decode(encoded))
+            .and_then(|result| result.ok())
+            .and_then(|decoded| String::from_utf8(decoded).ok())
+            .and_then(|json_str| serde_json::from_str::<Value>(&json_str).ok())
+            .unwrap_or_else(|| serde_json::json!({}));
+
```
phiSgr: duplicated logic in `apps/framework-cli/src/cli/logger.rs`?

PR #1746, opened 2024-09-16
fix moosepy lib

DatGuyJonathan:
Ran into this error trying to use the moosepy lib:

Screenshot 2024-09-16 at 4 28 10 PM

This just moves the implementation to the file that it's trying to import from.


packages/py-moose-lib/setup.py
```diff
@@ -1,6 +1,8 @@
 import sys
 from setuptools import setup, find_packages
 
+version = "0.0.0"
```
DatGuyJonathan, line 4: I was trying to make changes to py-moose-lib and use that in a local moose project. Tried to install the package from a local path, but ran into an error about not finding the version:

Screenshot 2024-09-16 at 3 15 46 PM

Seems like `setup.py` works for builds because the script gives it a version. Not sure if `pip install -e` is how others are doing this locally.

tg339 (approved):

PR #1749, opened 2024-09-17
Add streaming function to moosepy

DatGuyJonathan:
- Adds moosepy lib to default python project
- Adds Key to lib and use it in default project
- Adds StreamingFunction to lib and use it in default project

### DIFF
--- a/apps/framework-cli/src/framework/python/templates.rs
+++ b/apps/framework-cli/src/framework/python/templates.rs
@@ -39,7 +39,6 @@ setup(
         {{#each dependencies}}
         "{{{ this }}}",
         {{/each}}
-
     ],
 )
 "#;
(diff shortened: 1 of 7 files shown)


callicles (approved):

apps/framework-cli/src/framework/python/templates.rs
```diff
@@ -71,19 +65,15 @@ pub static PYTHON_BASE_STREAMING_FUNCTION_TEMPLATE: &str = r#"
...
+from moose_lib import StreamingFunction
+from typing import Optional
 from datetime import datetime
 
-@dataclass
-class Flow:
-    run: Callable
-
-def flow(activity: {{source}}) -> Optional[{{destination}}]:
+def fn(activity: {{source}}) -> Optional[{{destination}}]:
```
armanjindal, line 72: @okane16 - do we want to the default name here to be `activity`?  I've re-named it `source` in the github analytics re-write.

PR #1751, opened 2024-09-17
Add blocks to moosepy

DatGuyJonathan:
- Adds blocks to lib and use it in default project
- Adds MooseClient to consumption api example
- Adds cliLog to lib. Can be used like this:

```
from moose_lib import cli_log, CliLogData

log_data = CliLogData(action="UserLogin", message="User has logged in successfully.")
cli_log(log_data)
```

### DIFF
--- a/packages/py-moose-lib/moose_lib/main.py
+++ b/packages/py-moose-lib/moose_lib/main.py
@@ -2,7 +2,7 @@ from clickhouse_connect.driver.client import Client
 from dataclasses import dataclass
 from enum import Enum
 from string import Formatter
-from typing import Callable, Generic, List, Optional, TypeVar, Union
+from typing import Callable, Generic, Optional, TypeVar, Union
 import sys
 
 type Key[T: (str, int)] = T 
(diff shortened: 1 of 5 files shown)


phiSgr (commented):
Upper case `List` is deprecated
https://docs.python.org/3/library/typing.html#typing.List

packages/py-moose-lib/moose_lib/main.py
```diff
@@ -10,6 +11,26 @@
...
+#     JSON_ARRAY = "JSON_ARRAY"
+
+# T = TypeVar('T')
+
+# @dataclass
+# class IngestionConfig:
+#     format: Optional[IngestionFormat] = None
+
+# @dataclass
+# class StorageConfig(Generic[T]):
```
phiSgr, line 25: In TS I was trying to be cute and used `keyof T` for `order_by_fields`. Python probably can't do this, so we don't need the generic type param.

phiSgr (commented):
LGTM other than the upper case `List`

DatGuyJonathan:
> Upper case `List` is deprecated https://docs.python.org/3/library/typing.html#typing.List

Good call, I've updated it.

phiSgr (approved):

PR #1755, opened 2024-09-18
Adding Redis support (wip)

cjus:


apps/framework-cli/src/infrastructure/redis/redis_client.rs
```diff
@@ -0,0 +1,541 @@
...
+                        }
+                    }
+                }
+            }
+        }));
+
+        info!("Periodic tasks started");
+    }
+
+    pub async fn stop_periodic_tasks(&mut self) -> Result<()> {
```
callicles: This doesn't have to be async too right?

callicles (approved):

apps/framework-cli/src/infrastructure/redis/redis_client.rs
cjus: That's right.

PR #1758, opened 2024-09-19
chore: change execution model for ts node processes

callicles:
This updates the way we are running the TS processes from the CLI.

We are now running a `bin` script, which is the main entry point and runs the proper process based on a CLI argument. That script is a `node` script that was precompiled and bundled. 

That script loads `ts-node` as a loader for loading and running the user's modules. 
We ensure the user's `tsconfig.json` is taken into account from the environment variable.

As part of this, I had to switch `import` -> `require` because I could not get `ts-node` to patch the `import` from Node directly. 

This is okay since the current default project loads fine and uses imports.

Screenshot 2024-09-19 at 2 37 20 PM

The consequence should be that we can now update the dependencies for running the script without telling users what to update. They have to make sure the version of the lib matches the version of the CLI.


phiSgr:
Does calling `require` fail if the imported script does not compile/typecheck?

callicles:
Good question - checking

callicles:
Confirmed we get the TS error in the console

packages/ts-moose-lib/package.json
```diff
@@ -1,14 +1,15 @@
 {
   "name": "@514labs/moose-lib",
-  "version": "0.0.2",
+  "version": "0.0.2-rc",
```
DatGuyJonathan: Do we need to update the release script to handle these tags?

DatGuyJonathan (approved):

packages/ts-moose-lib/package.json
callicles: Nope that was just for testing - I can change that back

PR #1759, opened 2024-09-19
fix: corev2 and consumption apis

callicles:
Fixes consumptions apis with core v2

### DIFF
--- a/apps/framework-cli/src/cli/local_webserver.rs
+++ b/apps/framework-cli/src/cli/local_webserver.rs
@@ -853,6 +853,7 @@ impl Webserver {
     pub async fn spawn_api_update_listener(
         &self,
         route_table: &'static RwLock<HashMap<PathBuf, RouteMeta>>,
+        consumption_apis: &'static RwLock<HashSet<String>>,
     ) -> mpsc::Sender<ApiChange> {
         log::info!("Spawning API update listener");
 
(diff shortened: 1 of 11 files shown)


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -872,20 +873,37 @@ impl Webserver {
...
+                                    .await
+                                    .insert(api_endpoint.path.to_string_lossy().to_string());
                             }
                         }
                     }
                     ApiChange::ApiEndpoint(Change::Removed(api_endpoint)) => {
                         log::info!("Removing route: {:?}", api_endpoint.path);
-                        route_table.remove(&api_endpoint.path);
+                        match api_endpoint.api_type {
+                            APIType::INGRESS {
```
phiSgr: `APIType::INGRESS { .. } => {`

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -899,14 +917,15 @@ impl Webserver {
                                 route_table.insert(
                                     after.path.clone(),
                                     RouteMeta {
-                                        format: after.format.clone(),
+                                        // When doing Ingress API the format is always known
+                                        format: after.format.unwrap().clone(),
```
phiSgr: If so we should move format to be part of `APIType::INGRESS`

callicles: Ha yeah good point

phiSgr (approved):

PR #1760, opened 2024-09-19
Add a directory for projects built with moose to the mono-repo

armanjindal:


awesome-moose-projects/README.md
```diff
@@ -0,0 +1,6 @@
+In this directory we have some of the projects we've built with Moose. From weekend hackathons to production grade 100TB/day observability-as-a-service products, so long as data is at the core, we've got you covered. 
+
+Feel free to browse, comment and eventually contribute back to this repo. If you've got any ideas please reach out to any of us on [slack](https://join.slack.com/t/moose-community/shared_invite/zt-2fjh5n3wz-cnOmM9Xe9DYAgQrNu8xKxg), and we'd be happy to roll up our sleeves and support you at any phase while you build something awesome with Moose.
```
callicles: We might want to use this file also as an index of moose `projects`. I.e., we could have a list with links, and the links may or may not point to this repo. 

We can't give write access to anonymous users who want to add a moose app to this repo; if we don't know and trust who they are, someone ill-intentioned could use it to introduce backdoors in the framework. 

As such, some of the projects will be better hosted in people's organizations, we could keep a pointer here to those locations.

callicles (approved):

awesome-moose-projects/README.md
armanjindal: Agreed! I updated the readme and directory name to reflect that.

PR #1765, opened 2024-09-20
Python data model decorator

phiSgr:
with env trigger

### DIFF
--- a/apps/framework-cli/src/framework/core/primitive_map.rs
+++ b/apps/framework-cli/src/framework/core/primitive_map.rs
@@ -1,12 +1,14 @@
+use log::warn;
 use std::{
     collections::{HashMap, HashSet},
     path::Path,
 };
-
 use walkdir::WalkDir;
 
 use super::code_loader::MappingError;
+use crate::framework::data_model::config::DataModelConfig;
 use crate::framework::data_model::DuplicateModelError;
+use crate::framework::languages::SupportedLanguages;
 use crate::framework::{
     consumption::loader::{load_consumption, ConsumptionLoaderError},
     core::infrastructure::table::ColumnType,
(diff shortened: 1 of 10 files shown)


packages/py-moose-lib/moose_lib/main.py
```diff
@@ -2,47 +2,82 @@
...
+
+def moose_data_model(arg: Any = None) -> Any:
+    def get_file(t: type) -> Optional[str]:
+        module = sys.modules.get(t.__module__)
+        if module and hasattr(module, '__file__'):
+            return module.__file__
+        return None
+
+    def decorator(data_class: type) -> type:
+        expected_file_name = os.environ.get("MOOSE_PYTHON_DM_DUMP")
```
phiSgr, line 72: Check trigger env var so that other usages of data model files won't have weird side effects.

apps/framework-cli/src/framework/core/primitive_map.rs
```diff
@@ -207,13 +207,14 @@ impl PrimitiveMap {
         )
         .await?;
 
-        for (config_variable_name, config) in data_models_configs.iter() {
+        // TODO: remove python data models without a config
```
phiSgr, line 210: probably will refactor it from

1. get all ts/py data models
2. get all ts/py configs
3. combine

to same workflow in ts, but for python

  for each file:
  1. get data models
  2. get configs
  3. remove data models without `moose_data_model`

apps/framework-cli/src/framework/core/primitive_map.rs
```diff
@@ -175,29 +187,33 @@ impl PrimitiveMap {
         Ok(())
     }
 
+    pub fn canonicalize_model_name(name: &str) -> String {
```
phiSgr, line 190: I have suspicion that we don't need to do that

packages/py-moose-lib/moose_lib/main.py
```diff
@@ -2,47 +2,87 @@
...
+
+
+def moose_data_model(arg: Any = None) -> Any:
+    def get_file(t: type) -> Optional[str]:
+        module = sys.modules.get(t.__module__)
+        if module and hasattr(module, '__file__'):
+            return module.__file__
+        return None
+
+    def remove_null(d: dict) -> dict:
```
phiSgr, line 67: with a `null`, the serde default does not kick in

tg339 (approved):

PR #1772, opened 2024-09-24
Export Metric Events

georgevanderson:


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -68,18 +71,13 @@ pub struct RouterRequest {
...
 }
 
-#[derive(Serialize, Deserialize, Debug, Clone)]
-enum Direction {
-    In,
-    Out,
-}
-
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct FlowMessages {
```
callicles: you could remove this all together right? it is the same as `StreamingFunctionEvent`

apps/framework-cli/src/metrics_inserter.rs
```diff
@@ -0,0 +1,160 @@
...
+    metric_endpoints: Option<serde_json::Map<String, serde_json::Value>>,
+) {
+    let mut interval = time::interval(Duration::from_secs(MAX_FLUSH_INTERVAL_SECONDS));
+    let client = Client::new();
+
+    loop {
+        interval.tick().await;
+        let mut buffer_owned = buffer.lock().await;
+        if buffer_owned.is_empty() {
+            drop(buffer_owned);
```
callicles: why do you need to `drop` here?

apps/framework-cli/src/metrics_inserter.rs
```diff
@@ -0,0 +1,160 @@
...
+                if let Some(labels_obj) = &metric_labels {
+                    payload_obj.extend(labels_obj.iter().map(|(k, v)| (k.clone(), v.clone())));
+                }
+
+                let _ = client.post(route).json(payload_obj).send().await;
+            }
+        }
+
+        buffer_owned.clear();
+        drop(buffer_owned);
```
callicles: same here?

apps/framework-cli/src/metrics_inserter.rs
```diff
@@ -0,0 +1,160 @@
...
+    loop {
+        interval.tick().await;
+        let mut buffer_owned = buffer.lock().await;
+        if buffer_owned.is_empty() {
+            drop(buffer_owned);
+            continue;
+        }
+
+        for chunk in buffer_owned.chunks(MAX_BATCH_SIZE) {
+            for event in chunk {
```
callicles, line 61: How is this batching? 

I might not be reading this correctly but it looks like it is sending all the events one by one

apps/framework-cli/src/metrics_inserter.rs
```diff
@@ -0,0 +1,160 @@
...
+    buffer: BatchEvents,
+    metric_labels: Option<serde_json::Map<String, serde_json::Value>>,
+    metric_endpoints: Option<serde_json::Map<String, serde_json::Value>>,
+) {
+    let mut interval = time::interval(Duration::from_secs(MAX_FLUSH_INTERVAL_SECONDS));
+    let client = Client::new();
+
+    loop {
+        interval.tick().await;
+        let mut buffer_owned = buffer.lock().await;
```
callicles, line 52: Not sure if this was what you wanted to do, this will execute every second and will send whatever is in the buffer in chunks.

callicles (commented):
Some small clarification comments

callicles (approved):

PR #1777, opened 2024-09-25
Support jwt in python consumption api

DatGuyJonathan:
Example:

```
from moose_lib import cli_log, CliLogData, ConsumptionApiResult, JWTPayload, MooseClient, ConsumptionApiResult
from typing import Optional, Any

# curl -X GET -i "http://localhost:4000/consumption/userActivity?limit=1" -H "Authorization: Bearer <token>"

def run(client: MooseClient, params: Any, jwt: Optional[JWTPayload]):
    # Do something with the JWT payload
    cli_log(CliLogData(action="UserActivity", message=f"JWT payload: {jwt}", message_type="Info"))    
    
    limit = int(params.get('limit', [10])[0])

    # Can also just return the query result directly
    query_result = client.query(
        '''SELECT *
        FROM UserActivity_0_0
        LIMIT {limit}''',
        {
            "limit": limit
        }
    )
    
    return ConsumptionApiResult(status=200, body=query_result)
```

### DIFF
--- a/apps/framework-cli/src/framework/python/consumption.rs
+++ b/apps/framework-cli/src/framework/python/consumption.rs
@@ -18,14 +18,21 @@ pub fn run(
         .as_ref()
         .map(|jwt| jwt.secret.clone())
         .unwrap_or("".to_string());
+
     let jwt_issuer = jwt_config
         .as_ref()
         .map(|jwt| jwt.issuer.clone())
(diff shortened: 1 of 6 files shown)


callicles (approved):
I believe the tests are failing but once they are passing - good to go!

PR #1785, opened 2024-09-26
validate jwt on ingest

DatGuyJonathan:


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -533,20 +537,29 @@ lazy_static! {
...
-            });
-
-            validate_token(bearer_token, key).await
-        }
+async fn check_authorization(
+    auth_header: Option<&HeaderValue>,
+    api_key: &Option<String>,
+    jwt_config: &Option<JwtConfig>,
+) -> bool {
+    let bearer_token = auth_header
```
DatGuyJonathan, line 545: This currently prefers api key. 
- Should it be reversed? 
- Do we want the second option to have a way to override? 
- Should we support both? E.g. `X-Moose-Api-Key: key` and `Authorization: Bearer jwt-token`

apps/framework-cli/src/utilities/auth.rs
```diff
@@ -0,0 +1,23 @@
...
+    aud: String,
+}
+
+pub fn validate_jwt(token: Option<&str>, public_key: &str, issuer: &str, audience: &str) -> bool {
+    token.map_or(false, |t| {
+        let mut validation = Validation::new(Algorithm::RS256);
+        validation.set_issuer(&[issuer]);
+        validation.set_audience(&[audience]);
+        validation.set_required_spec_claims(&["iss", "aud"]);
+        validation.validate_exp = false;
```
callicles: we should validate the expiration of the token

apps/framework-cli/src/cli/local_webserver.rs
callicles: I am guessing it should be reversed.

For 2 reasons:
1. When you add the `JWT` property type, it will add optionality/endpoint, and we can add the `enforce_on_all_ingest_apis` config parameter. So, token auth could be a fallback for an API.
2.  JWT is more secure than token auth, as such, it should have a preference

callicles (changes requested):
Small change on validating the expiration of the token, the change of order of validation I don't think is required to merge

apps/framework-cli/src/utilities/auth.rs
DatGuyJonathan: Okay, I've updated the required claims

callicles (approved):

PR #1792, opened 2024-09-30
Cron handler for moose leader

cjus:


tg339 (approved):

apps/framework-cli/src/cli/routines.rs
```diff
@@ -266,6 +267,29 @@ impl RoutineController {
...
+        .map_err(|e| anyhow::anyhow!(Box::<dyn std::error::Error + Send + Sync>::from(e)))?;
+    cron_registry.load_jobs().await?;
+    cron_registry.start().await?;
+    Ok(())
+}
+
+async fn setup_redis_client(project: &Project) -> anyhow::Result<RedisClient> {
+    let mut redis_client = RedisClient::new(project.name(), project.redis_config.clone()).await?;
+    redis_client.set_leadership_callback(|| {
+        tokio::spawn(async {
```
callicles: Why do you have to put it in a `spawn` ?

apps/framework-cli/src/infrastructure/processes/cron_registry.rs
```diff
@@ -27,30 +55,123 @@ impl CronRegistry {
...
     }
 
-    pub async fn start(&self) -> Result<(), JobSchedulerError> {
-        let scheduler: Arc<Mutex<JobScheduler>> = self.scheduler.clone();
-        task::spawn(async move {
-            let scheduler = scheduler.lock().await;
-            scheduler.start().await?;
+    pub async fn load_from_file(&self, file_path: &str) -> Result<()> {
+        let toml_str = fs::read_to_string(file_path).map_err(|e| anyhow!(e))?;
+        let config: CronConfig = toml::from_str(&toml_str).map_err(|e| anyhow!(e))?;
```
callicles: Why not leverage parsed config directly from the project?

apps/framework-cli/src/infrastructure/processes/cron_registry.rs
```diff
@@ -27,30 +55,123 @@ impl CronRegistry {
...
+                        .unwrap_or("");
+
+                    let output = match extension {
+                        "js" => {
+                            info!("<cron> Executing Node.js script: {}", path);
+                            Command::new("node").arg(path).output()
+                        }
+                        "ts" => {
+                            info!("<cron> Executing TypeScript script: {}", path);
+                            Command::new("ts-node").arg(path).output()
```
callicles: I don't think `ts-node` will be in the path as a default, if you want to use it directly you would need to expose is through ts-moose-lib.

apps/framework-cli/src/infrastructure/redis/redis_client.rs
```diff
@@ -197,6 +199,14 @@ impl RedisClient {
         if self.is_leader {
             info!("Instance {} became leader", self.instance_id);
             self.renew_lock().await?;
+
+            // Call the leadership callback if set
+            if let Some(callback) = &self.leadership_callback {
+                let callback = callback.clone();
+                tokio::spawn(async move {
```
callicles, line 206: Why do we spawn here too?

apps/framework-cli/src/cli/routines.rs
cjus: It's an effort to avoid an HTTP request that blocks - this also allows multiple jobs that might execute on the same time spec to not block each other.

apps/framework-cli/src/infrastructure/processes/cron_registry.rs
cjus: I was hoping to avoid dependencies - but I'll look into this

apps/framework-cli/src/infrastructure/redis/redis_client.rs
cjus: Spawn is not strictly necessary here. - but does ensure non-blocking concurrency.  A consideration is that from a library/module perspective, it is unsure how long a created callback might be and, thus, how long it would otherwise block.

apps/framework-cli/src/cli/routines.rs
callicles: Hum that makes sense but that feels like that should be the responsibility of the Cron Registry then, no?

ie the leadership callback here would be sync (if we want to add more stuff to it later we can) and the registry would manage concurrency

apps/framework-cli/src/infrastructure/processes/cron_registry.rs
cjus: I added the use of the project object.

apps/framework-cli/src/infrastructure/processes/cron_registry.rs
```diff
@@ -27,30 +48,99 @@ impl CronRegistry {
...
+                if let Some(ref path) = script_path {
+                    // Execute the script based on file extension
+                    let extension = Path::new(path)
+                        .extension()
+                        .and_then(std::ffi::OsStr::to_str)
+                        .unwrap_or("");
+
+                    let output = match extension {
+                        "js" => Command::new("node").arg(path).output(),
+                        "ts" => Command::new("moose-exec").arg(path).output(),
```
callicles: Now that you are using this - you will probably want to add as well the .bin to the path: https://github.com/514-labs/moose/blob/main/apps/framework-cli/src/framework/typescript/bin.rs#L26-L35

callicles (commented):
Looking good!

callicles (approved):

PR #1794, opened 2024-09-30
Fix build

DatGuyJonathan:


.github/workflows/release-cli.yaml
```diff
@@ -204,6 +204,8 @@ jobs:
           target: ${{ matrix.build.TARGET }}
           args: --release --features rdkafka/cmake-build --locked --target ${{ matrix.build.TARGET }} --out dist
           sccache: "true"
+          # https://github.com/briansmith/ring/issues/1728#issuecomment-1758180655
+          manylinux: ${{ matrix.build.TARGET == 'aarch64-unknown-linux-gnu' && '2_28' || '' }}
```
DatGuyJonathan, line 208: Successful dry-run: https://github.com/514-labs/moose/actions/runs/11114420026/job/30880854917

Not clear why it started failing recently though. The error is with `ring` package, but that dependency has been there for some time: https://github.com/514-labs/moose/blob/main/Cargo.lock#L2885-L2898

Seems like others ran into something similar and suggested fix is to pin to a specific build image
`https://github.com/briansmith/ring/issues/1728#issuecomment-1758180655`
`https://github.com/letsql/letsql/pull/257`

callicles: It looks like the default should be `auto`? instead of ``?

Ref: https://github.com/PyO3/maturin-action

DatGuyJonathan: I tried `auto`, but it failed on linux-x64: https://github.com/514-labs/moose/actions/runs/11114171479/job/30880102001#step:6:6

Previously, we didn't have the manylinux argument which is similar to giving it an empty string. Seems like that's different than `auto`. 

For example, with `auto`, I saw these versions in the log: `Successfully installed cffi-1.15.1 pycparser-2.21`. Without the manylinux argument, the versions are like this: `Successfully installed cffi-1.17.1 pycparser-2.22`.

callicles (approved):

PR #1795, opened 2024-10-01
Allow RedisClient to update the CONTEXT

cjus:
Allow RedisClient to update the CONTEXT when a moose instance becomes the leader or loses leadership

This will allow other application areas to determine whether they are the leader.


cjus:
We've decided against this approach. Closing this exploratory PR.

PR #1796, opened 2024-10-02
Pass jwt to streaming function

DatGuyJonathan:
For example, with models and a streaming function like this. We can send a request to ingest with the jwt token in the auth header, and inject that into their data model field. Works similarly in python.

```
import { DataModelConfig, IngestionFormat, JWT, Key } from "@514labs/moose-lib";

interface JWTPayload {
  iss: string;
  aud: string;
  exp: number;
  context: string;
}

export interface RawData {
  id: string;
  value: string;
  jwt: JWT<JWTPayload>;
}

export const RawDataConfig: DataModelConfig<RawData> = {
  ingestion: {
    format: IngestionFormat.JSON,
  },
  storage: {
    enabled: false,
    order_by_fields: ["id"],
  },
};

export interface ParsedData {
  id: Key<string>;
  value: string;
  stringifiedJWT: string;
}
```

```
import { RawData, ParsedData } from "datamodels/models";

export default function run(source: RawData): ParsedData | null {
  return {
    id: source.id,
    value: source.value,
    stringifiedJWT: JSON.stringify(source.jwt),
  };
}
```

### DIFF
--- a/packages/ts-moose-lib/src/dataModels/dataModelTypes.ts
+++ b/packages/ts-moose-lib/src/dataModels/dataModelTypes.ts
@@ -12,6 +12,7 @@ export interface Column {
   required: boolean;
   unique: false; // what is this for?
   primary_key: boolean;
+  jwt: boolean;
   default: null;
 }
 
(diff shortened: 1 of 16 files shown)


callicles (approved):
Nice work!

packages/ts-moose-lib/src/dataModels/dataModelTypes.ts
```diff
@@ -12,6 +12,7 @@ export interface Column {
   required: boolean;
   unique: false; // what is this for?
   primary_key: boolean;
+  jwt: boolean;
```
phiSgr: I think this field should go inside `Nested`?
Because this can only be in an object type.

phiSgr (approved):

PR #1798, opened 2024-10-02
automated run of quickstart guide

phiSgr:


apps/framework-cli-e2e/test/cli.test.ts
```diff
@@ -0,0 +1,212 @@
...
+        if (devProcess!.killed) return;
+        console.error("Dev server did not start or complete in time");
+        devProcess!.kill("SIGINT");
+        reject(new Error("Dev server timeout"));
+      })();
+    });
+
+    console.log("Server started, waiting before sending test request...");
+
+    await setTimeoutAsync(2000);
```
DatGuyJonathan: Would hitting `http://localhost:4000/health` in a loop until success (with a timeout) be more accurate?

apps/framework-cli-e2e/test/cli.test.ts
```diff
@@ -0,0 +1,212 @@
...
+            activity: "click",
+          }),
+        },
+      );
+
+      if (response.ok) {
+        console.log("Test request sent successfully");
+      } else {
+        console.error("Response code:", response.status);
+        console.error(`Test request failed:`, await response.text());
```
DatGuyJonathan: Should we fail early and exit here?

apps/framework-cli-e2e/test/cli.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    // Query the database
+    const client = createClient({
+      url: "http://localhost:18123",
+      username: "panda",
+      password: "pandapass",
+      database: "local",
+    });
+
+    try {
+      const result = await client.query({
```
DatGuyJonathan, line 183: Seems like this tests flow from UA to PA. If this is the default project, there should also be a block to aggregate PA to DailyActiveUsers, then a consumption api to fetch DailyActiveUsers. Maybe we just hit the consumption API instead of querying db directly?

DatGuyJonathan (approved):

apps/framework-cli-e2e/test/cli.test.ts
phiSgr: Actually we don't need the wait.

PR #1803, opened 2024-10-03
update py min version in doc

DatGuyJonathan:


tg339 (approved):

packages/ts-moose-lib/src/commons.ts
```diff
@@ -61,7 +61,7 @@ export const getClickhouseClient = ({
   });
 };
 
-type CliLogData = {
+export type CliLogData = {
```
DatGuyJonathan, line 64: Just making it consistent with python lib

PR #1807, opened 2024-10-03
Check python version

DatGuyJonathan:
For example, if I have a lower version than required:

unsupportedPyVersion

### DIFF
--- a/apps/framework-cli/src/framework/core/check.rs
+++ b/apps/framework-cli/src/framework/core/check.rs
@@ -1,7 +1,8 @@
-use async_trait::async_trait;
-use tokio::process::Command;
-
 use crate::project::LanguageProjectConfig;
+use async_trait::async_trait;
+
+use crate::framework::python::checker::PythonChecker;
+use crate::framework::typescript::checker::TypeScriptChecker;
 
 #[derive(Debug, thiserror::Error)]
 pub enum CheckerError {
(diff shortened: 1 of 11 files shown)


apps/framework-cli/src/framework/core/check.rs
```diff
@@ -0,0 +1,110 @@
...
+pub struct TypeScriptChecker;
+
+#[async_trait]
+impl SystemChecker for TypeScriptChecker {
+    async fn validate(&self) -> Result<(), CheckerError> {
+        Ok(())
+    }
+}
+
+async fn check_python_version(required_version: &str) -> Result<(), CheckerError> {
```
callicles: Feels like that should be in the `python` module since that's python specific

apps/framework-cli/src/framework/core/check.rs
```diff
@@ -0,0 +1,110 @@
...
+impl SystemChecker for PythonChecker {
+    async fn validate(&self) -> Result<(), CheckerError> {
+        check_python_version(&self.required_version).await
+    }
+}
+
+pub struct TypeScriptChecker;
+
+#[async_trait]
+impl SystemChecker for TypeScriptChecker {
```
callicles: This should be in the `typescript` module

apps/framework-cli/src/framework/core/check.rs
```diff
@@ -0,0 +1,110 @@
...
+pub trait SystemChecker {
+    async fn validate(&self) -> Result<(), CheckerError>;
+}
+
+pub struct PythonChecker {
+    pub required_version: String,
+}
+
+#[async_trait]
+impl SystemChecker for PythonChecker {
```
callicles: and this in the `python` module

callicles (commented):
Small moving of code to make sure we keep language specific parts in the same modules

apps/framework-cli/src/framework/core/check.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    for (current, required) in current_parts.iter().zip(required_parts.iter()) {
+        match current.cmp(required) {
+            std::cmp::Ordering::Greater => return true,
+            std::cmp::Ordering::Less => return false,
+            std::cmp::Ordering::Equal => continue,
+        }
+    }
+
+    current_parts.len() >= required_parts.len()
```
phiSgr: I think you can directly `cmp` the `Vec`s. It does lexicographical comparisons, which i think is what you're doing here.

apps/framework-cli/src/framework/core/check.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+pub struct Checker;
+
+impl Checker {
+    pub async fn validate_system_reqs(config: &LanguageProjectConfig) -> Result<(), CheckerError> {
```
phiSgr: it's already in the `check::` namespace. I don't see the need to wrap it inside `Checker`

apps/framework-cli/src/framework/core/check.rs
```diff
@@ -0,0 +1,110 @@
...
+            current_version, required_version
+        )))
+    }
+}
+
+fn compare_versions(current_version: &str, required_version: &str) -> bool {
+    let current_parts: Vec<u32> = current_version
+        .split('.')
+        .map(|s| s.parse().unwrap_or(0))
+        .collect();
```
phiSgr: This parsing logic exists in `src/framework/versions.rs`

phiSgr (approved):

PR #1813, opened 2024-10-04
generate infra map on build

phiSgr:


callicles (approved):
I love how simple that is

PR #1817, opened 2024-10-06
Multi-moose data migration increment

cjus:
This PR explores using the Redis-based leadership lock to ensure that only one moose instance in a multi-moose cluster can execute an initial migration. The `execute_initial_infra_change` seems like a good place to explore this approach.

However, it seems that the entire concept of how "plans" are handled needs to be considered in the multi-moose scenario. So, in this regard, this initial PR is arguably incomplete. I wanted to float this by you guys to get your take on the general approach/direction.

### DIFF
--- a/apps/framework-cli/src/cli/routines.rs
+++ b/apps/framework-cli/src/cli/routines.rs
@@ -375,6 +375,7 @@ pub async fn start_development_mode(
             api_changes_channel,
             metrics.clone(),
             &mut client,
+            &redis_client,
         )
         .await?;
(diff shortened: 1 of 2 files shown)


callicles (commented):
Lol I did not think it could look that easy.

One thing that I just remembered that I forgot to mention (you might have talked about it with @phiSgr ) is how syncing to the DB needs to be stopped while migrations are happening for tables that are getting migration data (so that we don't have duplicates)

cjus:
I am requesting that this PR be merged so I can build upon it.  As it stands, this PR ensures that only the leader can initiate a data migration.

callicles (approved):

PR #1818, opened 2024-10-07
deserialize nested for python streaming function

DatGuyJonathan:
So in the streaming function, we can do something like `activity.child.value` instead of `activity.child['value']`

### DIFF
--- a/apps/framework-cli/src/framework/python/scripts/streaming_function_runner.py
+++ b/apps/framework-cli/src/framework/python/scripts/streaming_function_runner.py
@@ -122,8 +122,6 @@ def parse_input(json_input):
         if dataclasses.is_dataclass(cls):
             field_types = {f.name: f.type for f in dataclasses.fields(cls)}
             return cls(**{name: deserialize(data.get(name), field_types[name]) for name in field_types})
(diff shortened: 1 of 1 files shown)


apps/framework-cli/src/framework/python/scripts/streaming_function_runner.py
```diff
@@ -118,7 +118,19 @@ def get_max_message_size(config_json: str) -> int:
 
 # parse json into the input type
 def parse_input(json_input):
-    return run_input_type(**json_input)
+    def deserialize(data, cls):
+        if dataclasses.is_dataclass(cls):
+            field_types = {f.name: f.type for f in dataclasses.fields(cls)}
+            return cls(**{name: deserialize(data.get(name), field_types[name]) for name in field_types})
+        elif isinstance(data, dict):
+            return {k: deserialize(v, cls) for k, v in data.items()}
```
phiSgr: Not sure if this is useful.

phiSgr (approved):
i was looking at the library `dataclasses_json` (for field name remap)
if we use that we won't have to write this ourselves

PR #1819, opened 2024-10-08
fix default features setting

DatGuyJonathan:


cjus (approved):
Looks good.

callicles (commented):
Can you explain what was wrong with the current version?

phiSgr (approved):

DatGuyJonathan:
@callicles I believe it's because `serde(default)` for the Features struct doesn't respect individual field-level `#[serde(default = "...")]`, so it kept setting `core_v2` as false when I didn't have it in config.toml

PR #1824, opened 2024-10-10
Set query_id to be the consumption path

georgevanderson:


callicles (approved):
This is for enabling linking clickhouse monitoring?

PR #1826, opened 2024-10-11
Redis message listener

cjus:


apps/framework-cli/src/cli/routines.rs
```diff
@@ -271,16 +271,46 @@ impl RoutineController {
 async fn setup_redis_client(project: Arc<Project>) -> anyhow::Result<RedisClient> {
     let mut redis_client = RedisClient::new(project.name(), project.redis_config.clone()).await?;
 
+    show_message!(
```
callicles, line 283: Do the user need to see the messages or is it for us moose developers? if it is the latter I could use `log::info!`

apps/framework-cli/src/cli/routines.rs
```diff
@@ -585,7 +662,7 @@ pub async fn start_production_mode(
             api_changes_channel,
             metrics.clone(),
             &mut client,
-            &redis_client,
+            &*redis_client.lock().await,
```
callicles: why does the redis client needs to be locked here?

apps/framework-cli/src/cli/routines.rs
```diff
@@ -375,7 +449,7 @@ pub async fn start_development_mode(
...
```
callicles: and here?

callicles (commented):
Small comment about some potential extra locks

apps/framework-cli/src/cli/routines.rs
cjus: It's for safety reasons. It seems safer to hand execute_initial_infra_change() a locked redis_client here rather than expect the redis_client to be locked in the execute_initial_infra_change() function. The danger is that the redis_client is a shared resource that can be accessed elsewhere, this at least expresses the intent more safely.

apps/framework-cli/src/cli/routines.rs
cjus: I think it's helpful for the developer to see this message as it's helpful for local debugging with a local Redis UI/command line client.

apps/framework-cli/src/cli/routines.rs
cjus: yes same as above - one for dev start and one for prod start.

apps/framework-cli/src/cli/routines.rs
callicles: You might not be passing the lock here - since the lock is released when the variable goes out of scope. 
If you want to pass in the locked mutex, you need to pass in the MutexGuard.

Feels like we should minimize the redis client locking to make the client as available as possible

callicles (approved):
LGTM

PR #1827, opened 2024-10-15
feat: enable in place table and topic edition in production

callicles:
* prevent panic when error on production start
* Adds alter table statement generation when trying to push to production a table change
* Adds topic configuration change when pushing to production


apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -28,6 +28,15 @@ pub struct PrimitiveSignature {
     pub primitive_type: PrimitiveTypes,
 }
 
+impl Default for PrimitiveSignature {
```
phiSgr: not sure if we should have this

apps/framework-cli/src/framework/core/plan.rs
```diff
@@ -44,8 +44,8 @@ pub async fn plan_changes(
     client: &mut ClientHandle,
     project: &Project,
 ) -> Result<InfraPlan, PlanningError> {
-    let mut target_infra_map = if project.is_production {
-        let json_path = Path::new(".moose/infrastructure_map.json");
+    let json_path = Path::new(".moose/infrastructure_map.json");
+    let mut target_infra_map = if project.is_production && json_path.exists() {
```
phiSgr, line 49: this is for running prod locally i guess? i worry that this might hide a bug

apps/framework-cli/src/infrastructure/olap/clickhouse.rs
```diff
@@ -71,30 +72,12 @@ pub async fn execute_changes(
...
-                    // In production - ideally we would run an alter statement if possible. Current functionlity is that
-                    // this should not happen - so we error out. This is to prevent accidental data loss
-                    // We should complain at the planning level an actually never get here.
-
-                    return Err(ClickhouseChangesError::NotSupported(format!(
-                        "Table update to {} are not supported in production",
-                        after.id()
-                    )));
+                log::info!("Updating table: {:?} to {:?}", before.id(), after.id());
+                let table_diff = compute_table_diff(before, after);
```
phiSgr: how do we prevent data loss in prod?

apps/framework-cli/src/infrastructure/stream/redpanda.rs
```diff
@@ -48,15 +48,9 @@ pub async fn execute_changes(
...
-                if !project.is_production {
-                    info!("Replacing topic: {:?} with: {:?}", before, after);
-                    delete_topics(&project.redpanda_config, vec![before.id()]).await?;
-                    create_topics(&project.redpanda_config, vec![after.id()]).await?;
-                } else {
-                    return Err(RedpandaChangesError::NotSupported(format!(
-                        "Updating topic {} is not supported in production mode",
-                        before.id()
-                    )));
+                if before.retention_period != after.retention_period {
```
phiSgr, line 51: i worry that we might add another config to topic, but forgot to add the condition to this check

apps/framework-cli/src/framework/core/plan.rs
callicles: Yes that's the reason, I mean we could add a log if you want.

phiSgr: probably add an "assert i'm not inside docker" if `is prod` && `json not exists`

apps/framework-cli/src/infrastructure/stream/redpanda.rs
callicles: Cool - will refactor with a diff on the field called in the topic

callicles: Actually, while working on the code, I also noticed that the update_topic_config is very specific to Kafka and would need to be updated too. 
I don't think we should be moving that implementation close to the `Topic` implementation since it is specify to Kafka/Redpanda, and thus should stay in this file.

Do you have a suggestion of a pattern that could help achieve this?

callicles:
@phiSgr Changed the structure of the changes to make sure they are part of the plan and that we will be able to visualize them as such. I believe I also addressed most of your comments

phiSgr (approved):

PR #1838, opened 2024-10-17
Redis infra map

phiSgr:


apps/framework-cli/build.rs
```diff
@@ -0,0 +1,16 @@
+use std::io::Result;
+
+fn main() -> Result<()> {
+    println!("cargo:rerun-if-changed=../../packages/protobuf");
```
phiSgr, line 4: it's time like this that makes me appreciate gradle's tracking of input/output change

callicles: 😆

apps/framework-cli/src/cli/routines.rs
```diff
@@ -387,6 +387,10 @@ pub async fn start_development_mode(
             &plan_result.target_infra_map,
         )
         .await?;
+        plan_result
+            .target_infra_map
+            .store_in_redis(&redis_client)
+            .await?;
```
callicles, line 468: Are you thinking - Add things to redis and deprecate the clickhouse side once redis is working properly?

callicles (approved):
Looks good to me!

apps/framework-cli/src/infrastructure/redis/redis_client.rs
```diff
@@ -399,6 +399,20 @@ impl RedisClient {
             .context("Failed to ping Redis server")?;
         Ok(())
     }
+
+    pub async fn set<K: ToRedisArgs + Send + Sync, V: ToRedisArgs + Send + Sync>(
```
cjus: I'm concerned about this function.  It accepts a key that may not strictly adhere to the naming conventions established by the RedisClient. I.e; this may break namespacing, which may complicate cleanup.

cjus: Also, I think this function should accept an optional value for key expiration - which will help with cleanup.

cjus (changes requested):
Let's discuss my comments.

apps/framework-cli/src/infrastructure/redis/redis_client.rs
```diff
@@ -386,6 +388,37 @@ impl RedisClient {
...
+        self.connection
+            .lock()
+            .await
+            .set::<K, V, ()>(key, value)
+            .await
+            .context("Failed to set value in Redis")?;
+        Ok(())
+    }
+
+    pub async fn set_infrastructure_map(
```
cjus: Does set_infrastructure_map need to be in the RedisClient.  It would be the only application-specific member in the module.  Given that it's just setting a key, a more generic set key function would be better.

cjus (commented):
Just a comment on the adding infra map member to the RedisClient.

apps/framework-cli/src/cli/routines.rs
phiSgr: I like that we have the historical versions in clickhouse, so i guess, might to worth doing, but not urgently.

apps/framework-cli/src/infrastructure/redis/redis_client.rs
phiSgr: Yeah we either have the keys for all operations in the `RedisClient` and hide the `set` function, or we expose the `set` function and hope they use the `prefix` functions.

cjus: Not sure.  I think the prefix in the RedisClient establishes a global namespace partition for all the keys that Moose uses.  You could have a generic function in RedisClient that performs the desired set operation that includes, via function signature, a string that appends one or more application-specific key segments.  See how `register_lock` works.

cjus (commented):
Offline discussion occurred. Looks good for now.  However, we should replace the explicit key formatting with calls for various Redis operations like presence_update, register_lock, send_message, broadcast_message, get_queue_message, post_queue_message, mark_queue_message, and start_message_listener. Then, the code would be consistent and more flexible and maintainable going forward.

PR #1846, opened 2024-10-22
fix: proper serilization of types using clickhouse serialization function

callicles:


phiSgr (approved):

cjus (approved):
Changes look responsible

PR #1849, opened 2024-10-23
Handle Input only streaming functions

georgevanderson:


apps/framework-cli/src/project.rs
```diff
@@ -591,7 +591,7 @@ impl Project {
                         functions_map
                             .entry(input.to_string())
                             .or_insert_with(Vec::new)
-                            .push(output.to_string())
+                            .push(output.unwrap().to_string())
```
phiSgr: This unwrap will panic?

apps/framework-cli/src/infrastructure/processes/functions_registry.rs
```diff
@@ -92,7 +92,7 @@ impl FunctionProcessRegistry {
                 let function_process = FunctionProcess::from_migration_function(
                     streaming_function,
                     &source_topic,
-                    &target_topic,
+                    &target_topic.unwrap(),
```
phiSgr, line 96: maybe add a comment here saying why this is safe

phiSgr (approved):

PR #1855, opened 2024-10-29
add jwt ingest flag

DatGuyJonathan:
Would allow the user to build jwt support piecemeal (i.e. handle consumption first, then ingest)

### DIFF
--- a/apps/framework-cli/src/cli/local_webserver.rs
+++ b/apps/framework-cli/src/cli/local_webserver.rs
@@ -202,7 +202,10 @@ async fn create_client(
         .status(status)
         .header("Access-Control-Allow-Origin", "*")
         .header("Access-Control-Allow-Method", "GET, POST")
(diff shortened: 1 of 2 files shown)


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -202,7 +202,10 @@ async fn create_client(
         .status(status)
         .header("Access-Control-Allow-Origin", "*")
         .header("Access-Control-Allow-Method", "GET, POST")
-        .header("Access-Control-Allow-Headers", "Content-Type")
+        .header(
+            "Access-Control-Allow-Headers",
+            "Authorization, Content-Type",
+        )
```
DatGuyJonathan, line 208: I guess this is a follow-up to https://github.com/514-labs/moose/pull/1326 ?

I needed to allow `Authorization` header to get my frontend to use auth with moose.

callicles (approved):

PR #1864, opened 2024-10-30
At least once delivery for Kafka <> Clickhouse sync

callicles:
* Adds at least once delivery
* Adds tests for the inserter

### DIFF
--- a/apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
+++ b/apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
@@ -1,76 +1,345 @@
-use crate::infrastructure::olap::clickhouse::client::ClickHouseClient;
+use crate::infrastructure::olap::clickhouse::client::ClickHouseClientTrait;
 use crate::infrastructure::olap::clickhouse::model::ClickHouseRecord;
+use std::collections::{HashMap, VecDeque};
 use std::time::Duration;
 
 use log::{debug, error};
+use rdkafka::error::KafkaError;
 use std::sync::Arc;
 use tokio::sync::Mutex;
 use tokio::time;
 
-use super::config::ClickHouseConfig;
-
 const MAX_FLUSH_INTERVAL_SECONDS: u64 = 1;
 const MAX_BATCH_SIZE: usize = 100000;
 
-pub type BatchRecords = Arc<Mutex<Vec<ClickHouseRecord>>>;
-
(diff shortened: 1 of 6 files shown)


apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
```diff
@@ -1,76 +1,308 @@
...
+
+            if let Some(batch) = queue.front() {
+                let batch_size = batch.records.len();
+
+                match self.client.insert(&table, &columns, &batch.records).await {
+                    Ok(_) => {
+                        debug!("Inserted {} records", batch_size,);
+
+                        for (partition, offset) in &batch.partition_offsets {
+                            let _ = commit_callback(*partition, *offset).map_err(|err| {
```
phiSgr: i'm surprised to see that this is not an async function.

BTW `if let Err(e) =` looks nicer than `map_err` and discarding the result.

apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
```diff
@@ -1,76 +1,308 @@
...
 
-pub type BatchRecords = Arc<Mutex<Vec<ClickHouseRecord>>>;
+pub type OffsetCommitCallback = Box<dyn Fn(i32, i64) -> Result<(), KafkaError> + Send + Sync>;
+pub type BatchQueue = Arc<Mutex<VecDeque<Batch>>>;
 
-pub struct Inserter {
-    buffer: BatchRecords,
+pub struct Inserter<C: ClickHouseClientTrait + 'static> {
+    queue: BatchQueue,
+    client: Arc<C>,
```
phiSgr, line 32: we can probably save an `Arc` by pointing to the `Inserter` rather than the `Inserter` having two pointers.

callicles: Feels like that's ok as is, to do what you are suggesting we would need to pull the `flush` function outside of the `Inserter`. We could iterate there if we see this as an issue.

apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
callicles: The `store` function in the kafka lib is synchronous.
Changed the error pattern matching

phiSgr (approved):

PR #1873, opened 2024-11-12
global variable to pause kafka clickhouse sync

phiSgr:
next is somehow to make leader know all followers have stopped syncing and migration is now safe to perform

---

`iterate_subscriber` was a premature abstraction. `select` has to be used to jump out of the `recv` call in one but not the other.


callicles:
Should we review?

DatGuyJonathan (approved):
This is freaking cool!

PR #1876, opened 2024-11-13
peek command

DatGuyJonathan:


apps/framework-cli/src/cli/routines/peek.rs
```diff
@@ -0,0 +1,96 @@
...
+                "Failed".to_string(),
+                "No state found".to_string(),
+            ))
+        })?;
+
+    let table = infra
+        .tables
+        .iter()
+        .find(|(key, _)| key.starts_with(&data_model_name))
+        .map(|(_, table)| table)
```
phiSgr: Maybe we should try exact match before searching by prefix.
Probably do some logging/printing like "{} not found, showing {}".
If we try to be helpful we should probably add case insensitivity and ignore underscores.

BTW `find_map`

apps/framework-cli/src/cli/routines/peek.rs
```diff
@@ -0,0 +1,96 @@
...
+            "Failed".to_string(),
+            "Error selecting data".to_string(),
+        ))
+    })?;
+
+    let mut success_count = 0;
+
+    while let Some(result) = stream.next().await {
+        if let Ok(value) = result {
+            if let Ok(json) = serde_json::to_string(&value) {
```
phiSgr: im pretty sure that calling `to_string` on a `serde_json::Value` is never gonna fail.

phiSgr (commented):
probably worth trying to reduce duplicate code with `select_all_as_json`

apps/framework-cli/src/cli/routines/peek.rs
phiSgr: to do exact match in the infra map, suffix the name with version twice (weird, i know)

https://github.com/514-labs/moose/pull/1877/files#diff-67db23e6b214a8a103adf10ce2ad885081921452099eaee572ccd491f65b4241R64

PR #1877, opened 2024-11-14
put version logic in one struct

phiSgr:


apps/framework-cli/src/framework/versions.rs
```diff
@@ -1,3 +1,94 @@
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::borrow::Cow;
+use std::cmp::Ordering;
+use std::fmt::{Debug, Display, Formatter};
+use std::hash::{Hash, Hasher};
+use std::sync::{Arc, OnceLock};
+
+#[derive(Clone)]
+pub struct Version<'a> {
```
callicles: Why does this needs to be this complicated?

ie why not?

ie
```rust
pub struct Version {
  version: String,
  parsed: Vec<i32>
}
```
with the values set at struct build time and no mutation?

callicles (commented):
I like the idea of having Version as a struct.

Not sure about the current implementation, it seems a bit complicated and could be done more simply?

apps/framework-cli/src/framework/versions.rs
phiSgr: `Cow<'a, str>`: For some reason I thought we'll construct a `Version` object with a `&str`, turns out we don't have that in our code. We might find such a case when we change more version strings to this struct. But maybe not.
`Arc<OnceLock<Vec<i32>>>`: I want to save the cost of parsing the string every time, so sth lazy and can be cloned cheaply.

General vibe I feel we're erring on the side of owning stuff too much in our rust code.

callicles: Hum yeah but I would say a Version should own its underlying data.
The thing using the Version could use a pointer to the Version

phiSgr: Yeah doing away the `Cow` makes sense.
but the `OnceLock` should be useful, no?

callicles: That feels like premature optimization. We are not doing high loads of generating Versions objects are we?

phiSgr: Fair enough🤔

callicles (approved):

PR #1880, opened 2024-11-19
chore: instrument spans for time spent in consumption api

callicles:
* Add instrumentation to get trace data from the consumption api

### DIFF
--- a/packages/ts-moose-lib/src/consumption-apis/runner.ts
+++ b/packages/ts-moose-lib/src/consumption-apis/runner.ts
@@ -3,7 +3,6 @@ import process from "node:process";
 import { getClickhouseClient } from "../commons";
 import { MooseClient, sql } from "./helpers";
 import * as jose from "jose";
-import * as Sentry from "@sentry/node";
 
 export const antiCachePath = (path: string) =>
   `${path}?num=${Math.random().toString()}&time=${Date.now()}`;
--- a/packages/ts-moose-lib/src/consumption-apis/runner.ts
+++ b/packages/ts-moose-lib/src/consumption-apis/runner.ts
@@ -3,6 +3,7 @@ import process from "node:process";
 import { getClickhouseClient } from "../commons";
 import { MooseClient, sql } from "./helpers";
 import * as jose from "jose";
+import * as Sentry from "@sentry/node";
 
 export const antiCachePath = (path: string) =>
   `${path}?num=${Math.random().toString()}&time=${Date.now()}`;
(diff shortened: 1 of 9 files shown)


packages/ts-moose-lib/src/instrumentation.ts
```diff
@@ -0,0 +1,26 @@
...
+    dsn: SENTRY_DSN,
+    environment: MOOSE_ENVIRONMENT,
+    integrations: [
+      // Add our Profiling integration
+      nodeProfilingIntegration(),
+    ],
+
+    // Add Tracing by setting tracesSampleRate
+    // We recommend adjusting this value in production
+    tracesSampleRate: 1.0,
```
cjus: Should these sample rate values be environment variables?

cjus (approved):
Changes make sense. Left one minor comment.

tg339 (approved):

packages/ts-moose-lib/src/consumption-apis/runner.ts
```diff
@@ -3,6 +3,7 @@ import process from "node:process";
 import { getClickhouseClient } from "../commons";
 import { MooseClient, sql } from "./helpers";
 import * as jose from "jose";
+import * as Sentry from "@sentry/node";
```
phiSgr: unused variable?

packages/ts-moose-lib/src/moose-runner.ts
```diff
@@ -10,6 +10,8 @@ register({
   experimentalTsImportSpecifiers: true,
 });
 
+import "./instrumentation";
```
phiSgr, line 13: i think it looks nicer if we call a function from the file instead of importing for side effects

packages/ts-moose-lib/src/instrumentation.ts
callicles: Probably

callicles: Adding it now

packages/ts-moose-lib/src/moose-runner.ts
callicles: That actually doesn't work - I just tried

packages/ts-moose-lib/src/consumption-apis/runner.ts
callicles: Removed

packages/ts-moose-lib/src/instrumentation.ts
callicles: Added

PR #1887, opened 2024-11-19
Nicolas/default load test

callicles:
* Looks like a bunch of time is currently spent in sentry
* Making the sampling smaller diminishes the footprint


cjus (approved):
Looks good

PR #1888, opened 2024-11-20
fix: add consumption client reuse

callicles:
Change the client to be reusable so that we reuse connections to the consumption web server


cjus (approved):
Nice reuse of the request client by passing it into other functions.  So reqwest::Client manages its own connection pool - can that be set externally without a recompile?

PR #1892, opened 2024-11-20
Implementation of a connection pool for clickhouse usage. 

cjus:
This reduces the overhead of managing a new connection to clickhouse per request.

### DIFF
--- a/packages/ts-moose-lib/src/consumption-apis/helpers.ts
+++ b/packages/ts-moose-lib/src/consumption-apis/helpers.ts
@@ -8,7 +8,7 @@
  * @returns 'FLoat', 'Int', 'Bool', 'String'
  */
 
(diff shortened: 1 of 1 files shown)


packages/ts-moose-lib/src/consumption-apis/helpers.ts
```diff
@@ -127,44 +127,80 @@ function emptyIfUndefined(value: string | undefined): string {
...
-    return this.client.query({
-      query,
-      query_params,
-      format: "JSONEachRow",
-      query_id: this.query_id_prefix + randomUUID(),
-    });
+      // We are not using the result of the ping
+      // but this ensures that if the clickhouse cloud service is idle, we
+      // wake it up.
+      client.ping();
```
callicles, line 189: We actually probably should remove this - this is extra that is using capacity on the node side

PR #1893, opened 2024-11-20
Nicolas/consumption server multi process

callicles:
* Adds parallelism to the node consumption API server

### DIFF
--- a/packages/ts-moose-lib/src/consumption-apis/runner.ts
+++ b/packages/ts-moose-lib/src/consumption-apis/runner.ts
@@ -1,9 +1,11 @@
 import http from "http";
 import process from "node:process";
+import cluster from "node:cluster";
 import { getClickhouseClient } from "../commons";
 import { MooseClient, sql } from "./helpers";
 import * as jose from "jose";
 import { ClickHouseClient } from "@clickhouse/client-web";
+import { Cluster } from "../cluster-utils";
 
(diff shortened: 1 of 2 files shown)


packages/ts-moose-lib/src/consumption-apis/runner.ts
```diff
@@ -152,26 +154,28 @@ const apiHandler =
...
+      const clickhouseClient = getClickhouseClient(clickhouseConfig);
+      let publicKey: jose.KeyLike | undefined;
+      if (JWT_SECRET) {
+        console.log("Importing JWT public key...");
+        publicKey = await jose.importSPKI(JWT_SECRET, "RS256");
+      }
 
-  const server = http.createServer(apiHandler(publicKey, clickhouseClient));
+      const server = http.createServer(apiHandler(publicKey, clickhouseClient));
+      server.listen(4001, () => {
```
phiSgr, line 167: multiple processes listening to same port? can it work🧐

cjus (approved):
This looks clean.

packages/ts-moose-lib/src/consumption-apis/runner.ts
callicles: Yeah that's the idea behind node cluster

https://nodejs.org/api/cluster.html

(Also I tested it)

PR #1895, opened 2024-11-22
put blocks behind leadership lock

DatGuyJonathan:


apps/framework-cli/src/framework/core/execute.rs
```diff
@@ -88,12 +90,15 @@ pub async fn execute_initial_infra_change(
         .await
         .map_err(ExecutionError::LeadershipCheckFailed)?
     {
-        // Execute migration changes only if we have the leadership lock
+        log::info!("Executing changes for leader instance");
+
+        processes::execute_olap_changes(&mut process_registries, &changes).await?;
```
callicles: There is a line above that says `olap::execute_changes` what's the difference with this?

callicles: Reading the other that changes that makes sense now - however would be great for this function to stand on its own. Could you add some comments or rename functions to make sure it is clear here the difference between the 2?

Thank you

DatGuyJonathan: Okay, I made some changes. Lmk if there's anything you had in mind!

callicles (approved):

PR #1896, opened 2024-11-22
Atomic pairing for SET NX and EXPIRE on resource locks

cjus:
Both commands (SET NX and EXPIRE) are executed in a single atomic transaction. There's no possibility of another process interfering between the two operations

### DIFF
--- a/apps/framework-cli/src/cli/routines.rs
+++ b/apps/framework-cli/src/cli/routines.rs
@@ -297,9 +297,6 @@ pub async fn setup_redis_client(project: Arc<Project>) -> anyhow::Result<Arc<Mut
         .register_lock("leadership", 10)
         .await?;
 
-    // Start the leadership lock management task
-    start_leadership_lock_task(redis_client.clone(), project.clone());
-
     let redis_client_clone = redis_client.clone();
     let callback = Arc::new(move |message: String| {
         let redis_client = redis_client_clone.clone();
@@ -309,6 +306,9 @@ pub async fn setup_redis_client(project: Arc<Project>) -> anyhow::Result<Arc<Mut
             }
         });
     });
+
+    // Start the leadership lock management task
+    start_leadership_lock_task(redis_client.clone(), project.clone());
 
     redis_client
         .lock()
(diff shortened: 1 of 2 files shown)


callicles (approved):

apps/framework-cli/src/infrastructure/redis/redis_client.rs
```diff
@@ -164,32 +164,23 @@ impl RedisClient {
...
     pub async fn attempt_lock(&self, name: &str) -> Result<bool> {
+        info!("<RedisClient> Attempting to lock {}", name);
         if let Some(lock) = self.locks.get(name) {
-            let result: bool = self
-                .connection
-                .lock()
-                .await
+            let result = redis::pipe()
+                .atomic()
                 .set_nx(&lock.key, &self.instance_id)
```
phiSgr, line 179: what about using `SetOptions` with `ExistenceCheck::NX` and `SetExpiry::EX(60)`?

apps/framework-cli/src/cli/routines.rs
```diff
@@ -364,7 +364,11 @@ fn start_leadership_lock_task(redis_client: Arc<Mutex<RedisClient>>, project: Ar
         loop {
             interval.tick().await;
             if let Err(e) = manage_leadership_lock(&redis_client, &project).await {
-                error!("Error managing leadership lock: {}", e);
+                if !e.to_string().contains("Failed to attempt lock") {
```
DatGuyJonathan, line 367: We might want to use a const for this so whoever throws this error uses the same message. If one side changes, we'll know about the other side.

apps/framework-cli/src/infrastructure/redis/redis_client.rs
cjus: didn't try or test that approach. We can certainly do that in a new test PR

apps/framework-cli/src/cli/routines.rs
cjus: Good call, we can update in a new PR.

PR #1900, opened 2024-11-26
wait for in-flight reqs to finish

DatGuyJonathan:
1. New pod starts up. Default RollingUpdate strategy allows it to co-exist with old pod
2. Health check for new pod succeeds
3. Traffic is routed to new pod
4. No new traffic is routed to old pod. Old pod gets a sigterm. 

Default `terminationGracePeriodSeconds` is 30 seconds. So we wait for in-flight reqs to finish instead of killing the server immediately.

### DIFF
--- a/apps/framework-cli/src/cli/local_webserver.rs
+++ b/apps/framework-cli/src/cli/local_webserver.rs
@@ -999,15 +999,23 @@ impl Webserver {
                         with_spinner("Stopping containers", || {
                              let _ = docker::stop_containers(&project);
                         }, true);
+                    } else {
+                        log::info!("Received sigint, waiting for in-flight requests to complete");
+                        tokio::time::sleep(Duration::from_secs(30)).await;
                     }
+
                     std::process::exit(0);
                 }
                 _ = sigterm.recv() => {
                     if !project.is_production {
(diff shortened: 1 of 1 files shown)


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -999,15 +999,23 @@ impl Webserver {
                         with_spinner("Stopping containers", || {
                              let _ = docker::stop_containers(&project);
                         }, true);
+                    } else {
+                        log::info!("Received sigint, waiting for in-flight requests to complete");
+                        tokio::time::sleep(Duration::from_secs(30)).await;
```
callicles, line 1004: Should we make this 20s instead so that it doesn't some up on the deadline for k8s?

callicles: Also would this affect the stop sown sequence in dev mode?

The guy in the talk had a sleep somewhere else?

I believe it was in the k8s manifest itself

phiSgr: i wonder if it's possible to not hard code this value

callicles:
We can close this, right?

PR #1901, opened 2024-12-01
Nicolas/parallelize streaming consumption

callicles:
* Adds at least once delivery to TS streaming functions
* Adds a planning validator, ie once a plan is done, different components can inspect the changes that are going to be applied and reject them if they are not compliant with the piece of infrastructure. example partitions in Kafka / Redpanda can't go down, only up
* Makes sure we display the new values for Topic edition
* Deleted Corev1 code that I could find, this was getting in the way of some changes I had to do, so removed the code instead of fixing it.
* Added parallelization for streaming functions and topics, (ie more than 1 partition in redpanda)

In a data model config one can now add:

```ts
export const UserActivityConfig: DataModelConfig<UserActivity> = {
  parallelism: 2,
};
```

to control parallelism, I would like to make that automatic maybe in the future, but for now it is manual


apps/framework-cli/src/framework/controller.rs
```diff
@@ -9,27 +9,24 @@ use crate::infrastructure::olap::clickhouse::config::ClickHouseConfig;
...
 use crate::infrastructure::stream::redpanda::{
     send_with_back_pressure, wait_for_delivery, RedpandaConfig,
 };
-use crate::project::Project;
 use crate::proto::infrastructure_map::InitialDataLoad as ProtoInitialDataLoad;
 use clickhouse_rs::errors::codes::UNKNOWN_TABLE;
 use clickhouse_rs::ClientHandle;
 use futures::StreamExt;
-use log::{debug, error, info, warn};
+use log::error;
```
cjus: Why is this line seperate from the following line?  Why not combined?

packages/ts-moose-lib/src/cluster-utils.ts
```diff
@@ -16,55 +19,113 @@ export class Cluster<C> {
...
   constructor(options: {
-    workerStart: () => Promise<C>;
+    workerStart: (w: Worker, paralelism: number) => Promise<C>;
     workerStop: (c: C) => Promise<void>;
+    maxCpuUsageRatio?: number;
+    maxWorkerCount?: number;
   }) {
     this.workerStart = options.workerStart;
     this.workerStop = options.workerStop;
+    if (
```
cjus, line 57: Nice!

packages/ts-moose-lib/src/cluster-utils.ts
```diff
@@ -73,7 +134,11 @@ export class Cluster<C> {
       console.info(
         `worker ${worker.process.pid} exited with code ${code} and signal ${signal}`,
       );
-      // Track if any workers exit with non-zero code during shutdown
+
+      if (!this.shutdownInProgress) {
+        setTimeout(() => cluster.fork(), 10000);
```
cjus: How was the 10 seconds derived?  Might be better as a const with comment.

cjus (approved):
Looks good. Some minor questions if you have time.

PR #1903, opened 2024-12-02
fix: add some buffer to the computation for message max size

callicles:
Adds some buffer for the consumption of messages and batching them to kafka

### DIFF
--- a/packages/ts-moose-lib/src/streaming-functions/runner.ts
+++ b/packages/ts-moose-lib/src/streaming-functions/runner.ts
@@ -18,6 +18,10 @@ const HEARTBEAT_INTERVAL_CONSUMER = 3000;
 const RETRY_FACTOR_PRODUCER = 0.2;
 const DEFAULT_MAX_STREAMING_CONCURRENCY = 100;
 const RETRY_INITIAL_TIME_MS = 100;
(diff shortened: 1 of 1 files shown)


packages/ts-moose-lib/src/streaming-functions/runner.ts
```diff
@@ -326,14 +332,14 @@ const sendMessages = async (
 
         // Start a new chunk
         chunks = [message];
-        chunkSize = messageSize;
+        chunkSize = messageSize + KAFKAJS_BYTE_MESSAGE_OVERHEAD;
```
phiSgr: `messageSize` already includes the overhead. we're adding the overhead twice?

packages/ts-moose-lib/src/streaming-functions/runner.ts
```diff
@@ -315,7 +319,9 @@ const sendMessages = async (
     let chunkSize = 0;
 
     for (const message of messages) {
-      const messageSize = Buffer.byteLength(message.value, "utf8");
+      const messageSize =
+        Buffer.byteLength(message.value, "utf8") +
```
phiSgr, line 323: maybe a comment saying `message.key` is not set, so that does not take up any length.

phiSgr (approved):

PR #1915, opened 2024-12-06
More specific feedback message (For llm)

georgevanderson:


phiSgr (approved):
there's an extra space in the diff

PR #1923, opened 2024-12-10
load python query params

phiSgr:


apps/framework-cli/src/framework/consumption/loader.rs
```diff
@@ -1,50 +1,90 @@
...
-pub fn load_consumption(project: &Project) -> Result<Consumption, ConsumptionLoaderError> {
-    let endpoint_files = walkdir::WalkDir::new(project.consumption_dir())
+#[derive(Debug, Deserialize)]
+struct QueryParamOutput {
+    params: Vec<ConsumptionQueryParam>,
+}
+
+pub async fn load_consumption(project: &Project) -> Result<Consumption, ConsumptionLoaderError> {
+    let mut endpoint_files = Vec::new();
+    for f in walkdir::WalkDir::new(project.consumption_dir())
```
callicles, line 26: Nice - are we compatible with nested paths now on the consumption side?

ie if I put something in `apis/test/nested` will the URL reflect that?

apps/framework-cli/src/framework/consumption/loader.rs
```diff
@@ -1,50 +1,90 @@
...
-        })
-        .collect();
+        }
+    }
 
     Ok(Consumption { endpoint_files })
 }
 
-fn build_endpoint_file(
+async fn load_python_query_param(
```
callicles: Can we not have `python` specific code that is not in the python module?
I would like to keep the separation clean

apps/framework-cli/src/framework/consumption/loader.rs
phiSgr: maybe that was the intent when we implemented it
but i don't think that works in python now because i don't see code that maps the `/` (in paths) to `.` (in python modules)

callicles (approved):

PR #1932, opened 2024-12-16
Update leadership lock handling

cjus:
- Extend leadership lock interval
- If same instance of moose was the leader and lost leaderhip then if it renewal leadership it should not register cron jobs.

### DIFF
--- a/apps/framework-cli/src/cli/routines.rs
+++ b/apps/framework-cli/src/cli/routines.rs
@@ -373,11 +373,14 @@ async fn manage_leadership_lock(
     if has_lock && is_new_acquisition {
         info!("<RedisClient> Obtained leadership lock, performing leadership tasks");
 
+        IS_RUNNING_LEADERSHIP_TASKS.store(true, Ordering::SeqCst);
+
         let project_clone = project.clone();
         let cron_registry: CronRegistry = cron_registry.clone();
         tokio::spawn(async move {
             if let Err(e) = leadership_tasks(project_clone, cron_registry).await {
                 error!("<RedisClient> Error executing leadership tasks: {}", e);
+                IS_RUNNING_LEADERSHIP_TASKS.store(false, Ordering::SeqCst);
             }
         });
 
--- a/apps/framework-cli/src/cli/routines.rs
+++ b/apps/framework-cli/src/cli/routines.rs
@@ -492,14 +492,6 @@ pub async fn start_development_mode(
         .start(route_table, consumption_apis, infra_map, project, metrics)
         .await;
 
-    // Only stop CronRegistry if we're the leader
-    let has_lock = redis_client.lock().await.has_lock("leadership").await?;
-    if has_lock {
-        if let Err(e) = CronRegistry::new().await?.stop().await {
-            error!("Failed to stop CronRegistry: {}", e);
-        }
-    }
-
     {
         let mut redis_client = redis_client.lock().await;
         let _ = redis_client.stop_periodic_tasks();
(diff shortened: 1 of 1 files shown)


apps/framework-cli/src/cli/routines.rs
```diff
@@ -362,24 +366,32 @@ async fn manage_leadership_lock(
...
-        let mut client = redis_client.lock().await;
-        if let Err(e) = client.broadcast_message("leader.new").await {
-            error!(
-                "<RedisClient> Failed to broadcast new leader message: {}",
-                e
-            );
         }
+    } else {
+        // If we lost the lock, reset the leadership tasks flag
+        IS_RUNNING_LEADERSHIP_TASKS.store(false, Ordering::SeqCst);
```
phiSgr, line 400: i think `false` should be set after the spawned tasks inside `leadership_tasks` are dead

apps/framework-cli/src/cli/routines.rs
```diff
@@ -362,24 +366,32 @@ async fn manage_leadership_lock(
...
-
-        let project_clone = project.clone();
-        tokio::spawn(async move {
-            if let Err(e) = leadership_tasks(project_clone).await {
-                error!("<RedisClient> Error executing leadership tasks: {}", e);
+    if has_lock {
+        if is_new_acquisition && !IS_RUNNING_LEADERSHIP_TASKS.load(Ordering::SeqCst) {
+            info!("<RedisClient> Obtained leadership lock, performing leadership tasks");
+
+            IS_RUNNING_LEADERSHIP_TASKS.store(true, Ordering::SeqCst);
```
callicles: Do you set this back to false when the leadership task is lost?

apps/framework-cli/src/cli/routines.rs
cjus: Yes nice catch updating PR

apps/framework-cli/src/cli/routines.rs
cjus: Yes updating PR

phiSgr (approved):
lgtm

apps/framework-cli/src/cli/routines.rs
```diff
@@ -384,6 +388,13 @@ async fn manage_leadership_lock(
                 e
             );
         }
+    } else if IS_RUNNING_LEADERSHIP_TASKS.load(Ordering::SeqCst) {
+        // Stop the CronRegistry first
+        if let Err(e) = CronRegistry::new().await?.stop().await {
```
phiSgr: use the `cron_registry` passed in

phiSgr (dismissed):

apps/framework-cli/src/cli/routines.rs
cjus: updated

apps/framework-cli/src/cli/routines.rs
```diff
@@ -384,6 +388,12 @@ async fn manage_leadership_lock(
                 e
             );
         }
+    } else if IS_RUNNING_LEADERSHIP_TASKS.load(Ordering::SeqCst) {
```
HDIBen, line 395: Where / how does IS_RUNNING_LEADERSHIP_TASKS becomes true? It's initialized as false on line 132 and is set to false on line 396 but I don't see where it's going to be true. Thanks for explaining.

cjus: This is correct and is an issue.  Updating PR now.

apps/framework-cli/src/cli/routines.rs
```diff
@@ -373,11 +373,14 @@ async fn manage_leadership_lock(
...
         info!("<RedisClient> Obtained leadership lock, performing leadership tasks");
 
+        IS_RUNNING_LEADERSHIP_TASKS.store(true, Ordering::SeqCst);
+
         let project_clone = project.clone();
         let cron_registry: CronRegistry = cron_registry.clone();
         tokio::spawn(async move {
             if let Err(e) = leadership_tasks(project_clone, cron_registry).await {
                 error!("<RedisClient> Error executing leadership tasks: {}", e);
+                IS_RUNNING_LEADERSHIP_TASKS.store(false, Ordering::SeqCst);
```
callicles: Feels like this should be outside of the `if`

callicles (commented):
Maybe last comment?

apps/framework-cli/src/cli/routines.rs
cjus: yes fixed

callicles (approved):

PR #1933, opened 2024-12-16
reuse the same cron registry

phiSgr:


callicles (approved):

apps/framework-cli/src/cli/routines.rs
```diff
@@ -346,7 +346,8 @@ fn start_leadership_lock_task(redis_client: Arc<Mutex<RedisClient>>, project: Ar
         let mut interval = interval(Duration::from_secs(LEADERSHIP_LOCK_RENEWAL_INTERVAL)); // Adjust the interval as needed
         loop {
             interval.tick().await;
-            if let Err(e) = manage_leadership_lock(&redis_client, &project).await {
+            let cron_registry = CronRegistry::new().await.unwrap();
```
cjus: Doesn't this create a new CronRegistry on every pass of the loop?

phiSgr: omg

apps/framework-cli/src/cli/routines.rs
```diff
@@ -344,9 +344,12 @@ async fn process_pubsub_message(
 fn start_leadership_lock_task(redis_client: Arc<Mutex<RedisClient>>, project: Arc<Project>) {
     tokio::spawn(async move {
         let mut interval = interval(Duration::from_secs(LEADERSHIP_LOCK_RENEWAL_INTERVAL)); // Adjust the interval as needed
+
+        let cron_registry = CronRegistry::new().await.unwrap();
```
cjus, line 348: cool this fixes it.

cjus (approved):

apps/framework-cli/src/cli/routines.rs
phiSgr: it's amusing how after some changes i managed to make the same silly mistake lol

PR #1987, opened 2025-01-07
fix blocks in filewatcher codepath

DatGuyJonathan:


apps/framework-cli/src/framework/core/execute.rs
```diff
@@ -122,6 +122,7 @@ pub async fn execute_online_change(
         .await
         .map_err(Box::new)?;
 
+    processes::execute_leader_changes(process_registries, &plan.changes.processes_changes).await?;
```
callicles, line 125: Can you explain this change?
Not sure I understand why this solves the issue

DatGuyJonathan: Recently, we put blocks behind leadership lock. That change meant `processes::execute_changes` was only updating stuff that can happen on any instance, and there's now `processes:: execute_leader_changes` that gets called in a leadership instance. That change also caused this bug because the file watcher codepath was only calling `processes::execute_changes`.

callicles: kk makes sense

callicles (approved):

PR #1993, opened 2025-01-15
openapi ingress (ts, py) & egress (py)

DatGuyJonathan:
Generates openapi spec when starting local webserver, & refreshes it during file watcher mode.

https://github.com/user-attachments/assets/5401e359-ed67-4a38-b77e-c6e007ccdbd2

TBD:
- egress for typescript. dependent on the typia integration

### DIFF
--- a/apps/framework-cli/src/cli/routines.rs
+++ b/apps/framework-cli/src/cli/routines.rs
@@ -89,11 +89,11 @@ use std::sync::Arc;
 use tokio::sync::{Mutex, RwLock};
 use tokio::time::{interval, Duration};
 
-use crate::framework::core::execute::execute_initial_infra_change;
-use crate::framework::core::plan::plan_changes;
-
+use crate::cli::routines::openapi::openapi;
 use crate::framework::controller::RouteMeta;
+use crate::framework::core::execute::execute_initial_infra_change;
 use crate::framework::core::infrastructure_map::InfrastructureMap;
+use crate::framework::core::plan::plan_changes;
 use crate::infrastructure::olap::clickhouse_alt_client::{get_pool, store_infrastructure_map};
 use crate::infrastructure::processes::cron_registry::CronRegistry;
 use crate::infrastructure::processes::kafka_clickhouse_sync::clickhouse_writing_pause_button;
@@ -117,6 +117,7 @@ pub mod logs;
 pub mod ls;
 pub mod metrics_console;
 pub mod migrate;
+pub mod openapi;
 pub mod peek;
 pub mod ps;
 pub mod streaming;
@@ -472,6 +473,8 @@ pub async fn start_development_mode(
         &plan.target_infra_map,
     )
     .await?;
+
+    openapi(&project).await?;
 
     plan.target_infra_map
         .store_in_redis(&*redis_client.lock().await)
--- a/apps/framework-cli/src/cli/routines.rs
+++ b/apps/framework-cli/src/cli/routines.rs
@@ -474,7 +474,7 @@ pub async fn start_development_mode(
     )
     .await?;
 
-    openapi(&project).await?;
+    let openapi_file = openapi(&project).await?;
 
     plan.target_infra_map
         .store_in_redis(&*redis_client.lock().await)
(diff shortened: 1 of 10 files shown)


tg339:
FUCK. YEAH

callicles:
Nicely done!

On the next steps, do we want to expose the full swagger or just have an HTTP API endpoint that exposes the yaml (maybe on the management webserver) so that LLM can read the yaml there and users could point a reader at that?

apps/framework-cli/src/cli/routines/openapi.rs
```diff
@@ -0,0 +1,231 @@
...
+    Save(String),
+}
+
+pub async fn openapi(project: &Arc<Project>) -> Result<(), OpenAPIError> {
+    let mut client = get_pool(&project.clickhouse_config)
+        .get_handle()
+        .await
+        .map_err(|_| OpenAPIError::Connection)?;
+
+    let infra = retrieve_infrastructure_map(&mut client, &project.clickhouse_config)
```
callicles, line 124: Do we want to build more on top of the clickhouse map or should we load it on top of the redis content?

apps/framework-cli/src/cli/routines/openapi.rs
```diff
@@ -0,0 +1,231 @@
...
+                            Some(serde_json::Value::String(Local::now().to_rfc3339())),
+                        ),
+                        ColumnType::Array(_) => (
+                            "array".to_string(),
+                            Some(serde_json::Value::Array(vec![serde_json::Value::String(
+                                "example".to_string(),
+                            )])),
+                        ),
+                        _ => (
+                            "string".to_string(),
```
phiSgr: Nested should be `object`.

apps/framework-cli/src/cli/routines/openapi.rs
```diff
@@ -0,0 +1,231 @@
...
+                            Some(serde_json::Value::Number(
+                                serde_json::Number::from_f64(1.0).unwrap(),
+                            )),
+                        ),
+                        ColumnType::DateTime => (
+                            "string".to_string(),
+                            Some(serde_json::Value::String(Local::now().to_rfc3339())),
+                        ),
+                        ColumnType::Array(_) => (
+                            "array".to_string(),
```
phiSgr: maybe add sth like `TODO: add items.type`

DatGuyJonathan:
> Nicely done!
> 
> On the next steps, do we want to expose the full swagger or just have an HTTP API endpoint that exposes the yaml (maybe on the management webserver) so that LLM can read the yaml there and users could point a reader at that?

Good call. I was looking at rust crates for swagger ui. There'd be an endpoint for the ui, but it would need to fetch the yaml from another endpoint, so I added a route to management router since that seems to be a good stepping stone to however the spec ends up being used.

apps/framework-cli/src/cli/routines/openapi.rs
DatGuyJonathan: I may have missed the convo on why infra map is in both ch & redis. Could you or @phiSgr give me some context?

phiSgr: the infra map stored on redis was added to allow boreal to read it.

DatGuyJonathan: Ah okay. I don't see `infrastructure_map` in boreal yet, is that feature still on the docket? @okane16 , would you happen to know?

I don't have a preference where we get the infra map from for local dev. If it's redis, I think there's some additional work to deserialize the protobuf.

callicles: @phiSgr don't we have the client already setup?

DatGuyJonathan:
@callicles @phiSgr @cjus , I updated the PR with egress for python. Check out the vid!

apps/framework-cli/src/cli/routines.rs
```diff
@@ -579,7 +589,14 @@ pub async fn start_production_mode(
...
 
     web_server
-        .start(route_table, consumption_apis, infra_map, project, metrics)
+        .start(
+            route_table,
+            consumption_apis,
+            infra_map,
+            project,
+            metrics,
+            None,
```
phiSgr, line 598: Why don't we provide the API doc in prod?

callicles: if this is on the public webserver, we don't want that to be public do we?

callicles (approved):
Let's merge - we can iterate on redis vs clickhouse later

PR #2001, opened 2025-01-24
Typia in consumption

phiSgr:


callicles (commented):
LGTM,

is that reverse compatible for existing customer or is there going to be a migration of code + dependencies necessary?

apps/framework-cli/src/project/typescript_project.rs
```diff
@@ -35,7 +35,11 @@ impl Default for TypescriptProject {
                 ("moose".to_string(), "moose-cli".to_string()),
                 ("build".to_string(), "moose-cli build --docker".to_string()),
             ]),
-            dependencies: HashMap::from([("@514labs/moose-lib".to_string(), "latest".to_string())]),
+            dependencies: HashMap::from([
+                ("@514labs/moose-lib".to_string(), "latest".to_string()),
+                ("ts-patch".to_string(), "^3.3.0".to_string()),
```
callicles, line 40: That will require upgrade instructions for customers

callicles (approved):

PR #2004, opened 2025-01-28
scripts logger

DatGuyJonathan:
Screenshot 2025-01-28 at 2 57 43 PM

Screenshot 2025-01-28 at 2 58 25 PM


apps/framework-cli/src/framework/python/scripts_worker.rs
```diff
@@ -74,14 +74,13 @@ pub async fn start_worker(project: &Project) -> Result<Child, WorkerProcessError
         while let Ok(Some(line)) = stdout_reader.next_line().await {
             let parts: Vec<&str> = line.split('|').collect();
             if parts.len() >= 3 {
-                let level = parts[0];
-                let name = parts[1];
```
DatGuyJonathan, line 78: The name was always 'python_worker_wrapper.logging' - not sure how useful that is

callicles: Good call

callicles (approved):
Nicely done!

PR #2007, opened 2025-01-29
trigger workflow from api

DatGuyJonathan:
Screenshot 2025-01-29 at 1 54 26 PM

### DIFF
--- a/apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
+++ b/apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
@@ -2,6 +2,7 @@ import argparse
 import dataclasses
 import json
 import os
+import subprocess
 import sys
 import traceback
 from datetime import datetime, timezone, date
@@ -100,11 +101,11 @@ class EnhancedJSONEncoder(json.JSONEncoder):
         return super().default(o)
 
 
-class MooseClient:
+class QueryClient:
     def __init__(self, ch_client: Client):
         self.ch_client = ch_client
 
-    def query(self, input, variables):
+    def execute(self, input, variables):
         params = {}
         values = {}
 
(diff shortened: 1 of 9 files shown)


apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
```diff
@@ -122,16 +123,64 @@ def query(self, input, variables):
...
-        # wake it up, before we send the query.
-        self.ch_client.ping()
 
         val = self.ch_client.query(clickhouse_query, values)
 
         return list(val.named_results())
 
+class WorkflowClient:
+    def execute(self, name: str, input_data: Any) -> None:
+        moose_cli = "/usr/local/bin/moose-cli"
```
DatGuyJonathan: Not quite sure what should be. It's called `moose` in prod container, and `moose-cli` locally 🤔

callicles: or should it be a network call instead using the temporal io lib?

DatGuyJonathan: My thinking was it's easier to maintain if we just have one instance of that code (in rust), rather than three (rust, python, soon-to-be-typescript).

callicles: 🤔 I was asking myself the same thing. It is kinda what do we want to standardize on, the code the send a request to the temporal server through their lib seem pretty simple. 

Also not having a dependency on the cli means we can use serverless envs eventually to deploy the backend. 
And I would bet the temporal API is going to be more stable than our cli api

DatGuyJonathan: I tried the network call approach. It worked and doesn't look too bad. I made the temporal client initialized once when the http server comes up.

callicles (approved):

PR #2008, opened 2025-01-30
Doc Tweaks

georgevanderson:


apps/framework-docs/src/components/cta-card.tsx
```diff
@@ -31,29 +33,19 @@ export function CTACard({
...
-        >
-          <div
-            className={cn(
-              "rounded-[18px] w-fit p-[2px]",
-              variant === "gradient" ? "bg-gradientDarkPink" : "bg-muted",
-            )}
-          >
-            <Icon className="m-3 h-[24px] w-[24px] text-white" />
-          </div>
+        <div className="flex gap-2">
```
tg339, line 44: I think you need an item center here to align

tg339 (approved):

apps/framework-docs/src/components/cta-card.tsx
tg339: nvm I'm nuts

tg339 (approved):

PR #2009, opened 2025-01-30
reduce time of lock holding by flush

phiSgr:


apps/framework-cli/src/infrastructure/processes/kafka_clickhouse_sync.rs
```diff
@@ -464,7 +464,7 @@ async fn sync_kafka_to_clickhouse(
             select! {
                 received = subscriber.recv() => match received {
                     Err(e) => {
-                        debug!("Error receiving message from {}: {}", source_topic_name, e);
+                        warn!("Error receiving message from {}: {}", source_topic_name, e);
```
callicles, line 467: Why is this a warning?

apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
```diff
@@ -120,6 +121,59 @@ impl<C: ClickHouseClientTrait + 'static> Inserter<C> {
         Ok(())
     }
 
+    async fn get_front_batch(&self) -> Option<Batch> {
```
callicles, line 123: I think this removes the at least once delivery no?

Can you explain why it doesn't?

The thought is that if another process claims the queue in parallel and that batch goes through then if the next batch gets committed then the intermediary batch won't get replayed.

apps/framework-cli/src/infrastructure/processes/kafka_clickhouse_sync.rs
phiSgr: My hunch is that the consumer did not resume after the `PollExceeded` because some internal state is messed up. Better to make sure this error get logged.

apps/framework-cli/src/infrastructure/olap/clickhouse/inserter.rs
phiSgr: i think there's only one `spawn`ed `task` doing the `flush`
if that insert failed, the batch gets returned to the queue, and consumed by that same flushing process

callicles: Feels like if we merge this we need to a test that shows we don't drop data since that's the current implementation just got

callicles: Could be a manual test on your laptop where you send a bunch of things kill the process, restart and see records not get dropped

callicles:
What should we do about this?

phiSgr:
I do think it's worth merging. I'll tweak the code to simulate failures to test.

callicles:
We removed the lock altogether

PR #2011, opened 2025-01-30
fix python project when scripts is disabled

DatGuyJonathan:


apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
```diff
@@ -278,10 +281,15 @@ def main():
     ch_client = get_client(interface=interface, host=host,
                            port=port, database=db, username=user, password=password)
 
-    print("Connecting to Temporal")
-    # Need to await on temporal client calls but this main function is sync
-    temporal_client = asyncio.run(get_temporal_client())
-
+    temporal_client = None
+    try:
+        # TODO: try to connect since it's still behind a feature flag
```
DatGuyJonathan, line 286: Could also pass the feature flag around, lmk!

callicles: Will everyone get the error message or will it bein the logs for moose hidden?

DatGuyJonathan: Hmm, the error message is in the logs like this, but not in the terminal where they run `moose-cli dev`, so I guess it's sort of hidden :stuck_out_tongue:

```
Failed to connect to Temporal. Is the feature flag enabled? Failed client connect: Server connection error: tonic::transport::Error(Transport, ConnectError(ConnectError("tcp connect error", Os { code: 61, kind: ConnectionRefused, message: "Connection refused" })))
```

callicles: Then I am fine with it

callicles (approved):

PR #2012, opened 2025-01-31
dump ts query param

phiSgr:


apps/framework-cli/src/framework/typescript/export_collectors.rs
```diff
@@ -21,21 +26,26 @@ pub enum ExportCollectorError {
...
     })?;
 
     let args = vec![file_path_str];
-    let process = bin::run(EXPORT_SERIALIZER_BIN, project_path, &args)?;
+    let process = bin::run(command_name, project_path, &args)?;
 
     let mut stdout = process
         .stdout
-        .expect("Data model config process did not have a handle to stdout");
+        .unwrap_or_else(|| panic!("{process_name} process did not have a handle to stdout"));
```
callicles, line 44: Not  a big deal since that feels like exactly the change, why the change?

callicles (approved):

apps/framework-cli/src/framework/core/infrastructure/api_endpoint.rs
```diff
@@ -29,6 +29,8 @@ pub enum APIType {
     },
     EGRESS {
         query_params: Vec<ConsumptionQueryParam>,
+        #[serde(default)]
```
DatGuyJonathan, line 32: Does this mean we don't have to update `backwards_compatible_deserialize`?

DatGuyJonathan (approved):
I strive to write code as clean as this  :raised_hands:

apps/framework-cli/src/framework/core/infrastructure/api_endpoint.rs
DatGuyJonathan: Never mind, I missed it down there

apps/framework-cli/src/framework/typescript/export_collectors.rs
phiSgr: Two process names, so the error message has to be dynamic. And clippy is not happy with `expect(format!(...))` (allocating a string even if no error)

PR #2015, opened 2025-02-03
add retry & timeout in workflow run cmd

DatGuyJonathan:
For example, a config like this:

```
name = 'testretry'
schedule = ''
retries = 3
timeout = '1m'
steps = ['step1']
```

would give each workflow 1 minute for each of the 3 retries.

### DIFF
--- a/apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
+++ b/apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
@@ -6,7 +6,7 @@ import os
 import subprocess
 import sys
 import traceback
-from datetime import datetime, timezone, date
+from datetime import datetime, timezone, date, timedelta
 
 from http import HTTPStatus
 from http.server import HTTPServer, BaseHTTPRequestHandler
@@ -22,6 +22,7 @@ from clickhouse_connect import get_client
 from clickhouse_connect.driver.client import Client as ClickhouseClient
 
 from temporalio.client import Client as TemporalClient
+from temporalio.common import RetryPolicy
 
 
 parser = argparse.ArgumentParser(description='Run Consumption Server')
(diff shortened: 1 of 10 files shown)


apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
```diff
@@ -158,6 +158,7 @@ def execute(self, name: str, input_data: Any) -> Dict[str, Any]:
             }
     
     async def _start_workflow_async(self, name: str, input_data: Any):
+        # TODO: add retry & timeout
```
DatGuyJonathan: We'd need to parse the configs again in python consumption, or figure out how to pass the configs to consumption api, or just call `moose-cli workflow run`. I wanted to get opinions before I go into a rabbit hole.

callicles: Hum and there is the how to hot reload as well. 

I am thinking that points to needing to passing it in to be able to use the watcher in dev. Otherwise, we would have to do some watching logic here as well

cjus (approved):
LGTM - pending consideration of @callicles feedback

apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
DatGuyJonathan: I was leaning to that option as well, but the tricky thing is we don't know which workflow they'll call from which api. So we'd have to pass all the configs, and might run into string length limit when passing the configs to api. The workaround was to serialize the configs to a .moose file. Lmk what you think.

callicles: When I said passing it in I was thinking as a CLI argument when we start the consumption apis, since we trigger the restart of the consumption API server for all changes, we could hook changes to the script files to do the same think and pass in the config that way. It should not limit the length of what we pass in.

Then as I am writing this I am wondering if the context is the right one.

callicles (approved):

PR #2025, opened 2025-02-07
peek from kafka and better data type error message in python

phiSgr:


apps/framework-cli/src/cli/routines/peek.rs
```diff
@@ -43,47 +50,108 @@ pub async fn peek(
...
+                .stream()
+                .map(|message| {
+                    Ok(serde_json::from_slice::<Value>(
+                        message?.payload().unwrap_or(&[]),
+                    )?)
+                })
+                .take(limit.into()),
+        )
+    } else {
+        // ¯\_(ツ)_/¯
```
cjus, line 111: Why is this here?

cjus (approved):
Looks good. Just one little question...

apps/framework-cli/src/cli/routines/peek.rs
phiSgr: coz the double version suffix is ugly

PR #2031, opened 2025-02-12
Olivia/dm docs

okane16:
consolidated data modeling docs and made DCM hidden


apps/framework-docs/src/pages/data-models.mdx
```diff
@@ -74,29 +124,28 @@ from typing import List
...
-    activity: str
-
+class MyFirstDataModel:
+    primary_key: Key[str]
+    some_string: str
+    some_number: int
+    some_boolean: bool
+    some_date: datetime.datetime
+    some_array: List[str]
+    some_optional_string: Optional[str] # Optional fields are marked with a `?`
```
phiSgr: that's ts

apps/framework-docs/src/pages/data-models.mdx
```diff
@@ -1,67 +1,117 @@
...
 
-## Core Concepts
+@moose_data_model
+@dataclass
+class MyDataModel:
+    primary_key: Key[str]
+    some_string: str
+    some_number: int
+    some_boolean: bool
+    some_list: list[str]
```
phiSgr, line 41: this is not equivalent to the ts example - missing `datetime`

tg339 (approved):

PR #2039, opened 2025-02-13
ts workflow logger

DatGuyJonathan:


apps/framework-cli/src/cli/display.rs
```diff
@@ -234,6 +233,11 @@ pub fn show_table(headers: Vec<String>, rows: Vec<Vec<String>>) {
     );
 }
 
+pub fn show_message_wrapper(message_type: MessageType, message: Message) {
+    // Probably shouldn't do macro_export so we just wrap it
+    show_message!(message_type, message, false);
+}
```
phiSgr, line 239: maybe we should just change the macro to a regular function.

cjus (approved):
Looks reasonable.

phiSgr (approved):

apps/framework-cli/src/cli/display.rs
DatGuyJonathan: :stuck_out_tongue:

PR #2045, opened 2025-02-16
allows disabling the API endpoint & topic of a data model

phiSgr:


callicles (approved):

callicles:
We should also not spin up Redpanda in the compose file if it is not required

PR #2047, opened 2025-02-18
task retry ts workflow

DatGuyJonathan:
For example, this configures it to one retry. If the config is omitted, retries default to 3, like in python workflow.

```
export default function createTask(input?: object) {
    return {
        task: step1,
        config: {
            retries: 1,
        }
    } as TaskDefinition;
}
```

### DIFF
--- a/packages/ts-moose-lib/src/scripts/workflow.ts
+++ b/packages/ts-moose-lib/src/scripts/workflow.ts
@@ -2,11 +2,10 @@ import { log as logger, proxyActivities } from "@temporalio/workflow";
 import { WorkflowState } from "./types";
 import { mooseJsonEncode } from "./serialization";
 
-const { executeScript, readDirectory } = proxyActivities({
-  startToCloseTimeout: "10 minutes",
+const { getActivityRetry, readDirectory } = proxyActivities({
+  startToCloseTimeout: "1 minutes",
   retry: {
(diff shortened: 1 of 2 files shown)


packages/ts-moose-lib/src/scripts/workflow.ts
```diff
@@ -2,11 +2,10 @@ import { log as logger, proxyActivities } from "@temporalio/workflow";
...
 import { mooseJsonEncode } from "./serialization";
 
-const { executeScript, readDirectory } = proxyActivities({
-  startToCloseTimeout: "10 minutes",
+const { getActivityRetry, readDirectory } = proxyActivities({
+  startToCloseTimeout: "1 minutes",
   retry: {
-    // TODO: Use user's retry config
-    maximumAttempts: 3,
+    maximumAttempts: 1,
```
cjus, line 8: Why is this set to 1?

cjus (commented):
Looks good - just one question

packages/ts-moose-lib/src/scripts/workflow.ts
DatGuyJonathan: `const { getActivityRetry, readDirectory } = proxyActivities({`

These are our functions, not the user's task function.

cjus (approved):

PR #2052, opened 2025-02-19
Olivia/quickstart

okane16:


callicles (commented):
Why are we changing the quickstart with an example that is Foobar vs a real one?

okane16:
> Why are we changing the quickstart with an example that is Foobar vs a real one?

We've learned that its easier to grasp moose without anchoring it to a specific domain/real use case. and keeping things intentionally generic helps focus on moose capabilities instead

DatGuyJonathan (approved):

PR #2053, opened 2025-02-21
moose cleanup

DatGuyJonathan:
- Remove dcm related cli commands `bump-version` and `generate migrations`
- Remove unused dcm functions
- Remove leftover core_v1 code


apps/framework-cli/src/cli/commands.rs
```diff
@@ -58,12 +58,8 @@ pub enum Commands {
     Dev {},
     /// Start a remote environment for use in cloud deployments
     Prod {},
-    /// Generates missing migration files
+    /// Generates sdk for your data models
```
callicles: This is also used to generate the hash function - so probably not accurate helper

apps/framework-cli/src/framework/controller.rs
```diff
@@ -69,42 +61,6 @@ pub enum InitialDataLoadStatus {
...
-                config,
-                offset,
-            )
-            .await?;
-        }
-    };
-    Ok(())
-}
-
 pub async fn resume_initial_data_load(
```
callicles, line 75: Shouldn't this be deleted as well?

apps/framework-cli/src/infrastructure/processes/kafka_clickhouse_sync.rs
```diff
@@ -86,91 +81,6 @@ impl SyncingProcessesRegistry {
...
-    ) -> Result<(), rdkafka::error::KafkaError> {
-        info!("<DCM> Starting all syncing processes");
-
-        let kafka_config = self.kafka_config.clone();
-        let clickhouse_config = self.clickhouse_config.clone();
-
-        let available_topics: HashSet<String, RandomState> =
-            HashSet::from_iter(fetch_topics(&kafka_config).await?);
-
-        // Spawn sync for the current and old models
```
callicles, line 103: That feels like it should not get deleted

callicles (changes requested):
Looks like some stuff should not be deleted and some more could?

callicles:
or we could keep it - not sure

apps/framework-cli/src/cli/commands.rs
DatGuyJonathan: Gotcha, will update

apps/framework-cli/src/framework/controller.rs
DatGuyJonathan: There are some dcm functions that became part of core_v2. I believe this is still needed, so I just removed "dcm" from the log line.

apps/framework-cli/src/infrastructure/processes/kafka_clickhouse_sync.rs
DatGuyJonathan: The function isn't called by anything else, and I also checked prod logs for the absence of "Starting all syncing processes"

DatGuyJonathan:
> Also a callout for something that needs to be tested - with DCM we were using version with naming DMs with removing DCM we probably should remove the naming versioning?

I'm going to look into a one time script that can rename those tables. It will likely be another big pr that updates all the versioning, so I'm thinking we do this piecemeal and merge this in first.

apps/framework-cli/src/infrastructure/olap/clickhouse.rs
```diff
@@ -430,147 +377,12 @@ pub async fn fetch_tables_with_version(
...
-        .await?;
-
-    match result.len() {
-        // i keep getting 2 rows when I have this logic in the select query
-        1 => Ok(result[0].engine != "View" && result[0].total_rows == Some(0)),
-        _ => panic!("Expected 1 result, got {:?}", result),
-    }
-}
-
-/// Gets the number of rows in a table
```
callicles, line 548: Let's keep the docs for this method

apps/framework-cli/src/infrastructure/processes/kafka_clickhouse_sync.rs
callicles: Yeah - double checked as well - that was surprising to me but it looks we had dropped some DCM functionality already

apps/framework-cli/src/infrastructure/olap/clickhouse.rs
```diff
@@ -356,58 +355,6 @@ pub async fn check_ready(
...
-        .into_iter()
-        .map(|row| row.to_table())
-        .collect();
-
-    debug!("<DCM> Fetched tables: {:?}", tables);
-
-    Ok(tables)
-}
-
-/// Fetches tables matching a specific version pattern
```
callicles, line 398: This doc as well please 😄 

I think those are docs I added in other PRs that got merged

apps/framework-cli/src/infrastructure/olap/clickhouse.rs
DatGuyJonathan: Good catch! Rebase error, will fix

callicles (approved):
Just the docs to keep otherwise good to go.

Thank you for the work!

callicles:
>I'm going to look into a one time script that can rename those tables. It will likely be another big pr that updates all the versioning, so I'm thinking we do this piecemeal and merge this in first.

Not sure we would want to change the name of tables for deployments already made, maybe just use as is? and new names generated without versions?

I would prefer a rollout where we don't have to go through every deployment and fix it before / after we rollout if that's what you were suggesting with the one time script

phiSgr:
looks like we can remove some dependencies as well?

PR #2062, opened 2025-02-25
Olivia/docs cleanup

okane16:
- Feedback from quickstart issues
- Update apis docs to match boilerplate project
- removed alpha status & templates


apps/framework-docs/src/pages/_meta.json
```diff
@@ -35,7 +34,9 @@
     "title": "Reference"
   },
   "moose-cli": "Moose CLI",
-  "templates": "Templates",
+  "templates": {
```
DatGuyJonathan, line 37: Hmm, should this be removed? It's hidden here and the file itself is empty.

DatGuyJonathan (approved):

apps/framework-docs/src/pages/_meta.json
okane16: need to keep it for the hiddent to apply. theres a subfolder that gets picked up if i dont have the file

PR #2066, opened 2025-02-26
Improved redis client handling to managed panic potentials

cjus:
This pull request introduces several improvements to the Redis client setup and connection monitoring in the `apps/framework-cli/src/cli/routines.rs` file. The key changes include adding warning logs, handling connection monitor restarts more gracefully, and updating the connection state during restarts.

### Improvements to Redis client setup and monitoring:

* Added `warn` to the `log` imports to enable warning logs.
* Added a check to log a warning if the Redis client is not connected during setup, indicating that some features will be limited.

### Enhancements to connection monitor handling:

* Modified the `spawn_connection_monitor` function to handle the result directly without using `catch_unwind`, and added a warning log to indicate the monitor will be restarted after exiting gracefully.
* Updated the connection state to `false` during the monitor restart to ensure the connection state is accurately reflected.

### DIFF
--- a/apps/framework-cli/src/cli/routines.rs
+++ b/apps/framework-cli/src/cli/routines.rs
@@ -88,7 +88,7 @@
 
 use crate::framework::core::plan_validator;
 use crate::infrastructure::redis::redis_client::RedisClient;
-use log::{debug, error, info};
+use log::{debug, error, info, warn};
 use std::collections::{HashMap, HashSet};
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicBool, Ordering};
@@ -207,7 +207,11 @@ pub async fn setup_redis_client(project: Arc<Project>) -> anyhow::Result<Arc<Mut
     let redis_client = RedisClient::new(project.name(), project.redis_config.clone()).await?;
     let redis_client = Arc::new(Mutex::new(redis_client));
 
-    spawn_connection_monitor(redis_client.clone());
+    if redis_client.lock().await.is_connected() {
+        spawn_connection_monitor(redis_client.clone());
+    } else {
+        warn!("Redis connection is not available. Some features will be limited.");
+    }
 
     let (service_name, instance_id) = {
         let client = redis_client.lock().await;
@@ -703,10 +707,16 @@ fn spawn_connection_monitor(redis_client: Arc<Mutex<RedisClient>>) {
                 let client = redis_client.lock().await;
                 client.start_connection_monitor()
             };
-            match handle.await {
-                Ok(_) => {
+
+            // Handle the result directly without using catch_unwind
+            let result = handle.await;
+
+            match result {
+                Ok(()) => {
                     info!("Connection monitor exited gracefully");
-                    break;
+                    // Even if it exited gracefully, we should restart it
+                    warn!("Restarting connection monitor in 1 second...");
+                    sleep(Duration::from_secs(1)).await;
                 }
                 Err(err) => {
                     error!(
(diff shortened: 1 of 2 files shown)


callicles:
I am a bit confused by those changes, are you removing the connection manager from the lib?

cjus:
> I am a bit confused by those changes, are you removing the connection manager from the lib?

Not the connection manager is still used throughout the code.  But the connection manager alone is not enough to handle the monitor of Redis Connections.  This and the potential for unhandled panics is a known issue with the Redis Connection manager for Rust.

cjus:
Additional notes on this PR:

## Redis Reconnection

The reconnection logic is implemented primarily in two parts:

* The connection monitor task (spawned via start_connection_monitor) continuously checks the connection state and uses an exponential backoff strategy between reconnection attempts.
* The attempt_reconnection function is responsible for reestablishing both the normal Redis connection and the PubSub connection. It first checks the current connection state by issuing a PING (using a try_lock on the connection); if the connection appears unhealthy, it is marked as disconnected.
* In `attempt_reconnection`, before attempting to create new connections:
 * A simple ping (wrapped in catch_unwind and timeout) is used to verify whether the Redis server is available.
 * If available, the function attempts to create new connections with a 5-second timeout. Importantly, both the normal connection and the pub/sub connection are created using:
 
```rust 
  std::panic::AssertUnwindSafe(Self::create_connection_safely(client)).catch_unwind()
```
  
So that if the underlying redis-aio connection creation panics (e.g., in the MultiplexedConnection driver), the panic is caught. If a panic is caught, it logs the error, sets the connection state as false, and returns None.

Once both new connections have been successfully created, the function uses try_lock to acquire the locks and replace the old connection managers with the new ones. If the replacement is successful for both connections, the connection state is updated to “true.” Otherwise, errors are logged and the state remains false.

The reconnection mechanism uses robust error handling (timeouts, catch_unwind, try_lock) ***so that transient errors or panics from the underlying Redis library do not crash the process***. Also, the exponential backoff strategy helps to reduce log spam when Redis is unreachable.

## Mock Redis Client
This PR uses a mock Redis client as a safety fallback mechanism. Rather than failing immediately when a real Redis connection can't be established (for example, because the Redis server is temporarily unavailable or misconfigured), the application creates a mock client. This mock client lets the service continue running—albeit without a live Redis backend—allowing for things like reconnection attempts later on and preventing the entire service from crashing immediately. Essentially, it improves fault tolerance, especially during startup, testing, or in environments where Redis may not be critical for every operation.

Here’s how it works, both in local and cloud environments where an actual Redis server might not be available:

### Attempt to Create a Connection Manager:
The code first tries to create a connection manager by calling

```rust
ConnectionManager::new(client.clone()).await
```
where the client is created with the configured Redis URL.

### Fallback When Connection Fails:
If that attempt fails (for example, if there’s no Redis server available at the expected URL), the code logs a warning and then tries to create a “safer fallback”

### Create a Fallback Client:
It attempts to open a Redis client connecting to either:

```
redis://localhost:6379 or 
redis://127.0.0.1:6379
```

If neither of these work (which would likely be the case in a cloud environment with no Redis running on localhost), it falls back further to creating an in-memory client with 

```
redis://127.0.0.1:1
```

The idea here is that this in-memory client is not expected to connect anywhere—it’s just a placeholder.

> You can test that this works locally using `redis-cli -h 127.0.0.1 -p :1`

## Increased timeout TTLs

The increased timeout TTL values help make the system more resilient to temporary issues. For example, by increasing the presence update's TTL from 3 to 10 seconds, the client is less likely to be marked as disconnected if it experiences brief network hiccups or delays. Similarly, the longer timeouts during reconnection and ping checks give Redis a bit more time to respond, which is particularly useful when the server is restarting or under high load. Overall, these changes help prevent false negatives (i.e., mistakenly assuming a healthy connection is dead) and allow a more forgiving environment during transient issues.

callicles:
should we close this?

PR #2068, opened 2025-02-28
run workflows in temporal cloud

DatGuyJonathan:
Supports both auth mechanisms supported by temporal (mtls & api key)


callicles (approved):
How does the namespacing work? with the temporal URL?

DatGuyJonathan:
> How does the namespacing work? with the temporal URL?

Yep, exactly. The namespace is part of the temporal cloud url.

PR #2071, opened 2025-03-05
Cjus/new redis client

cjus:
This PR introduces a potential implementation for a new Redis client.

Here's an outline of the key benefits we're working on in the `new_redis_client.rs`:

1. **Modular Architecture**
   - The new implementation is split into distinct modules:
     - `connection` - Handles connection management and reconnection logic
     - `leadership` - Manages distributed locks and leader election
     - `presence` - Handles presence updates and heartbeats
     - `messaging` - Manages pub/sub messaging
     - `mock` - Provides a mock implementation for testing
   - This separation of concerns makes the code more maintainable and easier to understand
   - Each module can be tested and modified independently

2. **Better Error Handling and Fallback**
   - Implements a sophisticated fallback mechanism with `MockRedisClient`
   - More robust error handling with proper context in error messages
   - Graceful degradation when Redis is unavailable
   - Clear logging with consistent "<NewRedisClient>" prefix for better debugging

3. **Improved Connection Management**
   - More sophisticated reconnection logic with exponential backoff
   - Better state management using `AtomicBool` for connection status
   - Separate connection pools for regular operations and pub/sub
   - More robust connection monitoring with proper timeout handling

4. **Thread Safety Improvements**
   - Uses `tokio::sync::RwLock` instead of `std::sync::Mutex` for better concurrency
   - Better handling of shared state with `Arc` wrapped components
   - Clear separation between thread-safe and non-thread-safe components
   - More efficient locking patterns with read/write locks

5. **Enhanced Testing Support**
   - Mock client implementation for testing without Redis
   - Clear separation of interfaces making it easier to mock components
   - Better testability due to modular design
   - Ability to inject mock implementations for different components

6. **Backward Compatibility**
   - Maintains compatibility with old API through `_compat` methods
   - Gradual migration path for existing code
   - Clear separation between new and legacy interfaces
   - No breaking changes while allowing for new features

7. **Resource Management**
   - Better cleanup in the `Drop` implementation
   - More reliable task cancellation
   - Proper cleanup of Redis connections and resources
   - Better handling of background tasks

8. **Modern Rust Patterns**
   - Uses newer Rust async/await patterns consistently
   - Better use of type system for safety
   - More idiomatic Rust code with proper trait implementations
   - Better use of Rust's ownership and borrowing rules

9. **Performance Improvements**
   - More efficient connection pooling
   - Better handling of long-running operations
   - More efficient locking patterns with RwLock
   - Reduced clone operations through better architecture

10. **Configuration and Setup**
    - More flexible configuration options
    - Better initialization patterns
    - Clearer setup process
    - More configurable components

11. **Documentation and Maintainability**
    - Better documented code with clear module structure
    - More consistent error handling patterns
    - Clearer separation of responsibilities
    - More maintainable codebase structure

12. **Operational Improvements**
    - Better logging with consistent prefixes
    - More detailed error reporting
    - Better monitoring capabilities
    - Clearer operational status through connection state

This new implementation represents a potentially significant improvement in terms of code organization, reliability, and maintainability while maintaining compatibility with existing code. This modular approach makes it easier to extend and modify individual components without affecting the entire system.


apps/framework-cli/src/infrastructure/redis/new_redis_client.rs
```diff
@@ -0,0 +1,1049 @@
...
+
+    pub enum ConnectionState {
+        Connected,
+        Disconnected,
+        Reconnecting,
+    }
+
+    #[derive(Clone)]
+    pub struct ConnectionManagerWrapper {
+        pub connection: Arc<RwLock<ConnectionManager>>,
```
callicles: As per the engineering call - this is thread safe. I don't think you need the `Arc` / `RwLock`. Is there something else that could make it that we need it?
https://docs.rs/redis/latest/redis/aio/struct.ConnectionManager.html

apps/framework-cli/src/infrastructure/redis/new_redis_client.rs
```diff
@@ -0,0 +1,1049 @@
...
+        pub pub_sub: Arc<RwLock<ConnectionManager>>,
```
callicles: Same here

apps/framework-cli/src/infrastructure/redis/new_redis_client.rs
```diff
@@ -0,0 +1,1049 @@
...
+        pub state: Arc<AtomicBool>, // true = connected
```
callicles: the AtomicBool is thread safe as well, why do you need the `Arc`?

apps/framework-cli/src/infrastructure/redis/new_redis_client.rs
```diff
@@ -0,0 +1,1049 @@
...
+            Ok(ConnectionManagerWrapper {
+                connection: Arc::new(RwLock::new(conn)),
+                pub_sub: Arc::new(RwLock::new(pub_sub_conn)),
+                state: Arc::new(AtomicBool::new(true)),
+            })
+        }
+
+        pub async fn ping(&self) -> bool {
+            let mut conn_lock = self.connection.write().await;
+            let timeout_future = time::timeout(
```
callicles: Why do we need the timeout there? isn't that built in the redis client?

apps/framework-cli/src/infrastructure/redis/new_redis_client.rs
```diff
@@ -0,0 +1,1049 @@
...
+                Duration::from_secs(2),
+                AssertUnwindSafe(async {
```
callicles: What does this do? why is that necessary?

apps/framework-cli/src/infrastructure/redis/new_redis_client.rs
```diff
@@ -0,0 +1,1049 @@
...
+        pub fn new(instance_id: String, config: RedisConfig) -> Self {
+            LeadershipManager {
+                instance_id,
+                redis_config: config,
+            }
+        }
+
+        pub async fn attempt_lock(
+            &self,
+            conn: Arc<RwLock<ConnectionManager>>,
```
callicles: Same here

apps/framework-cli/src/infrastructure/redis/new_redis_client.rs
```diff
@@ -0,0 +1,1049 @@
...
+        }
+
+        fn presence_key(&self) -> String {
+            format!("{}::{}::presence", self.key_prefix, self.instance_id)
+        }
+
+        // Update presence using a TTL key
+        pub async fn update_presence(
+            &self,
+            conn: Arc<RwLock<ConnectionManager>>,
```
callicles: Same there

apps/framework-cli/src/infrastructure/redis/new_redis_client.rs
```diff
@@ -0,0 +1,1049 @@
...
+        }
+
+        pub fn channel_for(&self, target: &str) -> String {
+            format!("{}::{}::msgchannel", self.key_prefix, target)
+        }
+
+        // Publish a message to a channel
+        pub async fn publish_message(
+            &self,
+            conn: Arc<RwLock<ConnectionManager>>,
```
callicles: Same here 😄

apps/framework-cli/src/infrastructure/redis/new_redis_client.rs
```diff
@@ -0,0 +1,1049 @@
...
+    pub instance_id: String,
+    pub connection_manager: connection::ConnectionManagerWrapper,
+    pub leadership_manager: leadership::LeadershipManager,
+    pub presence_manager: presence::PresenceManager,
+    pub messaging_manager: messaging::MessagingManager,
+
+    pub message_callbacks: Arc<RwLock<Vec<MessageCallback>>>,
+    pub listener_task: Arc<Mutex<Option<JoinHandle<()>>>>,
+    pub presence_task: Arc<Mutex<Option<JoinHandle<()>>>>,
+    pub connection_monitor_task: Arc<Mutex<Option<JoinHandle<()>>>>,
```
callicles: Same here for all the `JoinHandle` - do we need the Arc Mutex?

apps/framework-cli/src/infrastructure/redis/new_redis_client.rs
```diff
@@ -0,0 +1,1049 @@
...
+            // For mock, we do nothing
+            // placeholder for now
+        }
+        Ok(())
+    }
+
+    // -------------------------------
+    // Lock Lifecycle Functions
+    // -------------------------------
+    pub async fn register_lock(&self, name: &str, ttl: i64) -> anyhow::Result<()> {
```
callicles: Why do we need local locks on top of redis locks?

apps/framework-cli/src/infrastructure/redis/new_redis_client.rs
```diff
@@ -0,0 +1,1049 @@
...
+                            .attempt_reconnection(&client, &config)
+                            .await;
+                    }
+                }
+            }));
+        } else {
+            log::error!("<NewRedisClient> Failed to lock connection_monitor_task");
+        }
+
+        let create_monitor_task = || {
```
callicles: Can you explain the difference between the task above and this one? Feels like they are both doing pings to reconnect

cjus: @callicles thanks for your review!! For your earlier comments: this is only a draft PR - the goal of which is to look at the reorganization of functionality with a focus on iterative testing.  Changes for the concurrency usage / model are under way based on your earlier feedback.

apps/framework-cli/src/infrastructure/redis/redis_client.rs
```diff
@@ -1,39 +1,127 @@
...
+/// This module offers a wrapper around the Redis crate to provide support for
+/// multi-moose distributed concerns such as interprocess communication and data
+/// processing coordination via leadership management and presence tracking.
+///
+/// CONCURRENCY NOTES:
+///
+/// The async `ConnectionManager` from the Redis crate implements `Send` and
+/// is therefore safe to move between threads. However, its API requires
+/// mutable access (`&mut self`) for sending commands, which means that
+/// concurrent command execution is not supported out of the box.
```
callicles: That's an awesome write up, that's the level of research and understanding I am looking for. 

Follow up question - the redis crate docs says:

```
A ConnectionManager is a proxy that wraps a [multiplexed connection](https://docs.rs/redis/latest/redis/aio/struct.MultiplexedConnection.html) and automatically reconnects to the server when necessary.

Like the [MultiplexedConnection](https://docs.rs/redis/latest/redis/aio/struct.MultiplexedConnection.html), this manager can be cloned, allowing requests to be sent concurrently on the same underlying connection (tcp/unix socket).
```

instead of a mutex when multiple mutable reference are needed, Have you tried using the cloning approach?

apps/framework-cli/src/infrastructure/redis/redis_client.rs
```diff
@@ -61,768 +172,661 @@ impl Default for RedisConfig {
...
+            let msg = mock_client.get_queue_message(&source_queue).await?;
+            Ok(msg)
+        } else {
+            Ok(None)
+        }
     }
 
     pub async fn mark_queue_message(
-        &mut self,
+        &self,
```
callicles, line 462: Removing all those mutable will make the use of this so much easier!

apps/framework-cli/src/infrastructure/redis/connection.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// - `bool` - true if the connection is healthy, false otherwise
+    pub async fn ping(&self) -> bool {
+        let timeout_future = time::timeout(
+            Duration::from_secs(2),
+            AssertUnwindSafe(async {
+                redis::cmd("PING")
+                    .query_async::<_, String>(&mut *self.connection.lock().await)
+                    .await
+            })
+            .catch_unwind(),
```
callicles: Not sure if you were able to look into the complexity here given all the other work done on the PR.

It is not a lot - I am wondering if necessary, ie I would expect the timeout to be included in the redis client.

callicles (commented):
Nice Work! 

This is getting really clean. I have a few outstanding questions but this is much better.

apps/framework-cli/src/infrastructure/redis/redis_client.rs
cjus: I haven't, I'm testing this today.

apps/framework-cli/src/infrastructure/redis/redis_client.rs
```diff
@@ -61,768 +176,788 @@ impl Default for RedisConfig {
...
+    pub presence_manager: PresenceManager,
+    pub messaging_manager: MessagingManager,
+
+    pub message_callbacks: Arc<RwLock<Vec<MessageCallback>>>,
+    pub listener_task: Arc<RwLock<Option<JoinHandle<()>>>>,
+    pub presence_task: Arc<RwLock<Option<JoinHandle<()>>>>,
+    pub connection_monitor_task: Arc<RwLock<Option<JoinHandle<()>>>>,
+    pub fallback: Option<MockRedisClient>,
+
+    pub locks: Arc<RwLock<std::collections::HashMap<String, (String, i64)>>>,
```
callicles: That's probably a lefover in the rollback, we don't need that right? since the locks are on the redis side

apps/framework-cli/src/infrastructure/redis/redis_client.rs
```diff
@@ -61,768 +176,788 @@ impl Default for RedisConfig {
...
-            .await
-            .map_err(|e| {
-                error!("<RedisClient> Failed to subscribe to channels: {}", e);
-                e
-            })?;
+    pub fn stop_periodic_tasks(&mut self) -> anyhow::Result<()> {
+        let rt = tokio::runtime::Handle::try_current()?;
 
-        let callback = self.message_callbacks.clone();
+        rt.block_on(async {
```
callicles: Why do this need to be blocking?

apps/framework-cli/src/cli/routines.rs
```diff
@@ -332,12 +332,9 @@ async fn manage_leadership_lock(
             IS_RUNNING_LEADERSHIP_TASKS.store(false, Ordering::SeqCst);
         });
 
-        let mut client = redis_client.lock().await;
```
callicles, line 335: Do we still need the lock here?

apps/framework-cli/src/infrastructure/redis/connection.rs
cjus: That is no longer part of the code.

apps/framework-cli/src/infrastructure/redis/redis_client.rs
cjus: Redis does manage the locks on the server side, having this local cache does allow for local checks without making network calls to Redis for every lock status query.  The issue is keeping those in sync - which the current code doesn't do well and may be more trouble than it's worth.  So, I'm refactoring this out in the next commit.

apps/framework-cli/src/cli/routines.rs
cjus: Good catch, this wasn't strictly needed, now that that broadcast_messge is thread safe.

apps/framework-cli/src/infrastructure/redis/redis_client.rs
cjus: It's currently safe to remove this block.  Refactor will be part of next commit.

callicles (approved):

PR #2073, opened 2025-03-06
run workflow from ts api

DatGuyJonathan:


apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
```diff
@@ -191,6 +195,57 @@ def parse_timeout_to_timedelta(self, timeout_str: str) -> timedelta:
...
+    client_options: Dict[str, Any] = {
+        "target_host": temporal_url,
+        "namespace": namespace,
+    }
+
+    if "localhost" not in temporal_url:
+        # Handle non-local connections with TLS or API key
+        cert_path = os.environ.get("MOOSE_TEMPORAL_CONFIG__CLIENT_CERT", "")
+        key_path = os.environ.get("MOOSE_TEMPORAL_CONFIG__CLIENT_KEY", "")
+        api_key = os.environ.get("MOOSE_TEMPORAL_CONFIG__API_KEY", "")
```
callicles: I am not a big fan of going through env variable that are defined by the moose cli here. If someone changes the structure of the config they won't get a compile error that they need to change this part. 

If we could pass it in like the rest of the variables as arguments that would be better.

apps/framework-cli/src/framework/python/wrappers/scripts/python_worker_wrapper/worker.py
```diff
@@ -35,6 +35,57 @@ def collect_activities(workflow_dir: str) -> List[str]:
     log.info(f"Found {len(script_paths)} scripts")
     return script_paths
 
+async def create_temporal_connection(temporal_url: str) -> Client:
```
callicles: Looks like this code is duplicated? any way we could not replicate and reuse the version from the lib?

DatGuyJonathan: Okay, I did some refactoring in the latest.

callicles (approved):

PR #2081, opened 2025-03-12
Read config.toml to figure out the tasks order

514Ben:


callicles:
@514Ben is it ready for review?

514Ben:
> @514Ben is it ready for review?

Nope, I'll close and change approach to keep the filename the source of truth for the order

PR #2082, opened 2025-03-12
chore: another attempt at fixing the python release

callicles:


.github/workflows/release-cli.yaml
```diff
@@ -542,3 +539,50 @@ jobs:
         with:
           slack-bot-token: ${{ secrets.SLACK_BOT_TOKEN }}
           slack-webhook-url: ${{ secrets.SLACK_GITHUB_ACTIONS_WEBHOOK_URL }}
+
+  cleanup-pypi:
```
jmsuzuki, line 543: heh. finally pulled this out!
if we start using it across repos/projects, it would be nice to pull it a step farther and put it in it's own repo.

callicles: One thing at a time 😄

PR #2088, opened 2025-03-14
dmv2 olaptable -> inframap

DatGuyJonathan:


callicles (approved):

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -1314,11 +1315,99 @@ impl InfrastructureMap {
                 &project.project_location,
             )
             .await?;
-            todo!("transform the output to infra map, maybe with a intermediate struct")
+            let json = serde_json::to_string(&_objects)?;
```
phiSgr: you can remove the leading underscore

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -1314,11 +1315,99 @@ impl InfrastructureMap {
...
+            Self::from_json_string(&json).map_err(|e| {
+                anyhow::anyhow!("Failed to parse infrastructure map from TypeScript: {}", e)
+            })
         } else {
             todo!("Python implementation")
         }
     }
+
+    pub fn from_json_string(json: &str) -> Result<Self, serde_json::Error> {
+        let partial: PartialInfrastructureMap = serde_json::from_str(json)?;
```
phiSgr: you can use `serde_json::from_value` to avoid the work of parsing from string twice

PR #2091, opened 2025-03-18
dmv2 ingestpipeline-> inframap

DatGuyJonathan:


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -871,7 +871,11 @@ async fn router(
             ingest_route(
                 req,
                 // without explicit version, go to current project version
-                route.join(current_version),
+                if !project.features.data_model_v2 {
```
callicles, line 874: Can you explain why we need that flow control?

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -1392,22 +1429,138 @@ impl PartialInfrastructureMap {
...
+                        name: partial_topic.name.clone(),
+                        primitive_type: PrimitiveTypes::DataModel,
+                    },
+                };
+                (id.clone(), topic)
+            })
+            .collect()
+    }
+
+    fn convert_api_endpoints(&self, language: SupportedLanguages) -> HashMap<String, ApiEndpoint> {
```
callicles, line 1473: Are we changing the api for this as well?

apps/framework-cli/src/cli/local_webserver.rs
DatGuyJonathan: There are a couple of ways to hit the ingest api: `/ingest/foo` or `/ingest/foo/0.0`. This code was handling the former like the latter. So I added this flag to keep that while we're in this transitional phase, but dmv2 will just treat the former as-is.

apps/framework-cli/src/framework/core/infrastructure_map.rs
DatGuyJonathan: Sorry, I don't quite understand. Could you elaborate?

callicles: The current way of declaring consumption apis is through on disk layout. That's read through the rust part. Are we changing the API to declare consumption apis?

that might be the right thing to do, I am not sure I saw the paper prototypes there, hence the question

apps/framework-cli/src/project.rs
```diff
@@ -300,6 +307,8 @@ impl Project {
             }
         }
 
+        DATA_MODEL_V2_ENABLED.store(project_config.features.data_model_v2, Ordering::SeqCst);
```
phiSgr, line 310: i'm really uncomfortable with the global variable.

apps/framework-cli/src/framework/core/infrastructure/topic.rs
```diff
@@ -91,7 +92,11 @@ impl Topic {
             | PrimitiveTypes::DBBlock
             | PrimitiveTypes::ConsumptionAPI => {
                 // TODO have a proper version object that standardizes transformations
-                format!("{}_{}", self.name, self.version.as_suffix())
+                if is_data_model_v2_enabled() {
+                    self.name.to_string()
+                } else {
+                    format!("{}_{}", self.name, self.version.as_suffix())
```
phiSgr, line 98: If the ID matters that much maybe we can make `version` an `Option` that is `None` when it's dmv2?

apps/framework-cli/src/cli/local_webserver.rs
callicles: I wonder if we should keep a version attribute to the object in DMV2, so that the user can have similar experiences but without DCM, ie they would be declaring versions explicitly but have a field for it so that it doesn't have to be in the name.

apps/framework-cli/src/framework/core/infrastructure/topic.rs
DatGuyJonathan: :sob:


Screenshot 2025-03-18 at 11 15 07 AM

apps/framework-cli/src/project.rs
DatGuyJonathan: I think this is temporary, can be removed when we're not in the transitional phase anymore.

apps/framework-cli/src/cli/local_webserver.rs
DatGuyJonathan: Does that mean under the hood topics & tables are still versioned, but all the user facing stuff like apis and data model schemas always assume latest?

apps/framework-cli/src/framework/core/infrastructure_map.rs
DatGuyJonathan: I believe the current plan is to keep egress the same because the ux doesn't change much. We'd need to reconcile ingress & egress since they're read in different ways though.

But it sounds like blocks will be in dmv2, so maybe one day everything is through dmv2. We'd need to add that support in the serializer.

phiSgr (approved):

apps/framework-cli/src/cli/local_webserver.rs
DatGuyJonathan: Gonna merge this in so GL can test with python. I have the version thing we talked about on my todo list.

PR #2094, opened 2025-03-20
Python dmv2

phiSgr:


packages/py-moose-lib/moose_lib/dmv2.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def __init__(self, name: str, config: StreamConfig = StreamConfig(), **kwargs):
+        super().__init__()
+        self._set_type(name, self._get_type(kwargs))
+        self.config = config
+        _streams[name] = self
+
+    def add_transform(self, destination: "Stream[U]", transformation: Callable[[T], ZeroOrMany[U]]):
+        """Add a transformation that sends records to a single destination stream."""
+        self.transformations[destination.name] = (destination, transformation)
```
DatGuyJonathan, line 114: I had to change the ts side like this to use async transformation function. Not sure if we need to do something similar in python

Screenshot 2025-03-20 at 10 49 38 AM

phiSgr: good catch
i don't think async is as prevalent in python

callicles (approved):
Nice work - looks really clean

DatGuyJonathan (approved):

PR #2102, opened 2025-03-25
Update Redis leadership lock renewal logic and improve logging

cjus:
Updated the Lua script for lock expiration to enhance clarity and functionality. Modified the check_and_renew_lock method in RedisClient to streamline lock acquisition and renewal process.

### DIFF
--- a/apps/framework-cli/src/infrastructure/redis/leadership.rs
+++ b/apps/framework-cli/src/infrastructure/redis/leadership.rs
@@ -130,7 +130,15 @@ impl LeadershipManager {
         ttl: i64,
     ) -> anyhow::Result<bool> {
         let script = redis::Script::new(
-            "if redis.call('GET', KEYS[1]) == ARGV[1] then return redis.call('EXPIRE', KEYS[1], ARGV[2]) else return 0 end",
+            r#"
+            local current = redis.call('GET', KEYS[1])
+            if current == ARGV[1] then
+                redis.call('EXPIRE', KEYS[1], ARGV[2])
+                return 1
+            else
+                return 0
+            end
+            "#,
         );
 
         match script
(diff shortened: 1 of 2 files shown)


apps/framework-cli/src/infrastructure/redis/leadership.rs
```diff
@@ -141,11 +149,12 @@ impl LeadershipManager {
             .await
         {
             Ok(1) => {
-                log::debug!(
-                    "<RedisLeadership> Lock renewed: {} for instance {}",
-                    lock_key,
-                    instance_id
-                );
+                // Dont log the renewal, it's too verbose
```
callicles: Should we remove altogether? or set it to trace?

apps/framework-cli/src/infrastructure/redis/redis_client.rs
```diff
@@ -521,8 +521,27 @@ impl RedisClient {
     }
 
     pub async fn check_and_renew_lock(&self, name: &str) -> anyhow::Result<(bool, bool)> {
-        let had_lock = self.attempt_lock(name).await?;
-        let now_has_lock = self.renew_lock(name).await?;
+        let lock_key = format!("{}::{}::lock", self.config.key_prefix, name);
+        let ttl = LEADERSHIP_LOCK_TTL;
+
+        // First check if we already have the lock
+        let had_lock = self
```
callicles: Why do we check here if we are going to attempt locking anyway?

apps/framework-cli/src/infrastructure/redis/leadership.rs
cjus: I thought about that.  I think the trace would still end up logging in production, no?

apps/framework-cli/src/infrastructure/redis/redis_client.rs
cjus: It was an attempt to avoid a race condition.  However, after looking into this further I see the Lua script should be updated to return whether it has the lock  and whether it's a new lock acquisition.  There's potential here to clean this up.  Will work on an update to this PR.

apps/framework-cli/src/infrastructure/redis/leadership.rs
callicles: production is on `info` only

callicles (approved):

PR #2108, opened 2025-03-26
int in typescript data model

phiSgr:


callicles:
Do you need a review on this?

callicles (approved):

PR #2125, opened 2025-03-27
python dmv2 consumptionapi

DatGuyJonathan:
Edit:

Hm, not sure about these test failures. They passed when I pushed the branch https://github.com/514-labs/moose/actions/runs/14113075568, then I rebased and it started failing.

### DIFF
--- a/packages/py-moose-lib/moose_lib/query_param.py
+++ b/packages/py-moose-lib/moose_lib/query_param.py
@@ -175,24 +175,3 @@ def map_params_to_class(
             [v] = values
             constructor_args[field_name] = parse(v, field_type)
     return cls(**constructor_args)
-
-def normalize_query_params(query_params: dict[str, list[str]]) -> dict[str, Any]:
-    """
-    Normalize query parameters by unwrapping single-item lists to scalar values.
-
-    When query parameters are parsed with parse_qs(), all values are returned as lists,
-    even when there's only one value. This function unwraps single-item lists while
-    preserving multi-value lists.
-
-    Examples:
-    - {'color': ['red']} becomes {'color': 'red'}
-    - {'color': ['red', 'blue']} remains {'color': ['red', 'blue']}
-
-    Args:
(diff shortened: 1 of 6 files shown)


callicles (approved):

packages/py-moose-lib/moose_lib/query_param.py
```diff
@@ -175,3 +175,24 @@ def parse(param: str, t: scalar_types) -> Any:
...
+def normalize_query_params(query_params: dict[str, list[str]]) -> dict[str, Any]:
+    """
+    Normalize query parameters by unwrapping single-item lists to scalar values.
+
+    When query parameters are parsed with parse_qs(), all values are returned as lists,
+    even when there's only one value. This function unwraps single-item lists while
+    preserving multi-value lists.
+
+    Examples:
+    - {'color': ['red']} becomes {'color': 'red'}
```
phiSgr: What if `color` really is a list field? then the validation will fail.

what's wrong with using `map_params_to_class`?

DatGuyJonathan: Hm, how do I use it with ConsumptionApi class? It looks specific to how dmv1 is structured.

```
def map_params_to_class(
        params: dict[str, list[str]],
        field_def_list: list[QueryField],
        cls: type,
) -> Any:
```

phiSgr: The `ConsumptionApi` class has `columns: Columns[T]` inherited from `TypedMooseResource`. That function needs `list[QueryField]`.

But that feels weird, a consumption API does not have columns.

DatGuyJonathan: Gotcha. I did some refactoring, lmk if that's what you had in mind.

phiSgr (approved):

callicles:
Same problem I was fighting most of the day

callicles:
Fioxed on my PR

PR #2137, opened 2025-04-01
Enhance streaming functionality in py-moose-lib

cjus:
- Updated README.md to include new streaming features and usage examples.
- Added command-line interface (CLI) for managing streaming functions, including listing, running, and exporting transforms.
- Introduced streaming runner to handle message processing and transformations with Kafka integration.
- Implemented streaming function discovery and execution capabilities.
- Added support for multi-destination transforms.
- Updated requirements.txt to include kafka-python-ng.
- Added tests for streaming functionality to ensure proper behavior.

### DIFF
--- a/packages/py-moose-lib/README.md
+++ b/packages/py-moose-lib/README.md
@@ -1,3 +1,72 @@
 # Python Moose Lib
 
-Python package which contains moose utils
+Python package which contains moose utils and streaming functionality.
+
+## Features
+
+- Data model definitions
+- Streaming functions support
+- Stream transformations (single and multi-destination)
+
+## Streaming Functions
+
+The library now supports streaming functions, allowing Python projects to define transformations between streams and automatically execute them. This is equivalent to the functionality already provided for TypeScript Moose projects.
+
+### Basic Usage
+
+```python
+from pydantic import BaseModel
+from moose_lib.dmv2 import Stream, StreamConfig
+
+# Define your data models
+class SourceData(BaseModel):
+    name: str
+    value: int
+
+class DestinationData(BaseModel):
+    processed_name: str
+    processed_value: int
+
+# Create streams
+source_stream = Stream[SourceData](name="source_stream", config=StreamConfig())
+dest_stream = Stream[DestinationData](name="dest_stream", config=StreamConfig())
+
+# Define a transformation function
+def transform_data(data: SourceData) -> DestinationData:
+    return DestinationData(
+        processed_name=f"processed_{data.name}",
+        processed_value=data.value * 2
+    )
+
+# Add the transformation to the source stream
+source_stream.add_transform(dest_stream, transform_data)
+```
+
+### Multi-Destination Transforms
+
+For more complex scenarios, you can use multi-destination transforms:
+
+```python
+def multi_transform(data: SourceData):
+    return [
+        dest_stream1.routed(transform1(data)),
+        dest_stream2.routed(transform2(data))
+    ]
+
+source_stream.set_multi_transform(multi_transform)
+```
+
+### Command Line Interface
+
+The package includes a command-line interface for running streaming functions:
+
+```bash
+# List available streaming functions
+moose-streaming list
+
+# Run a streaming function
+moose-streaming run --source-topic source_topic --target-topic dest_topic --transform-key source_stream_0_0_dest_stream_0_0
+
+# Export streaming function information
+moose-streaming export
+```
--- a/packages/py-moose-lib/moose_lib/streaming.py
+++ b/packages/py-moose-lib/moose_lib/streaming.py
@@ -1,0 +1,187 @@
+"""
+Streaming function discovery and execution for Python Moose projects.
+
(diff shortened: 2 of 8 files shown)


packages/py-moose-lib/README.md
```diff
@@ -1,3 +1,72 @@
 # Python Moose Lib
 
-Python package which contains moose utils
+Python package which contains moose utils and streaming functionality.
```
DatGuyJonathan, line 3: I'm concerned about maintaining these docs and the ones on https://docs.getmoose.dev. Should we pick one?

packages/py-moose-lib/moose_lib/streaming.py
```diff
@@ -0,0 +1,187 @@
...
+            )
+            
+            # Create a special key for multi-transforms
+            # We could implement this differently, but for now using a special key format
+            multi_transform_key = f"{stream_name}_multi_transform"
+            transform_functions[multi_transform_key] = multi_transform_func
+    
+    return transform_functions
+
+def load_transforms(module_path: Optional[str] = None) -> Dict[str, Callable]:
```
DatGuyJonathan, line 73: This looks similar to the loader in `internal.py`. Can that one be re-used?

packages/py-moose-lib/moose_lib/streaming_cli.py
```diff
@@ -0,0 +1,117 @@
...
+import argparse
+import json
+import os
+import sys
+from typing import Dict, List, Optional
+
+from .streaming import get_streaming_functions, load_transforms, print_transform_info, get_transform_info
+from .streaming_runner import run_streaming_function
+
+def parse_args():
```
DatGuyJonathan, line 21: Ooh a proper parser!

packages/py-moose-lib/moose_lib/streaming_runner.py
```diff
@@ -0,0 +1,331 @@
...
+        
+        # Thread management
+        self._running = False
+        self._threads = []
+        
+    def _create_consumer(self) -> KafkaConsumer:
+        """Create and configure a Kafka consumer."""
+        config = {
+            'bootstrap_servers': self.broker,
+            'group_id': f'moose-streaming-{self.source_topic}-{self.target_topic or "multi"}',
```
DatGuyJonathan, line 114: iirc, there's an assumption for redpanda acls, it expects a certain format for the group id so if we change this, infra likely has to change as well.

packages/py-moose-lib/requirements.txt
```diff
@@ -6,6 +6,7 @@ importlib_metadata==8.0.0
 jaraco.classes==3.4.0
 jaraco.context==5.3.0
 jaraco.functools==4.0.1
+kafka-python-ng==2.2.2
```
DatGuyJonathan, line 9: Should we change `python-project.rs` to not add this dependency to the project?

cjus:
I updated this PR.  To debug and continue testing, create a moose python project.

Here are my `moose.config.toml` settings

```toml
[features]
streaming_engine = true
workflows = false
data_model_v2 = true
```

Here is a sample `models.py` data model:

```python
from pydantic import BaseModel
from typing import Optional
from datetime import datetime

class Foo(BaseModel):
    primary_key: str
    timestamp: float
    optional_text: Optional[str] = None

class Bar(BaseModel):
    primary_key: str
    utc_timestamp: datetime
    has_text: bool
    text_length: int 
```

and here is a sample `main.py` for the app folder:

```python
from datetime import datetime
from moose_lib.dmv2 import (
    IngestPipeline, DataModelConfigV2
)
from app.datamodels.models import Foo, Bar

def transform_foo_to_bar(foo: Foo) -> Bar:
    return Bar(
        primary_key=foo.primary_key,
        utc_timestamp=datetime.fromtimestamp(foo.timestamp),
        has_text=foo.optional_text is not None,
        text_length=len(foo.optional_text) if foo.optional_text else 0
    )

# Create Foo pipeline (stream + ingestion)
foo_pipeline = IngestPipeline[Foo](
    name="Foo",
    config=DataModelConfigV2(
        table=False,  # No table needed for Foo
        stream=True,  # Need stream for transformation
        ingest=True   # Need ingestion endpoint
    )
)

# Create Bar pipeline (table + stream)
bar_pipeline = IngestPipeline[Bar](
    name="Bar",
    config=DataModelConfigV2(
        table=True,   # Store Bar data
        stream=True,  # Need stream to receive transforms
        ingest=False  # No direct ingestion needed
    )
)

# Connect the pipelines
assert foo_pipeline.stream is not None
assert bar_pipeline.stream is not None
foo_pipeline.stream.add_transform(bar_pipeline.stream, transform_foo_to_bar)
```

I've been testing with the following curl command:

```sh
curl -X POST "http://localhost:4000/ingest/Foo" \
  -H "Content-Type: application/json" \
  -d '{"primary_key": "test2", "timestamp": '$(date +%s)', "optional_text": "another test"}'
  ```

PR #2139, opened 2025-04-01
Refactor Moose Init

georgevanderson:


callicles:
feels like we should have the `typescript` and `python` templates in the repo in this branch before we merge. So that it is equivalent functionality before we merge

templates/ts/template.config.toml
```diff
@@ -0,0 +1,10 @@
+language = "typescript" # Must be typescript or python
+description = "default ts project"
+post_install_print = """
+	$ cd {project_dir}
+
+Install Dependencies:
+	$ npm i
+Start Moose Server:
+	$ moose dev
```
phiSgr, line 9: I have a preference for using the moose (`npm run dev`) in the dependency - more self-contained.

templates/ts/app/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from "./datamodels/models";
+
+export * from "./functions/process";
+export * from "./apis/bar";
+export * from "./blocks/blocks";
```
phiSgr: do we still want to call them blocks?

templates/ts/app/functions/process.ts
```diff
@@ -0,0 +1,12 @@
+import { FooPipeline, BarPipeline, Foo, Bar } from "../datamodels/models";
+
+if (FooPipeline.stream && BarPipeline.stream) {
```
phiSgr: why do the check instead of `FooPipeline.stream!.addTransform(BarPipeline.stream!`

templates/ts/app/apis/bar.ts
```diff
@@ -0,0 +1,40 @@
...
+  ) => {
+    const query = sql`
+        SELECT 
+          dayOfMonth,
+          ${CH.column(orderBy)}
+        FROM BarAggregated_MV
+        WHERE 
+          dayOfMonth >= ${startDay} 
+          AND dayOfMonth <= ${endDay}
+        ORDER BY ${CH.column(orderBy)} DESC
```
phiSgr: `${mv.targetTable.columns[orderBy]}`

apps/framework-cli/build.rs
```diff
@@ -1,6 +1,25 @@
...
+
+    Command::new(&package_templates_dir)
+        .current_dir(root_dir)
+        .status()?;
+    Ok(())
+}
 
 fn main() -> Result<()> {
+    // Package templates first
+    package_templates()?;
```
phiSgr, line 22: Does it run every time I run cargo build?

phiSgr (commented):
Maybe don't check in the package lock?

phiSgr (approved):
LFG

PR #2156, opened 2025-04-02
wip streaming function dmv2

callicles:


packages/py-moose-lib/moose_lib/streaming/streaming_function_runner.py
```diff
@@ -0,0 +1,512 @@
...
+    Raises:
+        SystemExit: If module import fails or if no matching transformation is found
+    """
+    # Add the function directory to Python path so we can import the main module
+    sys.path.append(function_file_dir)
+    
+    try:
+        # Import the main module which should define the streams
+        import_module(function_file_name)
+        module = import_module("moose_lib.dmv2")
```
phiSgr: This can be on the top, now that it's in moose_lib?
`from moose_lib.dmv2 import _streams`

packages/py-moose-lib/moose_lib/streaming/streaming_function_runner.py
```diff
@@ -0,0 +1,512 @@
...
+    def topic_name_to_stream_name(self) -> str:
+        """Returns the topic name with any namespace prefix removed."""
+
+        name = self.name
+        if self.version is not None:
+            version_suffix = f"_{self.version}".replace(".", "_")
+            name = name.replace(version_suffix, "")
+
+        if self.namespace is not None:
+            name = name.split(self.namespace + ".")[-1]
```
phiSgr: `removeprefix`?
https://docs.python.org/3/library/stdtypes.html#str.removeprefix

phiSgr: maybe add a `startswith` assert as well

packages/py-moose-lib/moose_lib/streaming/streaming_function_runner.py
```diff
@@ -167,6 +167,14 @@ def load_streaming_function_dmv2() -> tuple[type, Callable]:
     Raises:
         SystemExit: If module import fails or if no matching transformation is found
     """
+    sys.path.append(function_file_dir)
```
phiSgr, line 170: my hunch is that we don't need to append to `sys.path` if the `function_file_name` is `"app.main"` - this is hardcoded into `dmv2-serializer.py` anyway.

phiSgr (approved):

PR #2157, opened 2025-04-02
better error message when trying to run dev before npm i

DatGuyJonathan:


callicles (approved):

callicles (approved):

phiSgr (approved):
this only helps when moose is globally installed - which i don't like :P

PR #2162, opened 2025-04-03
booooookkkkkssss — goodreads data with some edits to other prints

oatsandsugar:
ridiculous title because github cleaning forced me to create like a half dozen branches.


templates/goodreads/.aurora/books.csv
callicles, line 1: Should .aurora be committed to the repo? Usually those are hidden directories as a default on Mac OS

oatsandsugar: It was remoed

PR #2172, opened 2025-04-03
Olivia/default python workflow

okane16:
- add generate workflow to python boilerplate
- fix config issues in workflow docs
- update quickstart to fix typescript default project name switch
- rename py template to python to match typescript update
- a little python MV cleanup


georgevanderson (approved):

apps/framework-docs/src/pages/moose/building/materialized-views.mdx
```diff
@@ -128,14 +125,22 @@ ClickHouse's aggregate functions offer several advantages when used in materiali
...
+<TypeScript>
 4. **Type safety**: The `Aggregated<T>` type in Moose ensures correct mapping between TypeScript and ClickHouse aggregation types
+</TypeScript>
 
 When using the `AggregatingMergeTree` engine, ClickHouse maintains the state of your aggregations between updates, making it much more efficient than re-aggregating from raw data each time.
 
 ### Common Aggregate Functions
 
+<Python>
+Examples coming soon
```
armanjindal, line 137: Hopefully we can add the HR demo MV once it's in DMV2

I like the idea of taking a block of CH SQL and showing how this would be written in Moose code. We should also link to CH MV docs ( https://clickhouse.com/docs/materialized-views)

That way we can facilitate the translation step. Devs + LLMs can rely on clickhouse docs to generate the MV and then use these examples to translate to moose conventions.

PR #2174, opened 2025-04-03
terminate all workflows on dev shutdown

DatGuyJonathan:


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1331,6 +1332,39 @@ async fn shutdown(settings: &Settings, project: &Project, graceful: GracefulShut
         let should_shutdown_containers = settings.should_shutdown_containers();
 
         if should_shutdown_containers {
+            super::display::show_message_wrapper(
```
callicles: Why is this conditional on terminating the containers? Shouldn't we terminate anyway all the time?

apps/framework-cli/src/cli/routines/scripts.rs
```diff
@@ -261,6 +263,87 @@ pub async fn terminate_workflow(
...
+        namespace: DEFAULT_TEMPORTAL_NAMESPACE.to_string(),
+        page_size: 1000,
+        query: "ExecutionStatus = 'Running'".to_string(),
+        ..Default::default()
+    };
+
+    let response = client_manager
+        .execute(|mut client| async move {
+            client
+                .list_workflow_executions(request)
```
callicles, line 288: There is no pagination on this? all is returned?

apps/framework-cli/src/cli/routines/scripts.rs
```diff
@@ -261,6 +263,87 @@ pub async fn terminate_workflow(
...
+    }
+
+    let termination_futures: Vec<_> = executions
+        .into_iter()
+        .filter_map(|execution| execution.execution)
+        .map(|execution_info| {
+            let client_manager = Arc::clone(&client_manager);
+            async move {
+                let request = TerminateWorkflowExecutionRequest {
+                    namespace: DEFAULT_TEMPORTAL_NAMESPACE.to_string(),
```
callicles: Default? should that be dynamic?

cjus (approved):
LGTM

apps/framework-cli/src/cli/local_webserver.rs
DatGuyJonathan: Good call, will update.

apps/framework-cli/src/cli/routines/scripts.rs
DatGuyJonathan: Their api has a `page_size` which I'm setting as 1000. Should that be smaller?

apps/framework-cli/src/cli/routines/scripts.rs
DatGuyJonathan: Good call, will update.

apps/framework-cli/src/cli/routines/scripts.rs
callicles: I think that's ok for now - maybe add a comment in there.

callicles (approved):

apps/framework-cli/src/cli/routines/scripts.rs
DatGuyJonathan: Okay cool, updated

PR #2176, opened 2025-04-03
clean up for dmv2 in ts

phiSgr:


callicles (commented):
You should probably update all the templates with this as well

phiSgr:
no other dmv2 in templates

callicles (approved):

PR #2186, opened 2025-04-07
auto start scheduled workflows

DatGuyJonathan:


apps/framework-cli/src/framework/scripts/executor.rs
```diff
@@ -111,6 +113,62 @@ async fn execute_workflow_for_language(
...
+/// - Can only start workflows that don't require input parameters, as there's no way
+///   to determine what the input parameters should be at startup time.
+/// - Workflows must have a non-empty schedule string in their config.toml
+///
+/// # Arguments
+/// * `project` - The project configuration containing workflow settings and paths
+///
+/// # Returns
+/// * `Result<(), WorkflowExecutionError>` - Success or an error if workflow startup fails
+pub(crate) async fn execute_scheduled_workflows(
```
callicles: What happens if for some reason the workflow is already running?

I believe we have code to make sure it stops but what if it doesn't

DatGuyJonathan: Good catch, it would've error'd out. I've updated it to be more resilient.

callicles (approved):

PR #2198, opened 2025-04-08
moose ls for dmv2

phiSgr:
```
View Moose primitives & infrastructure

Usage: moose-cli ls [OPTIONS]

Options:
      --type <TYPE>  Filter by infrastructure type (tables, streams, ingestion, sql_resource, consumption)
      --name <NAME>  Filter by name (supports partial matching)
      --json         Output results in JSON format
  -h, --help         Print help
  ```

### DIFF
--- a/apps/framework-cli/src/cli/commands.rs
+++ b/apps/framework-cli/src/cli/commands.rs
@@ -3,9 +3,8 @@
 
 use std::path::PathBuf;
 
-use clap::{Args, Subcommand};
-
 use crate::framework::languages::SupportedLanguages;
+use clap::{Args, Subcommand};
 
 #[derive(Subcommand)]
 pub enum Commands {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: apps/framework-cli/src/cli/commands.rs [hunk 1]
```diff
@@ -84,16 +83,28 @@ pub enum Commands {
...
         version: Option<String>,
 
         /// View streaming topics
-        #[arg(short, long, default_value = "false")]
+        #[arg(short, long, default_value = "false", hide = true)]
         streaming: bool,
+
+        /// Filter by infrastructure type (tables, streams, ingestion, sql_resource, consumption)
+        #[arg(long)]
+        _type: Option<String>,
```
DatGuyJonathan, line 99: What does this flag look like if the name has an underscore?

apps/framework-cli/src/cli/routines/ls.rs
```diff
@@ -348,3 +355,326 @@ fn format_topics(
...
+            .collect(),
+        consumption_apis,
+        stream_transformations: infra_map
+            .function_processes
+            .into_values()
+            .filter(|api| name.is_none_or(|name| api.name.contains(name)))
+            .map(|p| p.into())
+            .collect(),
+    };
+    let listing: &dyn ResourceInfo = match _type {
```
DatGuyJonathan, line 651: Could they ls for multiple types?

apps/framework-cli/src/cli/commands.rs
phiSgr: `type` is a keyword in rust
`clap` knows that and turns `_type` to `type`, see updated description

apps/framework-cli/src/cli/routines/ls.rs
phiSgr: maybe, but probably worth merging this in now

DatGuyJonathan (approved):

PR #2200, opened 2025-04-10
added configuration docs

tg339:


apps/framework-docs/src/pages/moose/reference/configuration.mdx
```diff
@@ -0,0 +1,111 @@
+# Project Configuration
```
callicles, line 1: Instead of this I would recommend to put as reference a full toml file annotated with the config. That's less to translate as a user

okane16 (approved):

callicles (approved):

PR #2208, opened 2025-04-10
Nicolas/moose 1449 source only streaming functions

callicles:
* Fixes cleanup of python resources (streaming and workflow worker) when dev server is shutdown
* Adds e2e test for the python project
* Made the target optional for functions on both python and typescript
* Docker update Temporal
* Update templates to show the new functionality


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1403,12 +1435,12 @@ async fn shutdown(settings: &Settings, project: &Project, graceful: GracefulShut
     }
 
     // Final delay before exit to ensure any remaining tasks complete
-    tokio::time::sleep(std::time::Duration::from_millis(200)).await;
+    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
georgevanderson, line 1438: Out of curiosity, did this need to be bumped up? any reason for exactly 1s?

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -1788,6 +1788,7 @@ struct PartialTopic {
     pub partition_count: usize,
     pub transformation_targets: Vec<TransformationTarget>,
     pub target_table: Option<String>,
+    pub has_consumers: bool,
```
phiSgr, line 1791: i think having a count is better than just the boolean

apps/framework-cli/src/cli/local_webserver.rs
callicles: Hum that probably could be tuned down, I may do that in a subsequent PR.
There is still the node temporal worker that hangs and I will tackle that one in a follow up PR

phiSgr (commented):
do you think we really need to specifically have an api for stream consumers. what about simply using the multi transform api where the provided function can write to anywhere or nowhere

georgevanderson (approved):

apps/framework-cli/src/framework/core/infrastructure_map.rs
callicles: We can iterate our way there. I just wanted something that we can build on. 

Regarding the multi-transformer, honestly, I haven't looked into too much detail on how that works. I remember we talked about it, but I think we can start with this and change later if we need to.

packages/ts-moose-lib/src/dmv2/internal.ts
```diff
@@ -161,6 +168,12 @@ export const getStreamingFunctions = async () => {
       const transformFunctionKey = `${stream.name}_${destination.name}`;
       transformFunctions.set(transformFunctionKey, f);
     });
+
+    stream._consumers.forEach((consumer) => {
+      // TODO: Add version to dmv2 apis
+      const consumerFunctionKey = `${stream.name}_<no-target>`;
```
phiSgr, line 174: This will overwrite and only one will remain right?

georgevanderson (approved):

phiSgr (commented):
I'm expecting to find the code in `runner.ts` where it looks for the function in the field `_consumers`?

packages/ts-moose-lib/src/dmv2/internal.ts
callicles: Are you refereing to the case with multiple consumers?

if so not supported yet although the API is there.
I was thinking we could spin up 1 consumer and just chain the functions

PR #2218, opened 2025-04-11
Template-fixes

03cranec:


templates/typescript/app/apis/bar.ts
```diff
@@ -1,41 +1,52 @@
...
+
+    //Define the query for the API endpoint - plug in the query parameters and data model fields as variables
     const query = sql`
         SELECT 
-          dayOfMonth,
-          ${BarAggregatedMV.targetTable.columns[orderBy]}
-        FROM BarAggregated_MV
+          ${BACols.dayOfMonth} as dayOfMonth,
+          ${BACols[orderBy]}
+        FROM BarAggregated
```
phiSgr, line 40: this can be `${BarAggregatedMV.targetTable}` after my PR

phiSgr (commented):
need to `gitignore` `.ts-node` output and maybe also the package lock?

templates/typescript/app/apis/bar.ts
03cranec: how soon is your PR coming? should i hold off and make the update here, or @okane16 can do it later?

03cranec:
> need to `gitignore` `.ts-node` output and maybe also the package lock?

@phiSgr i tried adding those to gitignore.. but not sure if i did it right? they aren't greyed out in my IDE yet, like i'd expect.

templates/typescript/app/apis/bar.ts
okane16: ill do later

03cranec:
@okane16 i've now got merge conflicts here 😢 do you want to take this over?

PR #2219, opened 2025-04-11
various polishing for dmv2

phiSgr:
peek
fix ts nullable nested
`OlapTable` in ts sql interpolation, sql in `MaterializedViewOptions.selectStatement`

The example materialized view then become

```ts
const barTable = BarPipeline.table!
export const BarAggregatedMV = new MaterializedView<BarAggregated>({
  tableName: "BarAggregated",
  materializedViewName: "BarAggregated_MV",
  orderByFields: ["dayOfMonth"],
  selectStatement: sql`SELECT
    toDayOfMonth(${barTable.columns.utcTimestamp}) as dayOfMonth,
    count(${barTable.columns.primaryKey}) as totalRows,
    countIf(${barTable.columns.hasText}) as rowsWithText,
    sum(${barTable.columns.textLength}) as totalTextLength,
    max(${barTable.columns.textLength}) as maxTextLength
  FROM ${barTable}
  GROUP BY toDayOfMonth(utcTimestamp)
  `,
});
```


apps/framework-cli/src/cli/routines/peek.rs
```diff
@@ -89,6 +89,8 @@ pub async fn peek(
 
         consumer_ref = create_consumer(&project.redpanda_config, &[("group.id", &group_id)]);
         let consumer = &consumer_ref;
+        // in dmv2 we still have version as the suffix for topics
+        // but not so for tables. we might want to change that
```
callicles, line 93: I believe we are adding back versions in tables for backward compatibility with existing deployments

callicles (approved):

PR #2222, opened 2025-04-13
rfd005: Moose version manager system

callicles:
This is a proposal for how to handle versions between cli and lib.


rfd/0005/README.mdx
```diff
@@ -0,0 +1,453 @@
...
+
+## Scope
+
+This RFD focuses specifically on the Moose CLI version management system. The following are explicitly out of scope:
+
+- Language-specific libraries (moose-libs):
+  - These are managed through their respective language package managers (npm, pip, etc.)
+  - Installation and versioning handled by the project's dependency management system
+  - Libraries are project-local and not directly called by users
+  - Version compatibility is enforced through dependency specifications in package.json/requirements.txt
```
DatGuyJonathan, line 21: Would this change how we develop moose-lib locally? I typically use `npm link`. When we `npm install`, I think the version is the actual version (i.e. 0.4.x), but the ts-moose-lib package.json has `0.0.2`. I'm not sure if it will get confused.

rfd/0005/README.mdx
```diff
@@ -0,0 +1,453 @@
...
+curl -sSL https://www.fiveonefour.com/install.sh | sh -s moose
+
+# Install specific Moose version
+moose install 1.0.0
+
+# Install latest Moose version
+moose install latest
+
+# Update Moose project version
+moose update latest  # Updates project config files
```
DatGuyJonathan: If a user updates, things break, and want to go back to the version they had before, how do we do that?

rfd/0005/README.mdx
```diff
@@ -0,0 +1,453 @@
...
+   - Read versions from package.json/requirements.txt
+   - Basic version switching based on project
+   - Use latest for moose init
+   - Channel specification in project files
+
+### Phase 1: Enhanced Version Management
+Goal: Improve version management reliability and user experience
+
+1. Version Cache Management:
+   - Implement version limit (100)
```
DatGuyJonathan, line 353: I was trying to see how big the binary is. It looks like around 25MB? https://github.com/514-labs/moose/releases/tag/v0.4.89

100 would be around 2.5GB. Could that be too much (assuming binary size grow over time)?

rfd/0005/README.mdx
```diff
@@ -0,0 +1,453 @@
...
+  - Installation and versioning handled by the project's dependency management system
+  - Libraries are project-local and not directly called by users
+  - Version compatibility is enforced through dependency specifications in package.json/requirements.txt
+
+## Version Selection Rules
+
+The system follows these rules for selecting which version of the CLI to use:
+
+1. Inside a Moose Project:
+   - Use the version specified in the project's package.json or requirements.txt
```
jmsuzuki: should be packag-lock? and setup.py?

I guess my real question is, what should take precedence when all things exist?
package.json AND package-lock.json
setup.py AND requirements.txt

rfd/0005/README.mdx
```diff
@@ -0,0 +1,453 @@
...
+   - This ensures compatibility between CLI and project libraries
+
+2. Outside a Moose Project:
+   - Always use the latest installed version
+   - This applies to commands like `moose init` for creating new projects
+   - If no version is installed, automatically fetch and use the latest available version
```
jmsuzuki: how does this apply when moose isn't installed?
no moose, no moose commands

rfd/0005/README.mdx
callicles: that's a good point, I am thinking of adding a way to shortcut as an option. 

As I'm writing this, I was also thinking that since this is just a wrapper script that's going to be on the moose name, then you can still start moose in dev mode with directly accessing the binary.

rfd/0005/README.mdx
jmsuzuki: if this is running as a result of the version manager from trying to do something within a project... we should download and install the project version.

rfd/0005/README.mdx
callicles: yes! I am adding that

rfd/0005/README.mdx
callicles: Good call, I will add that

rfd/0005/README.mdx
```diff
@@ -0,0 +1,453 @@
...
+The proposed system consists of three main components:
+
+1. A version-aware wrapper script (installed as `moose`)
+2. A local version cache manager
+3. A remote version installer
+
+Key design principles:
+- Unlike NVM, versions are automatically selected based on project configuration
+- Moose CLI only operates within Moose projects
+- Latest installed version is used outside of project contexts (e.g., for `moose init`)
```
jmsuzuki, line 61: shouldn't we match the versions being run?

rfd/0005/README.mdx
callicles: Maybe?

I was thinking that it would be configurable so that people want less, they can have less.

rfd/0005/README.mdx
```diff
@@ -0,0 +1,453 @@
...
+- Single installation method (curl-based) works across all supported languages
```
jmsuzuki, line 62: important to note that the install commands expects user interaction...
may need to be thoughtful about how we implement this here.

rfd/0005/README.mdx
callicles: Yes, I believe that last hyphen doesn't apply, good catch

rfd/0005/README.mdx
callicles: match it with what in this case?

rfd/0005/README.mdx
callicles: Good reminder! Thank you

rfd/0005/README.mdx
```diff
@@ -0,0 +1,453 @@
...
+   - Install version management wrapper
+   - Support version-specific installation via flags
+   - Handle PATH setup for both interactive and MCP usage
+
+```bash
+# Install Aurora (unchanged behavior)
+curl -sSL https://www.fiveonefour.com/install.sh | sh -s aurora
+
+# Install Moose wrapper
+curl -sSL https://www.fiveonefour.com/install.sh | sh -s moose
```
jmsuzuki, line 213: should this also install latest?

rfd/0005/README.mdx
jmsuzuki: agreed! how do we use specific versions?

rfd/0005/README.mdx
callicles: Yes

rfd/0005/README.mdx
callicles: `moose use 0.3.4` ?

DatGuyJonathan (approved):

PR #2224, opened 2025-04-14
ts dmv2 versioning

DatGuyJonathan:


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -867,20 +870,62 @@ async fn router(
...
-                    route
-                },
-                configured_producer,
-                route_table,
-                is_prod,
-                jwt_config,
-            )
-            .await
+            if project.features.data_model_v2 {
+                // For v2, find the latest version if no version specified
```
DatGuyJonathan, line 874: In dmv1, we had a behavior where hitting `ingest/foo` takes you to `ingest/foo/0.0`.

Now that version is per dmv2 object, when the user hits `ingest/foo`, I try to search for the latest version. If none are found, hit `ingest/foo` directly. Lmk if that's not the behavior we want.

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -2038,7 +2057,7 @@ impl PartialInfrastructureMap {
 
             let data_model = crate::framework::data_model::model::DataModel {
                 name: partial_api.name.clone(),
-                // TODO pass through version from the TS / PY api
+                // TODO: what is this version?
                 version: Version::from_string("0.0".to_string()),
```
DatGuyJonathan, line 2067: ApiEndpoint has a data model object in it. What should this version? Same as the api?

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -2192,21 +2230,21 @@ impl PartialInfrastructureMap {
...
-                    name: process_id.clone(),
+                    name: process_name.clone(),
                     source_topic_id: source_topic.id(),
                     target_topic_id: Some(target_topic.id()),
                     executable: main_file.to_path_buf(),
                     language,
                     parallel_process_count: target_topic.partition_count,
-                    // TODO pass through version from the TS / PY api
+                    // TODO: what is this version? topic? which one?
                     version: "0.0".to_string(),
```
DatGuyJonathan: What should this be? Are functions actually versioned?

apps/framework-cli/src/framework/core/infrastructure/api_endpoint.rs
```diff
@@ -159,7 +165,8 @@ impl From<EndpointFile> for ApiEndpoint {
             },
             path: value.path.clone(),
             method: Method::GET,
-            version: Version::from_string("0.0.0".to_string()),
+            // TODO: why does consumption api use patch version, but others only have major.minor?
+            version: Some(Version::from_string("0.0.0".to_string())),
```
DatGuyJonathan, line 169: I believe this is dmv1 code path so it can prob stay hardcoded. But the comment still stands, I'm not sure why consumption api uses major.minor.path.

dmv2 would have to specify "0.0.0" for the diff to be the same.

callicles: That's because that was the default here and the default in `package.json` in the default application was `0.0` is there a good reason for that?

Unfortunately not

callicles (approved):

apps/framework-cli/src/framework/core/infrastructure_map.rs
callicles: Shouldn't they be versioned?
if someone is shipping a new version of the same function it probably should change topic output or you could have a blue green type of deployment and have both run at the same time with same source and target

apps/framework-cli/src/framework/core/infrastructure_map.rs
callicles: I wonder if that should be a datamodel or a reference. ie the datamodel Id. that way we would not have issues

phiSgr: I think we'll drop the `DataModel` from the infra map when we no longer have dmv1?

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -2183,7 +2220,8 @@ impl PartialInfrastructureMap {
             for transformation_target in &source_partial_topic.transformation_targets {
                 debug!("transformation_target: {:?}", transformation_target);
 
-                let process_id = format!("{}_{}", topic_name, transformation_target.name);
+                // In dmv1, the process name was the file name which had double underscores
+                let process_name = format!("{}__{}", topic_name, transformation_target.name);
```
phiSgr, line 2226: will this change cause problems when a user upgrades?

DatGuyJonathan: It shouldn't. I made this change because there was a diff when I compared dmv1 vs dmv2, so I believe they would've had a problem without this change.

apps/framework-cli/src/framework/core/infrastructure_map.rs
DatGuyJonathan: I've added versioning to functions. If the same function has multiple versions, it currently transforms to each version for now. I'll add a ticket for exploring sending a percentage to a version.

PR #2229, opened 2025-04-15
add workflow to typescript template, revamped quickstart

okane16:


apps/logs-frontend/next-env.d.ts
callicles, line 1: Why do you have this file? that should be deleted I believe

apps/product-analytics/next-env.d.ts
callicles, line 1: Same here

templates/typescript/package-lock.json
callicles, line 1: I don't think you want to commit this right? that got removed on purpose

callicles (commented):
I believe you have some tests that are failing - are you good solving them or do you need help looking into it?

callicles (approved):

PR #2234, opened 2025-04-15
remove package template from build.rs

phiSgr:


callicles:
Why?

phiSgr:
> Why?

I think the caching won't run the script (called by build.rs) again if the binary is already built.
We're seeing some flakiness with a test that depends on the script being run.
https://github.com/514-labs/moose/actions/runs/14475607273/job/40600408561

Then in the release workflow, build is run inside docker, so node is not available (unless we install it, of course). If we don't really need it as part of the rust build process I'd like it to "do one thing and one thing well".

callicles:
Ok - then you need to make it easy for people to use then

callicles (approved):

PR #2240, opened 2025-04-16
decimal, datetime64, date32

phiSgr:


apps/framework-cli/src/infrastructure/olap/clickhouse.rs
```diff
@@ -901,7 +901,7 @@ fn convert_clickhouse_type_to_column_type(ch_type: &str) -> Result<(ColumnType,
     }
 
     // Handle DateTime types with parameters
-    if ch_type.starts_with("DateTime") {
+    if ch_type.starts_with("DateTime") || ch_type == "Date32" {
```
phiSgr, line 904: I don't like that this function skips over `ClickHouseColumnType::from_type_str` and have duplicate logic.

callicles: ie you want to refactor in a follow up?

callicles (approved):

callicles:
Also how does the user specify those types? ie in the loop that @armanjindal  is doing he is building data models that correspond to those types.

apps/framework-cli/src/infrastructure/olap/clickhouse.rs
phiSgr: yap

PR #2249, opened 2025-04-18
egress & templates

okane16:


apps/framework-docs/src/pages/moose/building/consumption-apis.mdx
```diff
@@ -511,7 +520,92 @@ The `sql` template literal in Moose provides type-safe query construction with p
...
+        LIMIT {params.limit}
+    """
+    
+    return client.query.execute(query)
+```
+</Tabs.Tab>
+</Tabs>
+
+<Callout type="warning">
+When using Pattern 2 (f-strings), you MUST be extremely careful with your Pydantic model definitions:
```
phiSgr, line 598: i think an objection can be, why bother mentioning f-strings at all - because it makes the demos a little nicer to at?

armanjindal (approved):

PR #2267, opened 2025-04-23
rfc(0006): improving change handling and sql dependencies

callicles:
This pull request introduces a new RFD (Request for Discussion) proposing a robust system for managing SQL dependencies in Moose, specifically targeting ClickHouse Materialized Views, Views, and Tables. The proposed solution addresses issues with unordered DDL execution during development by introducing explicit dependency tracking and a phased execution plan based on dependency analysis. Below is a breakdown of the most critical changes:

### Dependency Management Enhancements:
* **Explicit Dependency Tracking:** Introduced a `depends_on` property in the Moose SDK for defining dependencies between SQL resources (e.g., Tables, Views, Materialized Views). Dependencies are explicitly declared by users and transformed into structured JSON for processing in the Rust backend.
* **Rust Dependency Representation:** Extended the `InfrastructureSignature` enum in Rust to include all resource types (e.g., Table, View, SqlResource) and updated relevant structs to use non-optional dependency fields. This ensures dependencies are explicitly represented in the infrastructure map.

### Phased Execution Plan:
* **Dependency Analysis Step:** Added an intermediate step to analyze dependencies, build a Directed Acyclic Graph (DAG), and generate ordered execution plans for teardown and setup/modify operations. This ensures correct DDL execution order and prevents dependency-related errors.
* **Simplified Execution Phase:** Updated the `execute_changes` function to process the ordered `teardown_plan` and `setup_modify_plan` sequentially

### DIFF
--- a/rfd/0006/readme.mdx
+++ b/rfd/0006/readme.mdx
@@ -1,0 +1,293 @@
+---
+authors: Nicolas Joseph <nicolas@fiveonefour.com>
+title: Moose SQL Dependency Management for ClickHouse
+state: prediscussion
+---
+
+## Abstract
+
+This RFD proposes a solution to reliably manage Data Definition Language (DDL) changes for dependent SQL objects within Moose,
+specifically focusing on ClickHouse Materialized Views (MVs), Views, and Tables during development (`moose dev`). The current approach
+executes changes independently, leading to potential DDL errors when dependent objects are modified concurrently. The proposed solution mandates
+explicit dependency tracking in resource definitions and introduces an intermediate step between infrastructure planning (diffing) and execution.
+This step will analyze dependencies using a graph, decompose changes into atomic operations, and generate an ordered, phased execution plan (Teardown,
+Setup/Modify) based on topological sorting, similar to migration tool principles.
+
+## Motivation / Problem Description
+
+Moose provides a declarative way for users to define infrastructure, including ClickHouse tables, materialized views, and other SQL resources.
+During development (`moose dev`), the framework compares the user's desired state with the current state, generates a plan of changes
+(Adds, Removes, Updates), and executes these changes against the ClickHouse database.
+
+Currently, each infrastructure component change (`TableChange`, `SqlResource` change, etc.) is calculated and executed independently.
+This poses a significant problem when dealing with SQL objects that have inherent dependencies, most notably:
+
+1.  **Materialized Views (MVs):** An MV reads from one or more source tables/views (potentially via JOINs) and writes to a target table.
+    Its creation/update depends on all sources and the target existing and being correct.
+2.  **Chained MVs/Views:** MV2/View2 might read from the target table of MV1 or another View1.
+3.  **Views on Tables/Views:** A standard View depends on the objects in its `SELECT` statement.
+
+If a user modifies both a dependent object (e.g., an MV's `SELECT` query) and any of its dependencies (e.g., a source table schema, the target
+table schema) in the same development iteration, the current system generates separate, unordered change operations. The execution order of
+these independent changes is not guaranteed to respect the database constraints.
+
+This can lead to errors during `moose dev` runs, such as:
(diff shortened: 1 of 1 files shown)


rfd/0006/readme.mdx
```diff
@@ -0,0 +1,293 @@
...
+    *   **JSON Output:** The SDK transforms this list of instances into the structured dependency JSON format
+        `[{ resource_type: "TypeString", id: "UniqueIDString" }, ...]` which is then sent to the Rust CLI.
+    *   Example (Conceptual TS - User Code):
+        ```typescript
+        import { sourceTable1, sourceTable2, intermediateView } from './resources';
+        // Assume sourceTable1 etc. are instances of OlapTable, View...
+
+        new MaterializedView<TargetTableType>('my_mv', {
+            materializedViewName: 'my_mv_name',
+            selectStatement: `SELECT ... FROM ${sourceTable1.name} JOIN ${sourceTable2.name} ...`,
```
phiSgr, line 86: the `selectStatement` can be a `sql` interpolated string. We can look for its dependencies in the interpolated values.

callicles: Yeah but it also can be a hardcoded string. So I am thinking we start with explicit and then we are making it optional if there is string interpolation.

rfd/0006/readme.mdx
```diff
@@ -0,0 +1,293 @@
...
+### Part 2: Intermediate Dependency Analysis & Phased Ordering Step
+
+This step generates an ordered execution plan based on the explicit dependencies.
+
+1.  **Position:** `diff -> [Dependency Analysis & Phased Ordering] -> execute_changes`
+2.  **Input:**
+    *   The "flat" list of `OlapChange`s generated by `diff`.
+    *   The target `InfrastructureMap` (containing resources with populated `depends_on` fields).
+3.  **Process:**
+    *   **Build Dependency Graph:** Construct a Directed Acyclic Graph (DAG) of all relevant resources (Tables, SqlResources representing MVs, Views)
```
DatGuyJonathan, line 157: Could the user accidentally create a circular dependency (A depends on B depends on C depends on A)? Prob easier to make this mistake with larger code bases when it gets harder to track where things are. Would we just error to the user to fix that manually?

DatGuyJonathan: Also, would it be helpful to be able to visualize this DAG? I suppose the user can do "find all references" manually in the ide but that can get convoluted in large code bases. And maybe it's useful to show on the platform too? 

https://docs.rs/petgraph/latest/petgraph/ seems cool, plus I like the name lol

rfd/0006/readme.mdx
```diff
@@ -0,0 +1,293 @@
...
+    4.  **`diff(actual, target)`:** Compares state from Step 2 and Step 3. Finds no difference regarding `Table T1` (it's absent in both).
+    5.  **Plan Generation:** The plan will *not* contain any operations for `Table T1`. It will focus only on other differences caused by the previous failure.
+    6.  **Execution:** The corrective plan is executed.
+    7.  **Save State:** On success, the framework saves the target state *without* `Table T1`.
+
+**Conclusion for Part 4:**
+
+By incorporating the step of determining the actual database state before diffing, the proposed phased execution model becomes resilient to inconsistencies
+caused by prior failures or external changes. It ensures that the generated plan always aims to transition the database from its *current reality*
+to the user's *intended target state*, rather than relying solely on the potentially stale saved state.
```
DatGuyJonathan, line 272: I really like this snippet. It was challenging to poke holes at this doc and I think this is the crux of it.

rfd/0006/readme.mdx
callicles: Yeah I believe that's spot on. We could later add a check to make sure there is no circular dependencies and flag it to the user if one pops up

callicles: Agreed on the graph - @514Ben was also saying how useful that is for him working with his customer on data wrangling

PR #2269, opened 2025-04-23
fix: turning off a topic->table sync should not fail with topic/table not found

phiSgr:
uses sync.id as the key, so we don't need to call `get_table` and `get_table`.


apps/framework-cli/src/infrastructure/processes/kafka_clickhouse_sync.rs
```diff
@@ -50,8 +50,10 @@ const MAX_BATCH_SIZE: usize = 100000;
 struct TableSyncingProcess {
     /// Async task handle for the sync process
     process: JoinHandle<anyhow::Result<()>>,
+    #[allow(dead_code)]
     /// Source Kafka topic name
     topic: String,
```
callicles: Why not just remove the properties if they are not used anymore?

callicles (approved):

PR #2270, opened 2025-04-23
Fix create topic to table sync processes

DatGuyJonathan:


apps/framework-cli/src/framework/core/partial_infrastructure_map.rs
```diff
@@ -575,7 +575,11 @@ impl PartialInfrastructureMap {
                 let target_table = tables
                     .values()
                     .find(|table| {
-                        let name_matches = *table.name == *target_table_name;
+                        // This uses starts_with because the target_table_name from dmv2 serializer
+                        // uses names without version and it's being compared to the infra map table name
+                        // that has version suffix (for dmv1 parity). So it's a partial match on name
+                        // then it does version match right after.
+                        let name_matches = table.name.starts_with(target_table_name);
```
callicles: That can't work. If there is a BarAggregated table and a Bar table it will be random which one get found

callicles (changes requested):

apps/framework-cli/src/framework/core/partial_infrastructure_map.rs
DatGuyJonathan: Gotcha, I've updated it to strict name check.

callicles (approved):

PR #2277, opened 2025-04-25
fix blocking script

DatGuyJonathan:


apps/framework-cli/src/framework/python/wrappers/scripts/python_worker_wrapper/activity.py
```diff
@@ -65,10 +67,16 @@ async def dynamic_activity(execution_input: ScriptExecutionInput) -> WorkflowSte
                 else:
                     result = await task_func()
             else:
+                # User could run blocking sync function (i.e. time.sleep)
+                # so we run it in in a thread to not block the event loop
+                # and let the worker's shutdown tasks clean up
```
DatGuyJonathan, line 72: If I do this:

```
@task(retries=1)
def task1(input: dict):
    task1.logger.info(f"task1 input: {input}")
    time.sleep(120)
    return {
        "task": "task1",
        "data": None
    }
```

Then ctrl+c, I noticed the shutdown tasks in `worker.py` don't get run because the user's sync function is blocking the event loop. Even after moose stops, there's still a worker python process.

callicles (approved):

PR #2281, opened 2025-04-29
Cjus/posthog to crate

cjus:
This PR introduces a specialized PostHog client (`posthog514client-rs`) designed specifically for use in Moose and other 514 CLI applications.


packages/posthog514client-rs/LICENSE
```diff
@@ -0,0 +1,20 @@
+MIT License
+Copyright (c) 2025 Tim Delisle, Nicolas Joseph
```
callicles: That should be Fiveonefour Labs Inc now

apps/framework-cli/src/analytics/posthog.rs
```diff
@@ -1,75 +1,62 @@
...
-#[derive(Debug, Clone)]
-pub struct PostHogEvent {
-    pub event: String,
-    pub distinct_id: String,
-    pub timestamp: DateTime<Utc>,
-    pub properties: serde_json::Value,
-}
-
 #[derive(Clone)]
 pub struct PostHogClient {
```
callicles: Why did you decide to keep this here? vs moving it to the lib?

packages/posthog514client-rs/LICENSE
cjus: Copied this from the moose LICENSE.  Should I update both?

callicles: yeah good call

apps/framework-cli/src/analytics/posthog.rs
cjus: I had a small wrapper that I was using in moose - I've removed that wrapper in favor of direct use.

callicles (approved):

PR #2282, opened 2025-04-29
Nicolas/moose 1460 add explicit dependency tracking for materialized views

callicles:
This pull request introduces several changes to enhance the infrastructure management system by adding improving data lineage capabilities, and updating protobuf definitions.

### DIFF
--- a/apps/framework-cli/src/framework/core/infrastructure.rs
+++ b/apps/framework-cli/src/framework/core/infrastructure.rs
@@ -31,6 +31,7 @@ pub mod topic_sync_process;
 pub mod view;
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Eq)]
+#[serde(tag = "kind")]
 /// Represents the unique signature of an infrastructure component.
 ///
 /// Infrastructure signatures are used to identify and reference various infrastructure
--- a/apps/framework-cli/src/framework/core/infrastructure.rs
+++ b/apps/framework-cli/src/framework/core/infrastructure.rs
@@ -14,6 +14,9 @@
 //!
 //! If components need to reference each other, they should do so by reference and not by value.
 
+use crate::proto::infrastructure_map::{
+    infrastructure_signature, InfrastructureSignature as ProtoInfrastructureSignature,
+};
 use serde::{Deserialize, Serialize};
 
 pub mod api_endpoint;
@@ -21,13 +24,13 @@ pub mod consumption_webserver;
 pub mod function_process;
 pub mod olap_process;
 pub mod orchestration_worker;
+pub mod sql_resource;
 pub mod table;
 pub mod topic;
 pub mod topic_sync_process;
 pub mod view;
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Eq)]
 /// Represents the unique signature of an infrastructure component.
 ///
 /// Infrastructure signatures are used to identify and reference various infrastructure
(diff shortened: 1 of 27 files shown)


apps/framework-cli/src/framework/core/infrastructure.rs
```diff
@@ -58,9 +129,6 @@ pub enum InfrastructureSignature {
 /// - Pulling: Active fetching of data from another component
 /// - Pushing: Active sending of data to another component
 pub trait DataLineage {
-    /// Returns infrastructure components that actively push data to this component.
-    fn receives_data_from(&self) -> Vec<InfrastructureSignature>;
```
callicles, line 62: This is the exact opposite of `pushes_data_to` as such the element that pushes the data will have that tagged already

apps/framework-cli/src/framework/core/infrastructure.rs
```diff
@@ -14,20 +14,24 @@
...
+pub mod sql_resource;
 pub mod table;
 pub mod topic;
 pub mod topic_sync_process;
 pub mod view;
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Eq)]
+#[serde(tag = "kind")]
```
phiSgr, line 34: this this impact backwards compatibility?

templates/typescript/app/views/barAggregated.ts
```diff
@@ -26,4 +26,5 @@ export const BarAggregatedMV = new MaterializedView<BarAggregated>({
   FROM ${barTable}
   GROUP BY toDayOfMonth(utcTimestamp)
   `,
+  selectTables: [barTable],
```
phiSgr, line 29: we're already referencing the columns from `barTable` in the `sql`, having to specify this again feels a bit repetitive.

but then this is explicit and matches the python api

packages/ts-moose-lib/src/dmv2/index.ts
```diff
@@ -12,29 +19,107 @@ import {
...
+/**
+ * Configuration options for a complete ingestion pipeline, potentially including an Ingest API, a Stream, and an OLAP Table.
+ * @template T The data type of the records being ingested.
+ */
 export type IngestPipelineConfig<T> = {
+  /**
+   * Configuration for the OLAP table component of the pipeline.
+   * If `true`, a table with default settings is created.
+   * If an `OlapConfig` object is provided, it specifies the table's configuration.
+   * If `false` or omitted, no OLAP table is created.
```
phiSgr: omitting is impossible though?

packages/py-moose-lib/moose_lib/main.py
```diff
@@ -159,9 +270,11 @@ def __init__(self, raw_strings: list[str], raw_values: list['RawValue']):
...
 
 def sigterm_handler():
+    """Handles SIGTERM signals by printing a message and exiting gracefully."""
     print("SIGTERM received")
     sys.exit(0)
 
 
 def stub():
+    """Placeholder function, prints a simple message."""
     print("Hello from moose-lib!")
```
phiSgr: maybe just remove this?

apps/framework-cli/src/framework/core/infrastructure/view.rs
```diff
@@ -61,6 +64,20 @@ impl View {
     }
 }
 
+impl DataLineage for View {
```
phiSgr, line 67: why are we adding stuff to this type anyway? it's subsumed by the more general `SqlResource`

callicles: Yeah we probably should remove the `View` type altogether. I added it for completness

packages/py-moose-lib/moose_lib/main.py
callicles: Good call

packages/ts-moose-lib/src/dmv2/index.ts
callicles: Fair will fix

templates/typescript/app/views/barAggregated.ts
callicles: Yeah - I am thinking we do it explicit for now and we improve the api after.
the select can also be a string that would be to be parsed and we would need to extract the table.
I would rather do that in a follow up improvement to the feature.

apps/framework-cli/src/framework/core/infrastructure.rs
callicles: I don't think it does - from what I can tell, this structure was not used in the storage.

callicles: It would impact backward compatibility between moose <> moose-lib but now that we enforce the versions are the same, I believe that should be fine.

apps/framework-cli/src/framework/core/infrastructure/view.rs
callicles: I am leaving it for now but we should clean that up in a subsequent PR

phiSgr (approved):

PR #2286, opened 2025-04-30
Add llms.txt file and markdown files to docs

georgevanderson:
This is to be consumed by aurora


apps/framework-docs/docs/llms.txt
callicles, line 1: should we call the folder `llm-docs` not to confuse with the main docs?

callicles (approved):

PR #2306, opened 2025-05-08
Remove IngestionFormat

georgevanderson:


packages/protobuf/infrastructure_map.proto
```diff
@@ -47,7 +47,7 @@ message ApiEndpoint {
 
 message IngressDetails {
   string target_topic = 1;
-  EndpointIngestionFormat format = 2;
+  EndpointIngestionFormat format = 2 [deprecated = true];
```
callicles: @georgevanderson That should be optional now no?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -837,24 +776,14 @@ async fn ingest_route(
...
-            EndpointIngestionFormat::JsonArray => Ok(handle_json_array_body(
-                &configured_producer,
-                &route_meta.kafka_topic_name,
-                &route_meta.data_model,
-                req,
-                &jwt_config,
-            )
-            .await),
-        },
+        Some((_, route_meta)) => Ok(handle_json_array_body(
```
callicles, line 760: Does this handle for arrays and single objects?

From the code that looks surprising

callicles (approved):

phiSgr (approved):

PR #2309, opened 2025-05-13
refresh command

phiSgr:


apps/framework-cli/src/cli/routines.rs
```diff
@@ -628,3 +638,153 @@ pub async fn remote_plan(
...
+                    debug!("Found matching table: {}", table.name);
+                    tables_to_integrate.push(table.name.clone());
+                }
+                Some(_) => warn_about_mismatch(&table.name),
+            }
+        }
+    }
+
+    // Check mismatched tables
+    for change in &reality_check.discrepancies.mismatched_tables {
```
callicles: Can you explain why you run this on the mismatched tables?

I was expecting it to skip given the `warn_about_mismatch` function above

phiSgr: warn about "local" and "remote reality" mismatch

"remote infra map" and "remote reality" mismatch is fine

callicles: ie you integrate the changes if they are between remote reality and remote infra map. But you don't if they are between local and remote reality

if that's it would be amazing to make that explicit in the comments

phiSgr: done

callicles (approved):

PR #2311, opened 2025-05-13
added sitemap and robot

tg339:


okane16 (approved):

callicles:
If you want to use Athena for docs you will need to set it with the API for it to be able to report user agents

PR #2326, opened 2025-05-17
Add Metadata map to propagate to infra map -- Python and TS

georgevanderson:
TS:

image
image
image

Python:

image
image
image


callicles:
you have tests that are failing :)

callicles (approved):

PR #2365, opened 2025-05-26
generate python code from database

phiSgr:
use it like so: `moose-cli init from_remote --from-remote 'http://panda:pandapass@localhost:18123/?database=local' --language python`

### DIFF
--- a/apps/framework-cli/src/cli.rs
+++ b/apps/framework-cli/src/cli.rs
@@ -18,7 +18,6 @@ use commands::{
 };
 use config::ConfigError;
 use display::{with_spinner, with_spinner_async};
-use home::home_dir;
 use log::{debug, info};
 use regex::Regex;
 use routines::auth::generate_hash_token;
@@ -34,6 +33,8 @@ use routines::scripts::{
     terminate_workflow, unpause_workflow,
 };
 use routines::templates::list_available_templates;
+use std::env;
+use std::io::Write;
 
 use settings::Settings;
 use std::path::Path;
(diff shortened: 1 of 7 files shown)


apps/framework-cli/src/cli.rs
```diff
@@ -151,12 +120,32 @@ pub async fn top_command_handler(
             location,
             template,
             no_fail_already_exists,
+            from_remote,
+            language,
+            // database,
```
callicles, line 110: Maybe can you split the body of this in smaller functions? that looks a lot of cases that are nested which males it harder to review

apps/framework-cli/src/cli/commands.rs
```diff
@@ -15,7 +15,11 @@ pub enum Commands {
         name: String,
 
         /// Template to use for the project
-        template: String,
+        #[arg(
+            conflicts_with = "from_remote",
```
callicles, line 18: That's cool! I did not know Clap had that

apps/framework-cli/src/cli/routines/templates.rs
```diff
@@ -318,6 +325,160 @@ pub async fn list_available_templates(
...
+                .map_err(|e| {
+                    RoutineFailure::error(Message {
+                        action: "Init".to_string(),
+                        details: format!("Failed to parse package.json: {}", e),
+                    })
+                })?;
+
+                if let Some(obj) = package_json.as_object_mut() {
+                    obj.insert(
+                        "name".to_string(),
```
callicles, line 388: Love this fix!

callicles (approved):
Maybe Small improvement on the body of cli.rs that I believe could be split in smaller chunks. Otherwise nice work! some bugs opportunistically squashed, I like it!

PR #2368, opened 2025-05-27
Nicolas/moose 1472 olaptable direct insert api

callicles:


packages/ts-moose-lib/src/dmv2/index.ts
```diff
@@ -541,15 +1233,17 @@ export class IngestPipeline<T> extends TypedBase<T, IngestPipelineConfig<T>> {
     config: IngestPipelineConfig<T>,
     schema: IJsonSchemaCollection.IV3_1,
     columns: Column[],
+    validators?: TypiaValidators<T>,
```
phiSgr: `?` not needed in this overload

PR #2369, opened 2025-05-27
typescript from-remote 

phiSgr:


apps/framework-cli/src/framework/typescript/generate.rs
```diff
@@ -0,0 +1,379 @@
...
+            precision: Some(precision),
+        } => format!(
+            "string & typia.tags.Format<\"date-time\"> & ClickHousePrecision<{}>",
+            precision
+        ),
+        ColumnType::Date => "string & typia.tags.Format<\"date\">".to_string(),
+        ColumnType::Date16 => {
+            "string & typia.tags.Format<\"date\"> & ClickHouseByteSize<2>".to_string()
+        }
+        ColumnType::Enum(data_enum) => enums.get(data_enum).unwrap().to_string(),
```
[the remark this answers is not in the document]
phiSgr: the enums map should have all of the definitions
same for the nested below

apps/framework-cli/src/cli/routines/code_generation.rs
```diff
@@ -73,7 +74,18 @@ pub async fn db_to_dmv2(remote_url: &str, dir_path: &Path) -> Result<(), Routine
 
     match project.language {
         SupportedLanguages::Typescript => {
-            todo!()
+            let table_definitions = tables_to_typescript(&tables);
+            let mut file = std::fs::OpenOptions::new()
+                .append(true)
+                .open("app/index.ts")
```
callicles: should that be a constant?

phiSgr: good call

PR #2378, opened 2025-05-28
From clickhouse docs

okane16:


apps/framework-docs/src/pages/moose/getting-started/from-clickhouse.mdx
```diff
@@ -0,0 +1,562 @@
...
+### 1. Get Some Data Locally
+
+```bash copy
+# CLI command to seed from remote ClickHouse
+moose seed from-remote-clickhouse <YOUR_CLICKHOUSE_CONNECTION_STRING> --rows 100
+```
+
+### 2. Build Materialized Views That Don't Break
+
+<ToggleBlock openText="See how Moose fixes common ClickHouse pain points" closeText="Hide details">
```
callicles, line 260: Why is this in an accordeon? That feels weird for the entire section to be empty unless you expand it.

callicles: I realize it is in the what's next, maybe it is weird because item 1 is not a collapsible, and it tells you what to do.

When running diagonally, it feels like a walkthrough that's off when it is not a walkthrough

callicles: ie I would be make eveything collapsable or make everything a full expanded section with instructions

03cranec (approved):

phiSgr (approved):

PR #2387, opened 2025-05-28
Moose seed

georgevanderson:


phiSgr (commented):
it seems if either the user's CH native port or CH http(s) port is non-standard,
there's no way to get it to work?

apps/framework-cli/src/utilities/clickhouse_url.rs
```diff
@@ -0,0 +1,44 @@
+use log::{debug, info};
+use reqwest::Url;
+
+/// Transforms a ClickHouse connection string to ensure it uses HTTP(S) protocol
+/// Handles both clickhouse:// and https:// URLs
+pub fn normalize_clickhouse_url(conn_str: &str) -> anyhow::Result<Url> {
```
phiSgr: this is not "normalizing"? more like "converting"

apps/framework-cli/src/cli/routines/seed_data.rs
```diff
@@ -0,0 +1,183 @@
...
+use log::{debug, info};
+
+fn parse_clickhouse_connection_string(conn_str: &str) -> anyhow::Result<ClickHouseConfig> {
+    let url = normalize_clickhouse_url(conn_str)?;
+
+    let user = url.username().to_string();
+    let password = url.password().unwrap_or("").to_string();
+    let host = url.host_str().unwrap_or("localhost").to_string();
+    let use_ssl = url.scheme() == "https";
+    let default_port = if use_ssl { 443 } else { 8123 };
```
phiSgr: clickhouse HTTPS is 443 though. having protocol default and application default on the same line feels weird

apps/framework-cli/src/cli/routines/seed_data.rs
```diff
@@ -0,0 +1,183 @@
...
+        .filter(|s| !s.is_empty())
+        .or_else(|| {
+            let path = url.path().trim_start_matches('/').to_string();
+            if !path.is_empty() {
+                Some(path)
+            } else {
+                None
+            }
+        })
+        .unwrap_or_else(|| "default".to_string());
```
phiSgr: i did `select database()` with from_remote

PR #2405, opened 2025-06-02
moose cache

DatGuyJonathan:


packages/ts-moose-lib/src/clients/redisClient.ts
```diff
@@ -0,0 +1,236 @@
+import { createClient, RedisClientType } from "redis";
+
+export class MooseCache {
```
DatGuyJonathan, line 8: Example:

```
interface Foo {
  value: string;
  timestamp: Date;
  blah?: string;
}

export const UserEventEgress = new ConsumptionApi<QueryParams, ResponseType>(
  "getUserEvents",
  async ({ limit = 5 }, { client, sql }) => {
    const cache = await MooseCache.get();
    await cache.set("testKeyString", "testStringValue");
    const value = await cache.get("testKeyString");

    const fooIn = {
      value: "tesObjecttValue",
      timestamp: new Date(),
    } as Foo;
    await cache.set("testKeyObject", fooIn);
    const fooOut = await cache.get<Foo>("testKeyObject");

    await cache.clearKeys("test");

    return {
      count: 2,
      events: [
        {
          id: "testKeyString",
          name: value || "not found",
        },
        {
          id: "testKeyObject",
          name: fooOut?.value || "not found",
        },
      ],
    }
  }
);
```

packages/ts-moose-lib/src/clients/redisClient.ts
```diff
@@ -0,0 +1,236 @@
+import { createClient, RedisClientType } from "redis";
```
jmsuzuki, line 1: love it! I believe the node style guide for file names is to spine case it.

jmsuzuki: https://elser.hashnode.dev/javascript-naming-convention-for-beginners#heading-file-names

cjus (commented):
Love that you maintained a sane keyspace path:  ${prefix}::moosecache::;

packages/ts-moose-lib/src/clients/redisClient.ts
```diff
@@ -0,0 +1,236 @@
...
+    value: string | object,
+    ttlSeconds?: number,
+  ): Promise<void> {
+    try {
+      await this.ensureConnected();
+      const prefixedKey = this.getPrefixedKey(key);
+      const stringValue =
+        typeof value === "object" ? JSON.stringify(value) : value;
+
+      if (ttlSeconds) {
```
cjus: If ttlSeconds isn't valid then the entry is cached indefinitely. We need to be careful with this.

DatGuyJonathan: Yea, do you think we should just enforce a max default?

packages/ts-moose-lib/src/clients/redisClient.ts
DatGuyJonathan: Yea, we haven't been consistent with ts-moose-lib. Most are camelcase, a handful are spinalcase

packages/ts-moose-lib/src/clients/redisClient.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  /**
+   * Gets the singleton instance of MooseCache. Creates a new instance if one doesn't exist.
+   * The client will automatically connect to Redis and handle reconnection if needed.
+   *
+   * @returns Promise<MooseCache> The singleton instance of MooseCache
+   * @example
+   * const cache = await MooseCache.get();
+   */
+  public static async get(): Promise<MooseCache> {
```
callicles, line 98: Did you double check that this was indeed making a singleton? I would do it differently

ie have an variable top of the file and return its value

packages/ts-moose-lib/src/clients/redisClient.ts
```diff
@@ -0,0 +1,236 @@
...
+   * @returns Promise<T | null> The value, parsed as type T if it was JSON, or as string if not. Returns null if key doesn't exist
+   * @example
+   * // Get a string
+   * const value = await cache.get("foo");
+   *
+   * // Get and parse an object with type safety
+   * interface Config { baz: number; qux: boolean; }
+   * const config = await cache.get<Config>("foo:config");
+   */
+  public async get<T = string>(key: string): Promise<T | null> {
```
callicles: can `T` really be anything?

packages/ts-moose-lib/src/clients/redisClient.ts
DatGuyJonathan: Added default ttl

packages/ts-moose-lib/src/clients/redisClient.ts
DatGuyJonathan: Good call, made it more explicit.

callicles (approved):

PR #2421, opened 2025-06-05
workflow dmv2 ts

DatGuyJonathan:
Example:

```
export interface Foo {
  value: string;
}

export interface Bar {
  value: string;
}

export interface Moo {
  value: string;
}

export const task3 = new Task<Moo>("task3", {
  run: async (input: Moo) => {
    console.log(`task3 input: ${JSON.stringify(input)}`);
  }
});

export const task2a = new Task<Bar, Moo>("task2a", {
  run: async (input: Bar) => {
    console.log(`task2a input: ${JSON.stringify(input)}`);

    return {
      value: `task2a processed: ${input.value}`,
    };
  },
  onComplete: [task3],
});

export const task2b = new Task<Bar>("task2b", {
  run: async (input: Bar) => {
    console.log(`task2b input: ${JSON.stringify(input)}`);
  }
});

export const task1 = new Task<Foo, Bar>("task1", {
  run: async (input: Foo) => {
    console.log(`task1 input: ${JSON.stringify(input)}`);

    return {
      value: `task1 processed: ${input.value}`,
    };
  },
  onComplete: [task2a, task2b],
  retries: 1,
  timeout: "5m"
});

export const testwfdmv2 = new Workflow("testwfdmv2", {
  startingTask: task1,
  retries: 1,
  timeout: "10m",
  schedule: "@every 45s",
});
```

### DIFF
--- a/apps/framework-cli/src/cli/routines/scripts.rs
+++ b/apps/framework-cli/src/cli/routines/scripts.rs
@@ -4,6 +4,7 @@ use std::sync::Arc;
 
 use crate::cli::display::{show_table, Message};
 use crate::cli::routines::{RoutineFailure, RoutineSuccess};
+use crate::framework::core::infrastructure_map::InfrastructureMap;
 use crate::framework::scripts::utils::get_temporal_namespace;
 use crate::framework::scripts::Workflow;
 use crate::infrastructure::orchestration::temporal_client::TemporalClientManager;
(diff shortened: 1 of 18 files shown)


callicles:
Love it!

callicles:
If not backwards compatible we may want to update the version to the 0.5 series. And also we need docs?

DatGuyJonathan:
> If not backwards compatible we may want to update the version to the 0.5 series. And also we need docs?

Yep, it is! I was running workflow 1.0 & 2.0 in parallel in the same project. Also added some draft docs, but hid the page from the nav bar since we don't have python examples yet. Still accessible by hitting the url directly if one of us needs some docs.

PR #2443, opened 2025-06-12
named tuples

phiSgr:


apps/framework-cli/src/infrastructure/olap/clickhouse_alt_client.rs
```diff
@@ -234,7 +234,7 @@ fn column_type_to_enum_mapping(t: &ClickHouseColumnType) -> Option<Vec<&str>> {
         | ClickHouseColumnType::AggregateFunction { .. }
         | ClickHouseColumnType::Bytes => None,
         ClickHouseColumnType::Array(t) => column_type_to_enum_mapping(t.as_ref()),
-        ClickHouseColumnType::Nested(_) => {
+        ClickHouseColumnType::NamedTuple(_) | ClickHouseColumnType::Nested(_) => {
```
callicles, line 237: Do we not need this anymore?

apps/framework-cli/src/infrastructure/processes/kafka_clickhouse_sync.rs
```diff
@@ -467,6 +467,7 @@ async fn sync_kafka_to_clickhouse(
                                 if let Ok(clickhouse_record) =
                                     mapper_json_to_clickhouse_record(&source_topic_columns, json_value)
                                 {
+                                    println!("ClickHouseRecord: {:?}", clickhouse_record);
```
callicles: leftover print?

apps/framework-cli/src/infrastructure/olap/clickhouse_alt_client.rs
phiSgr: still used in peek

PR #2457, opened 2025-06-20
s3source ts

DatGuyJonathan:
Example:

```
export interface PokemonResource {
  name: string;
  url: string;
}

export const PokemonPipeline = new IngestPipeline<PokemonResource>("Pokemon", {
  table: true,
  stream: true,
  ingest: true,
});

const s3source = new S3Source<PokemonResource>({
  bucket: "s3source",
  prefix: "pokemon-data",
  format: "json",
  auth: {
    accessKeyId: AWS_ACCESS_KEY_ID,
    accessKeySecret: AWS_SECRET_ACCESS_KEY_ID,
    region: AWS_REGION,
  },
});

const testConnectionResult = await s3source.testConnection();
console.log("Test connection result:", testConnectionResult);

await PokemonPipeline.table!.insert(await s3source.extract());

let count = 0;
for await (const pokemon of s3source) {
  console.log(`Found pokemon: ${pokemon.name}`);
  count++;
}
console.log(`Total pokemon found: ${count}`);
```


packages/ts-moose-lib/package.json
```diff
@@ -19,6 +19,7 @@
     "test": "mocha -r ts-node/register 'src/**/*.test.ts'"
   },
   "dependencies": {
+    "@aws-sdk/client-s3": "^3.832.0",
```
callicles: I don't think we should have the S3 source as part of `moose-lib` that should be a new package like:

`514labs/moose-s3` otherwise, for each connector and integration we will have a new dependency here and it will be bloated for people who don't use it.

callicles (commented):
Meta food for thought - we are going to have multiple sources that use various json, csv, etc... we need the logic for decoding those formats to be also extracted and collocated to be reused across connectors

packages/ts-moose-lib/package.json
DatGuyJonathan: Good call. I've moved it to its own package & tested with npm link locally. Updated the build scripts as well, I think it should just publish when it's merged to main.

DatGuyJonathan: I've split them up to their own packages. Lmk if that works!

callicles (approved):

PR #2461, opened 2025-06-20
import moose ConsumptionApi in streamlit app

phiSgr:


packages/py-moose-lib/moose_lib/dmv2/consumption.py
```diff
@@ -98,4 +102,38 @@ def get_response_schema(self) -> JsonSchemaValue:
         from pydantic.type_adapter import TypeAdapter
         return TypeAdapter(self.return_type).json_schema(
             ref_template='#/components/schemas/{model}'
-        )
\ No newline at end of file
+        )
+
+    def call(self, base_url: str, params: T) -> U:
```
callicles: Is this adding a proxy?

callicles (approved):

packages/py-moose-lib/moose_lib/dmv2/consumption.py
phiSgr: This is for python code outside of moose to connect to the rust server

PR #2473, opened 2025-06-24
Enhance workflow status retrieval with detailed failure summaries

cjus:
and improved event handling. Added failure summary output for verbose mode and refined error reporting for various event types.


cjus:
Note this PR only impacts the `moose workflow status` command. It works by fetching Temporal's event history for a particular workflow name and Run Id.  It returns errors, details and importantly, stack traces.

PR #2478, opened 2025-06-25
fix wf dmv2 from api

DatGuyJonathan:


apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
```diff
@@ -183,20 +217,16 @@ async def _start_workflow_async(self, name: str, input_data: Any):
...
-        )
+        return input_data, workflow_id
 
-        return workflow_handle.result_run_id
+    def _build_workflow_args(self, name: str, input_data: Any, is_dmv2: bool) -> list:
+        """Build workflow arguments based on workflow type."""
+        if is_dmv2:
+            return [f"{name}", input_data]
+        else:
+            return [f"{os.getcwd()}/app/scripts/{name}", input_data]
```
armanjindal, line 229: We should drop DMV1 backward compatibility soon!

apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
```diff
@@ -138,43 +138,77 @@ def __init__(self, temporal_client: TemporalClient):
...
-        timeout_str = self.configs[name]['timeout']
-        run_timeout = self.parse_timeout_to_timedelta(timeout_str)
+        return workflow_id, workflow_handle.result_run_id
+
+    def _get_workflow_config(self, name: str) -> Dict[str, Any]:
+        """Extract workflow configuration from DMv2 or legacy config."""
+        dmv2_workflow = get_workflow(name)
+        if dmv2_workflow is not None:
+            return {
+                'retry_count': dmv2_workflow.config.retries or 3,
```
armanjindal, line 188: May be nice to have these constant more explicit  (either in docs or in code)

PR #2482, opened 2025-06-27
browser compatible objects

phiSgr:
moved stuff around so that a next project can import a subproject called `moose-objects` which is also depended on by `apps/moose-backend`


callicles (approved):
This feels a bit weird to me, aren't method on the browser compatible objects not going to work?

callicles:
After reading the PR description that may not be what we expose to the user, hence that might be fine?

phiSgr:
> method on the browser compatible objects not going to work

Direct inserts won't work right now, but if the users want to do it on the browser we can do a switch and decide whether to load `@clickhouse/client` or `@clickhouse/client-web` (and not try to load config from files).

PR #2504, opened 2025-07-01
fixed template workflows and e2e testing [ENG-243]

camelCasedAditya:


templates/ads-b-frontend/moose/app/index.ts
oatsandsugar, line 1: do you want to export * or export the named object? This isn't a leading question, both can be correct!

oatsandsugar (approved):
Looks good! Just consider the export in `index.ts`; decide if you prefer * or { object }

callicles:
@oatsandsugar test

callicles:
@oatsandsugar another test

templates/ads-b-frontend/moose/app/index.ts
camelCasedAditya: Exporting * as of now to stay consistent with the rest of the template. Could change all of it for a more detailed approach if you would like

oatsandsugar (approved):
I tested the ADSB templates, they are good. Typescript templates look good. did you test?

camelCasedAditya:
> I tested the ADSB templates, they are good. Typescript templates look good. did you test?

Yes. Worked fine for me. Not sure if this (https://github.com/514-labs/moose/pull/2503/files) might be causing a conflict/issue.

oatsandsugar:
@camelCasedAditya you might need to change the new DLQ template too

oatsandsugar (changes requested):
need changes on new DLQ version of ADSB template too

callicles (approved):

oatsandsugar (changes requested):
@camelCasedAditya did you look at @03cranec's comment on slack about importing the data contracts from the data models in the other primitives? I think we have to prioritize that.

camelCasedAditya:
> @camelCasedAditya did you look at @03cranec's comment on slack about importing the data contracts from the data models in the other primitives? I think we have to prioritize that.

I took a look at the message. I added a commit with some changes to ensure the data contract in workflows

PR #2511, opened 2025-07-02
fast api impl of the bar consumption api

phiSgr:
run `fastapi dev app/main.py` along side `moose dev`
and test it with `curl 'localhost:8000/bar?order_by=rows_with_text'`


okane16:
@phiSgr maybe we create a separate, dedicated template called python-fastapi?

okane16 (approved):

PR #2512, opened 2025-07-02
audit unwrap expected panics and swallowed errors

cjus:
This pull request includes various improvements across multiple files in the `apps/framework-cli` project, focusing on error handling, logging enhancements, and code robustness. The most significant changes include better error reporting, improved logging for debugging, and enhanced handling of edge cases in parsing and processing.


callicles:
Looks like this is not building

cjus:
> Looks like this is not building

@callicles clippy feedback resolve and tests now pass.  PR ready for review.

callicles (approved):

apps/framework-cli/src/cli.rs
```diff
@@ -81,7 +81,14 @@ fn load_project() -> Result<Project, RoutineFailure> {
 }
 
 fn check_project_name(name: &str) -> Result<(), RoutineFailure> {
-    let project_name_regex = Regex::new(PROJECT_NAME_ALLOW_PATTERN).unwrap();
+    let project_name_regex = Regex::new(PROJECT_NAME_ALLOW_PATTERN).map_err(|e| {
```
phiSgr: We know this `Result` won't fail. It has no external dependency and its input is a `const`.

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -676,8 +676,8 @@ async fn handle_json_array_body(
                 Ok(Value::Object(value)) => vec![value],
                 _ => {
                     info!(
-                        "Received payload for {} is not valid JSON objects or arrays. Not sending them to DLQ.",
-                        topic_name
+                        "Received payload for {} is not valid JSON objects or arrays. Not sending them to DLQ. Body: {:?}",
```
phiSgr: User data which might be sensitive, shouldn't be in `info` log.

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -692,12 +692,18 @@ async fn handle_json_array_body(
...
                         "source": "api",
+                        "requestBody": String::from_utf8_lossy(&body),
+                        "topic": topic_name,
                     }))
                     .unwrap()
                 }),
             )
             .await;
         }
+        error!(
```
phiSgr: `error` should be reserved for huge errors that almost worth waking someone up in the night

apps/framework-cli/src/cli/watcher.rs
```diff
@@ -291,7 +291,9 @@ impl FileWatcher {
                 redis_client,
             )
             .await
-            .unwrap()
+            {
+                error!("File watcher task failed: {:?}", e);
+            }
```
phiSgr: if you don't crash, the process will not be hot reloading and the user won't know without digging into the logs.

apps/framework-cli/src/cli/local_webserver.rs
cjus: Downgrading to warning

apps/framework-cli/src/cli/local_webserver.rs
cjus: Agree - updating PR

apps/framework-cli/src/cli.rs
cjus: Good catch. Makes sense. Updating PR.

apps/framework-cli/src/cli/watcher.rs
cjus: agreed - updated PR

PR #2514, opened 2025-07-02
Refreshable MV -- TS support

georgevanderson:


callicles (approved):

phiSgr:
this will be reworked based on #3080

PR #2517, opened 2025-07-03
Enhance error handling in workflow execution error messages

cjus:
The user will now see only the concise error message, e.g. Could not start workflow 'dog-fact-ingest': status: AlreadyExists, message: Workflow execution is already running. WorkflowId: dog-fact-ingest, RunId: ...

### DIFF
--- a/apps/framework-cli/src/cli/routines/scripts.rs
+++ b/apps/framework-cli/src/cli/routines/scripts.rs
@@ -114,7 +114,7 @@ pub async fn run_workflow(
             RoutineFailure::new(
                 Message {
                     action: "Workflow".to_string(),
-                    details: format!("Could not start workflow '{name}': {}", e),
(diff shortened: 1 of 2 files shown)


apps/framework-cli/src/cli/routines/scripts.rs
```diff
@@ -111,10 +111,18 @@ pub async fn run_workflow(
         .start(&project.temporal_config, input)
         .await
         .map_err(|e| {
+            // Remove trailing ', details: [...]' and ', metadata: ...' from error string if present
```
DatGuyJonathan: Do we still need this? I suspect it's already trimmed down through `concise_msg` below

cjus: You're correct.  Updated PR.

DatGuyJonathan (approved):

PR #2519, opened 2025-07-03
ENG-111 write to dead letter table in ts template

phiSgr:


templates/typescript/app/ingest/transforms.ts
```diff
@@ -20,7 +18,7 @@ FooPipeline.stream!.addTransform(
     };
   },
   {
-    deadLetterQueue: fooDead,
+    deadLetterQueue: FooPipeline.deadLetterQueue,
```
callicles, line 21: I understand how the model is the same but for the sake of the template shouldn't this be 2 deadletter queues?

one for ingest malformation and another one for transformations gone array?

phiSgr: the `errorType` can be used to differentiate them

callicles: Hum interesting - that's on the dead letter model?

callicles (approved):

PR #2520, opened 2025-07-03
Refactor log file cleanup to elevate error logging level

cjus:
From INFO to WARN. This change ensures that potential issues during log rotation are highlighted, preventing silent failures that could lead to disk space problems. Updated comments to clarify rationale for logging decisions.


phiSgr (approved):
i don't think it deserves this much justification in the code though

PR #2522, opened 2025-07-03
Improve developer experience when needing to enable backtrace logging.

cjus:
Adds a global `--backtrace` flag that sets `RUST_LIB_BACKTRACE=1`, letting users print full stack traces for any `anyhow` error or panic without memorizing env-vars. This accelerates triage, shortens customer support loops, and leaves the fast path untouched (no perf impact when the flag isn’t used). Change is isolated to CLI arg parsing and runs before any threads spawn, so it’s safe and zero-risk.


phiSgr:
I always do `RUST_BACKTRACE=full ~/moose/target/debug/moose-cli dev` when i develop

having stack traces is good (less useful in rust though. the panic position can be very far away from the origin of the error)

but is this change useful to put in our users?

cjus:
> I always do `RUST_BACKTRACE=full ~/moose/target/debug/moose-cli dev`
> 
> having stack traces is good (less useful in rust though. the panic position can be very far away from the origin of the error)
> 
> but is this change useful to put in our users?

I asked an LLM the same question:

consider this:

Why add --backtrace?

Developer experience
- When users hit an error, a single rerun with --backtrace gives an immediate stack trace—no hunting for the right environment variable.
- Cuts triage time dramatically, especially in CI or remote environments where interactive debugging is hard.

Supportability
- When we ask customers for “the failing command + --backtrace output”, we get actionable data in one paste.
- Eliminates the ping-pong: “Please set RUST_LIB_BACKTRACE=1 and rerun…”

Zero performance cost by default
- Back-traces are only captured when the flag (or env var) is set.
- Fast path remains unchanged; no additional allocations or logging overhead for normal runs.

Standards-compliant & future-proof
- Relies on the official anyhow::Error::backtrace() and the stabilized RUST_LIB_BACKTRACE mechanism.
- Works equally for panics (RUST_BACKTRACE) and anyhow errors—consistent mental model.

Minimal, isolated change
- Net diff ≈ 15 lines: one flag in the clap struct, one env-var set in main.
- Executed before any threads spawn → no safety concerns.
- No impact on downstream crates or public API.

phiSgr:
> When we ask customers for “the failing command + --backtrace output”, we get actionable data in one paste.
> Eliminates the ping-pong: “Please set RUST_LIB_BACKTRACE=1 and rerun…”

appending `--backtrace` and prepending `RUST_BACKTRACE=1` is basically the same

cjus:
> > When we ask customers for “the failing command + --backtrace output”, we get actionable data in one paste.
> > Eliminates the ping-pong: “Please set RUST_LIB_BACKTRACE=1 and rerun…”
> 
> appending `--backtrace` and prepending `RUST_BACKTRACE=1` is basically the same

Yes basically - but not quite.
--backtrace is easier to remember and easier to type.

Happy to close the PR if it doesn't seem useful.

PR #2524, opened 2025-07-03
mitchell/chore-reduce: number-of-days-pypi-packages are saved

jmsuzuki:
**chore: reduced days pypi packages are saved from 100 to 80**
we hit a size limit in pypi for our project.
to give us more room, the number of days old packages are kept
has been reduced from 100 days to 80 days.


callicles (approved):

callicles:
I am going to be merging this - there are a bunch of fixes that need to go out so we need this to be shipped

PR #2531, opened 2025-07-04
Remove old cron functionality

callicles:
Remove deprecated Tokyo cron functionality, migrating to Workflows v2's native scheduling.


apps/framework-cli/src/infrastructure/processes/cron_registry.rs
```diff
@@ -1,288 +0,0 @@
...
-                                let path = env::var("PATH")
-                                    .unwrap_or_else(|_| "/usr/local/bin".to_string());
-                                let bin_path = format!(
-                                    "{}:{}/node_modules/.bin",
-                                    path,
-                                    project_path_clone.to_str().unwrap()
-                                );
-                                let ts_script_path =
-                                    project_path_clone.join(script_path.as_ref().unwrap());
-                                Command::new("moose-exec")
```
DatGuyJonathan, line 179: Should we remove `moose-exec` as well?

cjus (approved):
Looks clean

apps/framework-cli/src/infrastructure/processes/cron_registry.rs
callicles: Good Call - done

PR #2532, opened 2025-07-04
Ensure tests run on every pull request

callicles:
Restructure CI workflow to separate and matrix unit and E2E tests for improved coverage and parallelization.


callicles:
It creates the node version in a separate one for the tests that are ts or python specific.
For the CLI since most of the test are rust based and we are moving language specific implementation in the libs I thought I was not necessary

PR #2539, opened 2025-07-08
esm error hint

DatGuyJonathan:
https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V22.md#requireesm-is-now-enabled-by-default

### DIFF
--- a/packages/ts-moose-lib/src/dmv2/internal.ts
+++ b/packages/ts-moose-lib/src/dmv2/internal.ts
@@ -400,7 +400,8 @@ const _loadIndex = () => {
     const details = error instanceof Error ? error.message : String(error);
     if (details.includes("ERR_REQUIRE_ESM") || details.includes("ES Module")) {
       hint =
-        "The file or its dependencies are ESM-only. Switch to packages that dual-support CJS & ESM, or upgrade to Node 22.12+";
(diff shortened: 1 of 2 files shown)


packages/ts-moose-lib/src/dmv2/internal.ts
```diff
@@ -383,7 +383,7 @@ if (getMooseInternal() === undefined) {
  * and `end___MOOSE_STUFF___`) for easy extraction by the calling process.
  */
 export const loadIndex = async () => {
-  await require(`${process.cwd()}/app/index.ts`);
+  _loadIndex();
```
phiSgr: maybe rename this to `dumpMooseInternal`
and drop the leading underscore

phiSgr: I mean `loadIndex` -> `dumpMooseInternal`; `_loadIndex` -> `loadIndex`

DatGuyJonathan: Ahh

phiSgr: :P

phiSgr (approved):

PR #2541, opened 2025-07-08
Add optional dependency management to prod command

camelCasedAditya:


apps/framework-cli/src/utilities/prod-docker-compose.yml.hbs
```diff
@@ -0,0 +1,184 @@
+volumes:
+  redpanda-0: null
```
callicles, line 3: This should be optional as well if streams are deactivated

callicles (approved):

PR #2545, opened 2025-07-09
typescript template tweaks [PM-228]

camelCasedAditya:


callicles (approved):

oatsandsugar (changes requested):
can you remove the git keep from the scripts folder too?

oatsandsugar (approved):

PR #2546, opened 2025-07-09
Configurable proxy port

georgevanderson:


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -130,10 +130,17 @@ pub struct LocalWebserverConfig {
     /// The port to bind the management server to
     #[serde(default = "default_management_port")]
     pub management_port: u16,
+    /// The port to bind the proxy server to (for consumption APIs)
+    #[serde(default = "default_proxy_port")]
+    pub proxy_port: u16,
```
phiSgr, line 135: call it consumption_api_port?

apps/framework-cli/src/framework/core/plan.rs
```diff
@@ -328,7 +328,10 @@ mod tests {
                 native_port: 9000,
                 host_data_path: None,
             },
-            http_server_config: crate::cli::local_webserver::LocalWebserverConfig::default(),
+            http_server_config: crate::cli::local_webserver::LocalWebserverConfig {
+                proxy_port: crate::cli::local_webserver::default_proxy_port(),
+                ..crate::cli::local_webserver::LocalWebserverConfig::default()
```
phiSgr: `default_proxy_port` is already used in `LocalWebserverConfig::default()`?

apps/framework-cli/src/project.rs
```diff
@@ -261,7 +261,10 @@ impl Project {
             redpanda_config: KafkaConfig::default(),
             clickhouse_config: ClickHouseConfig::default(),
             redis_config: RedisConfig::default(),
-            http_server_config: LocalWebserverConfig::default(),
+            http_server_config: LocalWebserverConfig {
+                proxy_port: crate::cli::local_webserver::default_proxy_port(),
+                ..LocalWebserverConfig::default()
```
phiSgr: here as well?

PR #2554, opened 2025-07-10
Improve error message for missing dependencies

callicles:


callicles:
Let me see if cursor can do this

PR #2561, opened 2025-07-14
Fix terminal state after moose dev interruption

callicles:
Fix terminal state after `moose dev` interruption.

The previous shutdown process did not fully reset the terminal, leading to a stuck cursor and visible escape sequences after Ctrl+C. This PR adds a comprehensive terminal restoration function, including a global panic handler, to ensure a clean terminal state.

Only look after #2559  is merged


phiSgr (approved):
great find.
many magical sequences though
if it's LLM output, is it cross checked against some other sources?
or better yet, can these functionalities be found in crates?

callicles:
Actually this doesn't work - I am still working on this

callicles:
I am going to get rid of all the different dependencies that we use for terminal writing feels like they might be colliding with one another

apps/framework-cli-e2e/test/templates.test.ts
```diff
@@ -443,24 +472,32 @@ describe("Moose Templates", () => {
...
+          console.error("Response code:", response.status);
+          const text = await response.text();
+          console.error(`Test request failed: ${text}`);
+          throw new Error(`${response.status}: ${text}`);
+        }
+        responses.push(response);
       }
 
-      await utils.waitForDBWrite(devProcess!, "Bar", 1);
+      await utils.waitForDBWrite(devProcess!, "Bar", recordsToSend);
```
phiSgr, line 496: what if the test script is not sending fast enough and the framework does that in two batches?

apps/framework-cli/src/cli/display.rs
```diff
@@ -562,3 +575,276 @@ mod tests {
...
+
+    // impl MessageComponent {
+    //     pub fn new(styled_text: StyledText, message: &str) -> Self {
+    //         Self {
+    //             styled_text,
+    //             message: message.to_string(),
+    //             displayed: false,
+    //         }
+    //     }
+    // }
```
phiSgr: delete?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1421,7 +1423,7 @@ impl Webserver {
             show_message!(
                 MessageType::Highlight,
                 Message {
-                    action: "Next Steps".to_string(),
+                    action: "Next Steps  ".to_string(),
```
phiSgr, line 1428: are we removing the auto-padding?

apps/framework-cli-e2e/test/templates.test.ts
callicles: Good call - I will change that

apps/framework-cli/src/cli/display.rs
```diff
@@ -55,13 +58,11 @@ use crate::framework::core::{
...
-/// The CommandTerminal wraps a console::Term for stdout handling and tracks the
+/// The CommandTerminal wraps stdout handling and tracks the
 /// number of messages written for potential future needs (such as clearing or
 /// updating specific messages).
 #[derive(Debug, Clone)]
 pub struct CommandTerminal {
-    /// The underlying terminal output handler
-    pub term: console::Term,
     /// Counter tracking the number of messages written to the terminal
     pub counter: usize,
```
phiSgr: i don't think we ever read this variable? maybe delete the whole struct?

apps/framework-cli/src/utilities/process_output.rs
```diff
@@ -0,0 +1,105 @@
...
+
+impl ProcessOutputProxy {
+    /// Create a new output proxy that forwards subprocess output to logs
+    pub fn new(stdout: ChildStdout, stderr: ChildStderr, label: &str) -> Self {
+        let label_stdout = format!("[{label}]");
+        let label_stderr = label_stdout.clone();
+
+        let stdout_task = tokio::spawn(async move {
+            let mut reader = BufReader::new(stdout).lines();
+            while let Ok(Some(line)) = reader.next_line().await {
```
phiSgr: maybe handle the `Err` case as well with some extra logging?

apps/framework-cli-e2e/test/templates.test.ts
callicles: I actually changed the waitForDbWrite to accumulate all the count it sees. That way we can actually count the total number of records landed in the db

apps/framework-cli/src/cli/local_webserver.rs
callicles: No - it is in there, but it is auto padding to the left. This is a weird case where we actually want the action field to look more centered

apps/framework-cli/src/cli/display.rs
callicles: Yep - good catch - that cleaned up a bunch of things

callicles:
@cursoragent can you review this PR?

callicles:
bugbot run

callicles:
bugbot run

PR #2563, opened 2025-07-15
Version paths for apis [ENG-22]

camelCasedAditya:


apps/framework-cli-e2e/test/templates.test.ts
```diff
@@ -604,7 +604,7 @@ describe("Moose Templates", () => {
 
     it("should successfully ingest data and verify through consumption API", async function () {
       const eventId = randomUUID();
-      const response = await fetch(`${TEST_CONFIG.server.url}/ingest/foo`, {
```
callicles: If the change broke this

I believe we had it before as URLs / Path are case insensitve. That's more user friendly

apps/framework-cli/src/cli/routines/ls.rs
```diff
@@ -450,31 +451,60 @@ fn to_info(endpoint: &ApiEndpoint) -> Either<IngestionApiInfo, ConsumptionApiInf
...
             data_model: _,
-        } => Either::Left(IngestionApiInfo {
-            name: endpoint.name.clone(),
-            destination: target_topic_id.clone(),
-        }),
+        } => {
+            // For ingestion APIs, show the user-facing URL format instead of internal storage format
+            // Convert internal format "ingest/Foo/1" to user-facing format "ingest/v1/Foo"
+            let user_facing_path = if let Some(version) = &endpoint.version {
+                format!("ingest/v{}/{}", version.as_str(), endpoint.name)
```
callicles: I understand the ticket says `/ingest/version/name`. But the hierachy is actually `/ingest/name/version` the reason being the `version` is an attribute of the Ingest API endpoint that the user declares. Not the other way around where the user declares a version and attaches paths to it.

apps/framework-cli/src/cli/routines/ls.rs
```diff
@@ -450,31 +451,60 @@ fn to_info(endpoint: &ApiEndpoint) -> Either<IngestionApiInfo, ConsumptionApiInf
...
-            name: endpoint.name.clone(),
-            params: query_params
-                .iter()
-                .map(|param| param.name.clone())
-                .collect(),
-            path: format!("consumption/{}", endpoint.name),
-        }),
+        } => {
+            // Check if the endpoint name already contains a version prefix (like "v1/bar")
+            let versioned_name = if endpoint.name.contains('/') {
```
callicles: This test is not doing what it says. ie checking that the endpoint name contains "/" doesn't mean it contains a version number

apps/framework-cli/src/cli/routines/ls.rs
callicles: Also we have had the version as a suffix all throughout the code so far

apps/framework-cli/src/cli/routines/ls.rs
```diff
@@ -450,31 +451,60 @@ fn to_info(endpoint: &ApiEndpoint) -> Either<IngestionApiInfo, ConsumptionApiInf
...
+                // Name already contains version prefix (e.g., "v1/bar" from Python)
+                endpoint.name.clone()
+            } else if let Some(version) = &endpoint.version {
+                // Name doesn't contain version, add it (e.g., "bar" -> "v1/bar" from TypeScript)
+                format!("v{}/{}", version.as_str(), endpoint.name)
```
callicles: we don't want to add the `v` I believe people don't necessarily expect it

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -243,6 +274,20 @@ async fn get_consumption_api_res(
         headers.insert(key, value.clone());
     }
 
+    // Add version information as a custom header if available
+    if let Some(version) = &version_segment {
```
callicles: In general it would be a better design if the version was passed in through as an argument to the method so that we did not have to parse it from a string.

packages/py-moose-lib/moose_lib/dmv2/consumption.py
```diff
@@ -88,13 +88,50 @@ def curried_constructor(*args, **kwargs):
...
         self.query_function = query_function
-        self.metadata = config.metadata
-        _egress_apis[name] = self
+        
+        # Initialize metadata object if not present
+        self.metadata = getattr(config, 'metadata', {}) or {}
+        
+        # Add source file information for debugging
+        import inspect
+        frame = inspect.currentframe()
```
callicles: How is this used?

packages/ts-moose-lib/src/consumption-apis/runner.ts
```diff
@@ -132,78 +152,122 @@ const apiHandler =
...
+
+          if (requestVersion) {
+            const versionedApiKey = `v${requestVersion}/${endpointName}`;
+            userFuncModule = egressApis.get(versionedApiKey);
+          } else {
+            userFuncModule = egressApis.get(endpointName);
+
+            if (!userFuncModule) {
+              const availableVersions: string[] = [];
+              for (const apiKey of egressApis.keys()) {
```
DatGuyJonathan: We might not want to call this `apiKey`. It sounds like something for auth but it looks like just the api with possible version.

apps/framework-cli/src/cli/routines/ls.rs
```diff
@@ -450,31 +451,65 @@ fn to_info(endpoint: &ApiEndpoint) -> Either<IngestionApiInfo, ConsumptionApiInf
...
 
 impl ResourceInfo for Vec<IngestionApiInfo> {
     fn show(&self) {
         show_table(
             "Ingestion APIs".to_string(),
-            vec!["name".to_string(), "destination".to_string()],
+            vec![
+                "name".to_string(),
+                "destination".to_string(),
+                "path".to_string(),
```
DatGuyJonathan, line 509: This reminded me, how do these versioned paths look in openapi? When moose starts, it creates an `./moose/openapi.yaml`. Then they could point an ide extension to that file or `http://localhost:5001/openapi.yaml` so they have some kind of UI.

camelCasedAditya: /ingest/Foo/2:
    post:
      summary: Ingress endpoint for Foo
      parameters: []
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Foo'
      responses:
        '200':
          description: Successful operation
  /ingest/Foo/1:
    post:
      summary: Ingress endpoint for Foo
      parameters: []
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Foo'
      responses:
        '200':
          description: Successful operation

camelCasedAditya:
```  
/ingest/Foo/2:
    post:
      summary: Ingress endpoint for Foo
      parameters: []
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Foo'
      responses:
        '200':
          description: Successful operation
  /ingest/Foo/1:
    post:
      summary: Ingress endpoint for Foo
      parameters: []
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Foo'
      responses:
        '200':
          description: Successful operation
```

apps/framework-cli/src/cli/routines/ls.rs
```diff
@@ -450,31 +451,65 @@ fn to_info(endpoint: &ApiEndpoint) -> Either<IngestionApiInfo, ConsumptionApiInf
...
             dead_letter_queue: _,
             data_model: _,
-        } => Either::Left(IngestionApiInfo {
-            name: endpoint.name.clone(),
-            destination: target_topic_id.clone(),
-        }),
+        } => {
+            // For ingestion APIs, show the user-facing URL format instead of internal storage format
+            let user_facing_path = if let Some(version) = &endpoint.version {
+                format!("ingest/{}/{}", endpoint.name, version.as_str())
```
callicles, line 457: Will this show the right path? if you have nested paths?

I think that's a PR you did right?

or will that be included in the `name`?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -198,27 +205,55 @@ async fn get_consumption_api_res(
...
-            .unwrap_or(req.uri().path());
+        let consumption_apis = config.consumption_apis.read().await;
 
-        if !consumption_apis.contains(consumption_name) {
-            if !is_prod {
+        // Generate consumption API key consistently
+        let consumption_name = if let (Some(version_str), Some(endpoint_str)) =
+            (&request_version, &request_endpoint)
+        {
+            format!("v{version_str}/{endpoint_str}")
```
callicles, line 246: you did not invest here?

callicles, line 246: That seems to be in the wrong order

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -246,6 +281,17 @@ async fn get_consumption_api_res(
     for (key, value) in req.headers() {
         headers.insert(key, value.clone());
     }
+    if let Some(version_str) = &request_version {
+        let version_number = if let Some(stripped) = version_str.strip_prefix('v') {
```
callicles, line 285: We should not have `v` in there that's pretty error prone

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -969,57 +1015,86 @@ async fn router(
...
-                // First find matching routes, then get latest version
-                for (path, meta) in route_table_read.iter() {
-                    let path_str = path.to_str().unwrap();
-                    if path_str.starts_with(base_path) {
-                        if let Some(version) = &meta.version {
-                            if latest_version.is_none() || version > latest_version.unwrap() {
-                                latest_version = Some(version);
+                // Case-insensitive route lookup
+                let route_str = route.to_str().unwrap().to_lowercase();
+                let matching_route = route_table_read
```
callicles, line 1041: This is ingest and performance sensitive code.
Here you are doing a loop over all the routes and then doing the same loop in the else statement. 

The previous version was handling versions it looks like and doing only one loop. 

Why did we change the logic here?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1313,10 +1407,22 @@ impl Webserver {
...
                             APIType::EGRESS { .. } => {
-                                consumption_apis
-                                    .write()
-                                    .await
-                                    .insert(api_endpoint.path.to_string_lossy().to_string());
+                                let mut apis_to_insert = Vec::new();
+
+                                if let Some(version) = &api_endpoint.version {
+                                    apis_to_insert.push(format!(
+                                        "v{}/{}",
```
callicles, line 1414: let's remove the `v`

callicles: And looks like the wrong order

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1327,10 +1433,24 @@ impl Webserver {
...
-                                consumption_apis
-                                    .write()
-                                    .await
-                                    .remove(&api_endpoint.path.to_string_lossy().to_string());
+                                let mut apis_to_remove = Vec::new();
+
+                                if let Some(version) = &api_endpoint.version {
+                                    // Remove versioned path
+                                    apis_to_remove.push(format!(
+                                        "v{}/{}",
```
callicles, line 1441: Same here

apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
```diff
@@ -104,31 +103,69 @@ def log_request(self, code = "-", size = "-"):
                               str(size)))
         def do_GET(self):
             parsed_path = urlparse(self.path)
-            module_name = parsed_path.path.lstrip('/')
+            path_parts = parsed_path.path.lstrip('/').split('/')
+            
+            module_name = path_parts[0]
+            version_from_path = None
+            if len(path_parts) >= 2:
+                version_from_path = path_parts[1]
```
callicles, line 111: the version should now be the last part

apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
```diff
@@ -104,31 +103,69 @@ def log_request(self, code = "-", size = "-"):
...
+            
+            version_from_header = self.headers.get('x-moose-api-version')
+            
+            request_version = version_from_path or version_from_header
+            
+            versioned_module_name = None
+            if request_version:
+                versioned_module_name = f"v{request_version}/{module_name}"
```
callicles, line 119: other way around

apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
```diff
@@ -104,31 +103,69 @@ def log_request(self, code = "-", size = "-"):
...
+            try:
+                if is_dmv2_enabled():
+                    api_to_use = None
+                    if request_version:
+                        api_to_use = get_consumption_api(module_name, request_version)
+                    else:
+                        api_to_use = get_consumption_api(module_name)
+                        
+                        if not api_to_use:
+                            from ._registry import _egress_apis
```
callicles, line 146: Imports in the middle of a function that's performance sensitive is a bad idea

apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
```diff
@@ -104,31 +103,69 @@ def log_request(self, code = "-", size = "-"):
...
+            if request_version:
+                versioned_module_name = f"v{request_version}/{module_name}"
+            
+            jwt_payload = None
+            auth_header = self.headers.get('Authorization')
+            if auth_header:
+                try:
+                    bearer_token = auth_header.split(' ')[1] if auth_header.startswith('Bearer ') else None
+                    if bearer_token:
+                        jwt_payload = jwt.decode(bearer_token, options={"verify_signature": False})
```
callicles, line 127: why not using the verify_jwt method already there?

apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
```diff
@@ -148,35 +184,69 @@ def do_GET(self):
...
 
-                    args = [moose_client, query_params]
-                    if jwt_payload is not None:
-                        args.append(jwt_payload)
-                    response = module.run(*args)
+                self.send_response(200)
+                self.send_header('Content-type', 'application/json')
+                self.send_header('Access-Control-Allow-Origin', '*')
+                self.send_header('Access-Control-Allow-Methods', 'GET, POST')
+                self.send_header('Access-Control-Allow-Headers', 'Authorization, Content-Type, baggage, sentry-trace, traceparent, tracestate')
```
callicles, line 229: Why all those changes?

apps/framework-cli/src/framework/python/wrappers/consumption_runner.py
```diff
@@ -148,35 +184,69 @@ def do_GET(self):
...
                 self.end_headers()
-                self.wfile.write(str(e).encode())
+                error_response = json.dumps({"error": str(e)})
+                self.wfile.write(error_response.encode())
+
+        def do_OPTIONS(self):
+            self.send_response(200)
+            self.send_header('Access-Control-Allow-Origin', '*')
+            self.send_header('Access-Control-Allow-Methods', 'GET, POST')
+            self.send_header('Access-Control-Allow-Headers', 'Authorization, Content-Type, baggage, sentry-trace, traceparent, tracestate')
```
callicles, line 248: Same here

packages/py-moose-lib/moose_lib/dmv2/consumption.py
```diff
@@ -18,6 +18,16 @@
 _global_base_url: Optional[str] = None
 
 
+def _generate_api_key(name: str, version: Optional[str] = None) -> str:
+    """   
+    Returns:
+        The API key string in the format "v{version}/{name}" or just "{name}" if no version.
+    """
+    if version:
+        return f"v{version}/{name}"
```
callicles, line 27: other way around and without the v

packages/py-moose-lib/moose_lib/dmv2/ingest_api.py
```diff
@@ -12,6 +12,17 @@
...
 from ._registry import _ingest_apis
 
+
+def _generate_api_key(name: str, version: Optional[str] = None) -> str:
+    """
+    Returns:
+        The API key string in the format "v{version}/{name}" or just "{name}" if no version.
+    """
+    if version:
+        return f"v{version}/{name}"
```
callicles, line 22: same here

PR #2564, opened 2025-07-15
restart child process after it exits

phiSgr:


apps/framework-cli/src/utilities/system.rs
```diff
@@ -49,3 +55,71 @@ pub async fn kill_child(child: &Child) -> Result<(), KillProcessError> {
...
+                                }
+                                Err(e) => {
+                                    error!("Error waiting for process: {:?}", e);
+                                }
+                            }
+                            'restart: loop {
+                                debug!("Delaying 1 second before restarting");
+                                sleep(Duration::from_secs(1)).await;
+
+                                info!("Attempting to restart process...");
```
callicles: Could have a more precise logs? like this is the process that gets restarted. for each one of those logs

apps/framework-cli/src/utilities/system.rs
```diff
@@ -49,3 +55,71 @@ pub async fn kill_child(child: &Child) -> Result<(), KillProcessError> {
...
+                                        warn!("Process exited with non-zero status: {:?}", exit_status);
+                                    }
+                                }
+                                Err(e) => {
+                                    error!("Error waiting for process: {:?}", e);
+                                }
+                            }
+                            'restart: loop {
+                                debug!("Delaying 1 second before restarting");
+                                sleep(Duration::from_secs(1)).await;
```
callicles: Could we have exponential backoff?

callicles (approved):

PR #2568, opened 2025-07-17
Safe Database Schema Migration and Change Management

phiSgr:


rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+- Backward compatibility with the current system
+- Handles edge cases through user customization
+
+#### Challenges
+- Requires user to anticipate changes done by moose
+- Potential for user errors in transformation logic
+
+## Open Questions
+
+1. **Update Atomicity**: How do we handle/stop concurrent writes when a migration is in progress?
```
callicles, line 181: I believe we play migrations there like other migration systems with the audit logs of applied migrations. 
We should play this out here.

rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+2. **Performance at Scale**: How do migrations perform on large tables?
+  Changing subfields in Nested/Tuple requires an `ALTER TABLE UPDATE` mutation.
+
+3. **Plan Data structure**:
```
callicles: Maybe a mix of those?
ie lists of SQL in JSON that way we can also use the plan for the topic changes

rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+### Overview
+
+The plan is created and approved before deployment,
+and is not committed to the code base nor included in the container image.
+
+1. **Planning**: System generates migration plan before deployment
+2. **Review**:
+   - Runs image with `moose-cli plan` entry point, OR
+   - Send the target infrastructure map to running instance for plan generation
+3. **Approval**: User reviews, edits, and approves the plan
```
callicles: Something missing here is edit no? how does the user correct the output

rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+- **Changing non-Moose managed database objects**: External tables, views, or procedures created outside of Moose
+
+## Solution: Migration Plan Review and Approval
+
+### Overview
+
+The plan is created and approved before deployment,
+and is not committed to the code base nor included in the container image.
+
+1. **Planning**: System generates migration plan before deployment
```
jmsuzuki: how is this run? on moose startup? on api or cli call?

rfd/0007/readme.mdx
jmsuzuki: wouldn't the user need to make the relevant code change so that the infrastructure map has their intended changes?

thought here is if I'm editing my infra map to apply it, then I'll have to continuously make those manual changes _unless_ they are part of our code.

rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+2. **Review**:
+   - Runs image with `moose-cli plan` entry point, OR
+   - Send the target infrastructure map to running instance for plan generation
+3. **Approval**: User reviews, edits, and approves the plan
+4. **Execution**: Approved plan gets injected into deployed container
```
jmsuzuki: in a local environment this isn't necessarily true... but yes.

rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+- No more push-to-main-and-deploy workflow
+unless the user turns on auto-approval - which is the same as the current solution.
+
+
+## Alternative Approaches Considered
+
+### SQL Generation Workflow
+
+#### Overview
+Generate SQL migration scripts as part of the development process rather than executing changes directly at runtime.
```
jmsuzuki, line 120: migrating and/or copying large datasets is a thing...
a simple query is not able to perform the entire migration on the Hosted Telemetry's ParsedLogs table.

rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+**Auto Approval Workflow**:
+- Runtime environment variable enables automatic plan approval and execution
+- Suitable for development and staging environments
+
+#### Benefits
+- Minimal time between plan generation and deployment.
+
+#### Downsides
+- No more push-to-main-and-deploy workflow
+unless the user turns on auto-approval - which is the same as the current solution.
```
jmsuzuki, line 94: it might be nice for us to have multiple modes here for auto-approvals.
1. auto-approve if nothing is deleted
2. auto-approve no matter what.

rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+### Git History Integration
+
+#### Overview
+Embed git history into Docker images to maintain change lineage and enable safe redeployments.
+
+#### Process
+1. **Change Tracking**: Include git commit history in container images
+2. **State Consistency**: Use git history to determine what changes have been applied
+3. **Safe Redeployment**: Prevent dropping changes that were applied in previous versions
```
jmsuzuki, line 148: only if we allow migration plans to be edited by a person before applying.
enforcing that users update their code to create an updated infrastructure plan makes redeployments safe-ish.

I think we would still need to worry about tables being dropped and recreated though.

rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+#### Benefits
+- Solves redeployment safety issues
+- Maintains change history
+- Enables rollback tracking
+
+#### Challenges
+- Container size increase
```
jmsuzuki, line 156: I don't see this as being an issue...

rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+#### Challenges
+- Requires user to anticipate changes done by moose
+- Potential for user errors in transformation logic
+
+## Open Questions
+
+1. **Update Atomicity**: How do we handle/stop concurrent writes when a migration is in progress?
+
+2. **Performance at Scale**: How do migrations perform on large tables?
+  Changing subfields in Nested/Tuple requires an `ALTER TABLE UPDATE` mutation.
```
jmsuzuki, line 184: I actually voiced this concern above with real world experience attached.

rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+3. **Plan Data structure**:
+   How should a plan be represented?
+   It can be a custom data type, e.g. `{"type": "DropColumn", "column_name": "stuff"}` (or in YAML).
+   This is a high level representation but developers have to learn the schema.
+   The plan can also be in SQL statements, but it is harder for Moose to validate.
```
jmsuzuki: yaml and/or json is fine... could even be user defined.

I would use terraform and pulumi as examples here and show change diffs.

rfd/0007/readme.mdx
callicles: That's a good callout - I believe we should only allow changes to the plan that are isomorphic to the code changes.

ie you change a column - you edit the plan to be rename instead of delete and create that works.
But you can't edit the plan to add a new table that's not in code

rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+This RFD proposes a comprehensive approach to database schema migration and change management for Moose deployments
+that prioritizes safety, user control, and minimal operational overhead.
+The current solution lacks essential safeguards against unintended data loss and
+provides insufficient user control over schema evolution.
+
+The proposed solution implements a **Migration Plan Review and Approval** workflow where the system generates
+migration plans before deployment, allows user review and editing, and requires explicit approval before
+execution. This approach provides explicit user consent for destructive operations while maintaining
+flexibility for automated deployments in development environments.
```
oatsandsugar, line 17: is this a common pattern here? I know terraform has something like this, but anything similar in the data space?

rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+3. **State Inconsistency**:
+   The `integrate-changes` command modifies the infrastructure map,
+   but redeploying the same version means the stored infrastructure map in the image doesn't contain the new changes,
+   causing the system to drop tables as they're not in the desired state.
+
+4. **Async Cast Failures**:
+   `ALTER TABLE MODIFY COLUMN` operations don't change underlying data, and cast operations can fail asynchronously.
+
+### Goals
+
```
oatsandsugar, line 45: other goals I'd see are:
1. transparency -> the user knows what will happen
2. sane defaults -> we recommend a good course of action by default (perhaps that minimizes data loss?)

rfd/0007/readme.mdx
oatsandsugar: shouldn't we allow that change / option to be expressable in the code?

oatsandsugar: Where is the approval happening? I always imagined this happened in the PR, but I think this implies in the CLI?

rfd/0007/readme.mdx
```diff
@@ -0,0 +1,157 @@
...
+### Overview
+
+The plan is created and approved before deployment,
+and is not committed to the code base nor included in the container image.
+
+1. **Planning**: System generates migration plan before deployment
+2. **Review**:
+   - Runs image with `moose-cli plan` entry point, OR
+   - Send the target infrastructure map to running instance for plan generation
```
DatGuyJonathan: How do we do this? Trying to understand if there's a way that can be fudged with by a bad actor along the way.

rfd/0007/readme.mdx
phiSgr: i think "full control" and "explicit" mostly overlap with "transparency"
"sane defaults" is roughly "Minimize work required for most common schema change scenarios"

worth wordsmithing though

rfd/0007/readme.mdx
phiSgr: Not PR, the idea is the plan is transient and not committed to the repo.

This can happen in Boreal UI, or the user's custom infra code.

rfd/0007/readme.mdx
phiSgr: Right, the main point is to put the plan (say, as a file) `moose prod` can access.
I wrote "deployed container" as the most likely deployment setup.

rfd/0007/readme.mdx
callicles: We have been going back and forth on it. That may be too much for MVP. ie keeping the dynamic aspect of the plan generation and allowing static diff editing in git commited code. <br><br>What we are thinking here is that the edits to the plan will be saved in an audit trail. but not in the repo code.

callicles (approved):

PR #2572, opened 2025-07-21
fix: no case mapping for columns in code generation

phiSgr:


apps/framework-cli/src/framework/typescript/generate.rs
```diff
@@ -340,7 +328,7 @@ mod tests {
             r#"export interface Address {
     street: string;
     city: string;
-    zipCode: string | undefined;
+    zip_code: string | undefined;
```
callicles, line 331: Why is the python convention in the ts file and vice versa?

phiSgr: to make sure the column name is not adapted to the languages' respective convention

callicles (approved):

PR #2576, opened 2025-07-23
Enable externally managed resources in moose

phiSgr:


apps/framework-cli/src/framework/core/infrastructure/topic.rs
```diff
@@ -47,6 +51,7 @@ impl Topic {
                 primitive_type: PrimitiveTypes::DataModel,
             },
             metadata: None,
+            life_cycle: LifeCycle::FullyManaged,
```
callicles, line 53: Why do you use a default here and are not trying to propagate from user definitions?

This might be outdated code. if it is we should remove no?

apps/framework-cli/src/framework/core/infrastructure/topic.rs
```diff
@@ -73,6 +78,7 @@ impl Topic {
                 primitive_type: PrimitiveTypes::Function,
             },
             metadata: None,
+            life_cycle: LifeCycle::FullyManaged,
```
callicles: Same here? outdated code?

callicles (approved):
Small comments - otherwise looks good to me

apps/framework-cli/src/framework/core/infrastructure/topic.rs
phiSgr: dmv1 code has no user definition for this field

apps/framework-cli/src/framework/core/infrastructure/topic.rs
phiSgr: yap

PR #2577, opened 2025-07-23
Optionally Spin up Storage containers + Multi Moose features

georgevanderson:
- Optionally use storage features
- remove hardcoded container names
- redis port configurable

WIP Left:
Testing the trailing container names, fix the docker-compose.hbs

### DIFF
--- a/apps/framework-cli/src/cli/local_webserver.rs
+++ b/apps/framework-cli/src/cli/local_webserver.rs
@@ -423,15 +423,20 @@ async fn health_route(
         }
     };
 
-    // Check ClickHouse connectivity
-    let olap_client =
-        crate::infrastructure::olap::clickhouse::create_client(project.clickhouse_config.clone());
-    let clickhouse_healthy = match olap_client.client.query("SELECT 1").execute().await {
-        Ok(_) => true,
-        Err(e) => {
-            warn!("Health check: ClickHouse unavailable: {}", e);
-            false
+    // Check ClickHouse connectivity only if storage is enabled
+    let clickhouse_healthy = if project.features.storage {
+        let olap_client = crate::infrastructure::olap::clickhouse::create_client(
+            project.clickhouse_config.clone(),
+        );
+        match olap_client.client.query("SELECT 1").execute().await {
+            Ok(_) => true,
+            Err(e) => {
+                warn!("Health check: ClickHouse unavailable: {}", e);
+                false
+            }
         }
+    } else {
+        true // Skip ClickHouse health check when storage is disabled
     };
 
     // Prepare healthy and unhealthy lists
@@ -448,10 +453,12 @@ async fn health_route(
     } else {
         unhealthy.push("Redpanda")
     }
-    if clickhouse_healthy {
-        healthy.push("ClickHouse")
-    } else {
-        unhealthy.push("ClickHouse")
+    if project.features.storage {
+        if clickhouse_healthy {
+            healthy.push("ClickHouse")
+        } else {
+            unhealthy.push("ClickHouse")
+        }
     }
 
     // Create JSON response
@@ -485,12 +492,6 @@ async fn admin_reality_check_route(
         return e.to_response();
     }
 
-    // Create OLAP client and reality checker
-    let olap_client =
-        crate::infrastructure::olap::clickhouse::create_client(project.clickhouse_config.clone());
-    let reality_checker =
-        crate::framework::core::infra_reality_checker::InfraRealityChecker::new(olap_client);
-
     // Load infrastructure map from Redis
     let infra_map = match InfrastructureMap::load_from_redis(redis_client).await {
         Ok(Some(map)) => map,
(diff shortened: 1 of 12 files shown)


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -423,15 +423,20 @@ async fn health_route(
...
+        );
+        match olap_client.client.query("SELECT 1").execute().await {
+            Ok(_) => true,
+            Err(e) => {
+                warn!("Health check: ClickHouse unavailable: {}", e);
+                false
+            }
         }
+    } else {
+        true // Skip ClickHouse health check when storage is disabled
```
callicles: Will we show a healthy clickhouse in the logs nonetheless?

if so feels like we should not run the check altogether

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -448,10 +453,12 @@ async fn health_route(
     } else {
         unhealthy.push("Redpanda")
     }
-    if clickhouse_healthy {
-        healthy.push("ClickHouse")
-    } else {
-        unhealthy.push("ClickHouse")
+    if project.features.storage {
```
callicles: Feels weird to have this here and not have it above for streaming? which one is buggy?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -504,25 +505,42 @@ async fn admin_reality_check_route(
...
 
-    // Perform reality check
-    match reality_checker.check_reality(project, &infra_map).await {
-        Ok(discrepancies) => {
-            let response = serde_json::json!({
-                "status": "success",
-                "discrepancies": discrepancies
-            });
+    // Perform reality check only if storage is enabled
+    let discrepancies = if project.features.storage {
```
callicles: In this case shouldn't we return an error to the user? 

if check reality is only for tables (for now) then we should return something like you are using it but there is no tables. Not implemented for streams or something like that

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1945,6 +1963,14 @@ async fn admin_integrate_changes_route(
             }
         };
 
+    // Skip integration if storage is disabled
+    if !project.features.storage {
```
callicles: Yeah something like that.

I would move it as an early return after the error though, no need to get a proper json in if you can't do anything with it

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -677,24 +677,30 @@ impl InfrastructureMap {
         process_changes.append(&mut function_process_changes);
 
         // TODO Change this when we have multiple processes for blocks
-        process_changes.push(ProcessChange::OlapProcess(Change::<OlapProcess>::Added(
-            Box::new(OlapProcess {}),
-        )));
+        // Only add OLAP process if storage is enabled
+        if project.features.storage {
+            process_changes.push(ProcessChange::OlapProcess(Change::<OlapProcess>::Added(
```
callicles, line 682: I don't remember, is this still used? feels like that was blocks - which we removed

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -677,24 +677,30 @@ impl InfrastructureMap {
...
         )));
 
-        process_changes.push(ProcessChange::OrchestrationWorker(Change::<
-            OrchestrationWorker,
-        >::Added(
-            Box::new(OrchestrationWorker {
-                supported_language: project.language,
-            }),
-        )));
+        let mut orchestration_worker_process_changes: Vec<ProcessChange> = self
```
callicles: Why did you change this?

apps/framework-cli/src/infrastructure/olap.rs
```diff
@@ -58,6 +58,12 @@ pub async fn execute_changes(
     project: &Project,
     changes: &[OlapChange],
 ) -> Result<(), OlapChangesError> {
+    // Skip OLAP changes if storage is disabled
+    if !project.features.storage {
```
callicles: Feels like we should not generate the olaps diff in the plan all together? and if the list is empty here nothing is attempted?

apps/framework-cli/src/infrastructure/redis/redis_client.rs
```diff
@@ -134,6 +134,8 @@ pub struct RedisConfig {
     pub url: String,
```
callicles, line 135: Does the url specification supposed to have a port?

if so we should change to host + port?

feels contradictory to have port and url at the same time

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -423,15 +423,20 @@ async fn health_route(
...
-    // Check ClickHouse connectivity
-    let olap_client =
-        crate::infrastructure::olap::clickhouse::create_client(project.clickhouse_config.clone());
-    let clickhouse_healthy = match olap_client.client.query("SELECT 1").execute().await {
-        Ok(_) => true,
-        Err(e) => {
-            warn!("Health check: ClickHouse unavailable: {}", e);
-            false
+    // Check ClickHouse connectivity only if storage is enabled
+    let clickhouse_healthy = if project.features.storage {
```
jmsuzuki: I guess this is configureable now?
by default or atleast in our own dev projects, we should default this to false...
on going health checks would keep a CH service alive instead of allowing it to idle.

apps/framework-cli/src/utilities/prod-docker-compose.yml.hbs
```diff
@@ -1,186 +1,82 @@
...
+  CLICKHOUSE_PASSWORD=${CLICKHOUSE_PASSWORD:-pandapass} -
+  CLICKHOUSE_DEFAULT_ACCESS_MANAGEMENT=1 ports: -
+  "${CLICKHOUSE_HOST_PORT:-18123}:8123" - "9000:9000" deploy: resources:
+  reservations: cpus: '4.0' memory: 16G healthcheck: test: ["CMD-SHELL",
+  "clickhouse-client -q 'select version()'"] interval: 5s retries: 2
+  start_period: 20s timeout: 20s ulimits: nofile: soft: 20000 hard: 40000
+{{/if}}
+
+{{#if scripts_feature}}
+  postgresql: environment: POSTGRES_PASSWORD: ${TEMPORAL_DB_PASSWORD:-temporal}
```
callicles: Something feels weird and unintended here

callicles (changes requested):

apps/framework-cli/src/infrastructure/redis/redis_client.rs
```diff
@@ -162,6 +164,35 @@ impl RedisConfig {
...
+    /// This method constructs the Redis URL by checking if the configured URL already contains
+    /// a port. If it doesn't, it uses the port field to construct the URL.
+    ///
+    /// # Returns
+    ///
+    /// A string containing the effective Redis URL
+    pub fn effective_url(&self) -> String {
+        // If the URL already contains a port (indicated by presence of :port), use it as-is
+        if self.url.contains(":6379") || self.url.matches(':').count() >= 2 {
+            self.url.clone()
```
jmsuzuki: this must be a rust thing...

apps/framework-cli/src/utilities/constants.rs
```diff
@@ -38,8 +38,8 @@ pub const VSCODE_DIR: &str = ".vscode";
 pub const SAMPLE_STREAMING_FUNCTION_SOURCE: &str = "Foo";
 pub const SAMPLE_STREAMING_FUNCTION_DEST: &str = "Bar";
 
-pub const CLICKHOUSE_CONTAINER_NAME: &str = "clickhousedb-1";
-pub const REDPANDA_CONTAINER_NAME: &str = "redpanda-1";
```
jmsuzuki, line 42: I see that you added these 1's to the container names above...
I guess this is to support many CH and redpanda containers in a dev environemnt?

apps/framework-cli/src/utilities/prod-docker-compose.yml.hbs
```diff
@@ -1,186 +1,82 @@
...
-      - TEMPORAL_ADDRESS=temporal:${TEMPORAL_PORT:-7233}
-      - TEMPORAL_CORS_ORIGINS=${TEMPORAL_UI_CORS_ORIGINS:-http://localhost:3000}
-    image: temporalio/ui:${TEMPORAL_UI_VERSION}
-    networks:
-      - temporal-network
-    ports:
-      - ${TEMPORAL_UI_PORT:-8080}:8080
+{{#if streaming_engine}}
+  redpanda: image: docker.redpanda.com/redpandadata/redpanda:v24.3.11 ports: -
+  "9092:9092" - "19092:19092" - "9644:9644" volumes: -
```
jmsuzuki: while we're here, why not make these configurable?

tg339 (approved):

apps/framework-cli/src/utilities/prod-docker-compose.yml.hbs
georgevanderson: next round of modularization, this is minimal changes to get area-code working

callicles (approved):

PR #2583, opened 2025-07-25
fix temporal localhost checks

DatGuyJonathan:


callicles:
Feels like the logic is weird?

Shouldn't it be use the default namespace if none provided, otherwise use the one provided?

DatGuyJonathan:
Hm, I don't think there's a temporal namespace config in moose. When it's temporal cloud, the namespace is derived from their hostname. When it's local, it's "default" everywhere. Lemme see about adding that config.

phiSgr (approved):
might be worth adding a few lines of comments on the assumptions (i think they are valid) behind the code

callicles (approved):
Love it! 

Thank you

phiSgr (approved):

PR #2590, opened 2025-07-29
fix: olap diff

callicles:
This pull request introduces a ClickHouse-specific table diffing strategy to handle schema changes and updates the existing infrastructure to use this strategy. The changes include implementing the new strategy, integrating it into the infrastructure map diffing process, and refactoring the handling of table updates to rely on database-specific strategies. Below are the most important changes grouped by theme:

### ClickHouse-specific diffing strategy:
* Added a new module `diff_strategy` in `apps/framework-cli/src/infrastructure/olap/clickhouse/` that implements `ClickHouseTableDiffStrategy`. This strategy handles ClickHouse-specific limitations, such as requiring drop-and-create operations for changes to the `ORDER BY` clause, primary key structure, or deduplication settings.

### Integration of ClickHouse-specific strategy:
* Updated `admin_plan_route` in `apps/framework-cli/src/cli/local_webserver.rs` to use the `ClickHouseTableDiffStrategy` for diffing infrastructure maps.
* Updated `plan_changes` in `apps/framework-cli/src/framework/core/plan.rs` to use the `ClickHouseTableDiffStrategy` for diffing reconciled infrastructure maps.
* Updated `remote_refresh` in `apps/framework-cli/src/cli/routines.rs` to switch from a generic table diff method to a simpler `simple_table_diff` method.

### Refactoring of table update handling:
* Removed the `handle_non_mutable_change_operation` function in `apps/framework-cli/src/infrastructure/olap/ddl_ordering.rs` and replaced `handle_table_update` with `handle_table_column_updates`, which now assumes that database-specific transformations are handled during the diff phase. [[1]](diffhunk://#diff-bfe1723e1a57eea016d6865666bea7751a5a98d1ed3b4694274b7012ce42f28bL403-L438) [[2]](diffhunk://#diff-bfe1723e1a57eea016d6865666bea7751a5a98d1ed3b4694274b7012ce42f28bL662-R650)
* Updated tests in `ddl_ordering.rs` to reflect the new approach, focusing on column-level changes rather than handling `ORDER BY` or primary key changes directly. [[1]](diffhunk://#diff-bfe1723e1a57eea016d6865666bea7751a5a98d1ed3b4694274b7012ce42f28bL2262-R2255) [[2]](diffhunk://#diff-bfe1723e1a57eea016d6865666bea7751a5a98d1ed3b4694274b7012ce42f28bR2324-R2395)

### Code organization:
* Added the new `diff_strategy` module to `apps/framework-cli/src/infrastructure/olap/clickhouse.rs`.
* Updated imports in `apps/framework-cli/src/framework/core/plan.rs` to include `ClickHouseTableDiffStrategy`.

### DIFF
--- a/apps/framework-cli/src/cli/local_webserver.rs
+++ b/apps/framework-cli/src/cli/local_webserver.rs
@@ -2106,7 +2106,11 @@ async fn admin_plan_route(
     };
 
     // Calculate the changes between the submitted infrastructure map and the current one
-    let changes = current_infra_map.diff(&plan_request.infra_map);
+    // Use ClickHouse-specific strategy for table diffing
+    let clickhouse_strategy =
+        crate::infrastructure::olap::clickhouse::diff_strategy::ClickHouseTableDiffStrategy;
+    let changes =
+        current_infra_map.diff_with_table_strategy(&plan_request.infra_map, &clickhouse_strategy);
 
     // Prepare the response
     let response = PlanResponse {
--- a/apps/framework-cli/src/infrastructure/olap/clickhouse/diff_strategy.rs
+++ b/apps/framework-cli/src/infrastructure/olap/clickhouse/diff_strategy.rs
@@ -169,8 +169,8 @@ mod tests {
         let after = create_test_table("test", vec!["id".to_string()], true);
 
         let order_by_change = OrderByChange {
-            before: vec![],
-            after: vec![],
+            before: before.order_by.clone(),
+            after: after.order_by.clone(),
         };
 
         let changes = strategy.diff_table_update(&before, &after, vec![], order_by_change);
@@ -207,8 +207,8 @@ mod tests {
         }];
 
         let order_by_change = OrderByChange {
-            before: vec![],
-            after: vec![],
+            before: before.order_by.clone(),
+            after: after.order_by.clone(),
         };
 
         let changes = strategy.diff_table_update(&before, &after, column_changes, order_by_change);
--- a/apps/framework-cli/src/infrastructure/olap/clickhouse/diff_strategy.rs
+++ b/apps/framework-cli/src/infrastructure/olap/clickhouse/diff_strategy.rs
@@ -21,6 +21,10 @@ use crate::framework::core::infrastructure_map::{
 pub struct ClickHouseTableDiffStrategy;
 
 impl TableDiffStrategy for ClickHouseTableDiffStrategy {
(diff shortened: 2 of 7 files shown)


callicles:
cursor run

callicles:
cursor run

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -2102,7 +2102,11 @@ async fn admin_plan_route(
     };
 
     // Calculate the changes between the submitted infrastructure map and the current one
-    let changes = current_infra_map.diff(&plan_request.infra_map);
+    // Use ClickHouse-specific strategy for table diffing
+    let clickhouse_strategy =
+        crate::infrastructure::olap::clickhouse::diff_strategy::ClickHouseTableDiffStrategy;
```
DatGuyJonathan, line 2111: In the future, would this be a selector instead (i.e. check what the db is, then return the appropriate one)?

apps/framework-cli/src/infrastructure/olap/clickhouse/diff_strategy.rs
```diff
@@ -0,0 +1,323 @@
...
+    /// or require a drop+create operation.
+    fn diff_table_update(
+        &self,
+        before: &Table,
+        after: &Table,
+        column_changes: Vec<ColumnChange>,
+        order_by_change: OrderByChange,
+    ) -> Vec<OlapChange> {
+        // Check if ORDER BY has changed
+        let order_by_changed = order_by_change.before != order_by_change.after;
```
DatGuyJonathan, line 36: I wonder what happens here if they accidentally add whitespaces when the fields are actually the same. I think I would prefer the plan to show that it changed, then I fix it myself instead of moose trimming whitespaces, so the code that's checked in is exact.

DatGuyJonathan (approved):

apps/framework-cli/src/cli/local_webserver.rs
callicles: Yep - that's the idea

apps/framework-cli/src/infrastructure/olap/clickhouse/diff_strategy.rs
callicles: Hum - yeah I guess that's a matter of preference. 
I can leave it as is for now and we see if we get feedback on that

PR #2591, opened 2025-07-29
Versioned Consumption APIs

camelCasedAditya:


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -217,25 +217,50 @@ async fn get_consumption_api_res(
...
-                    consumption_apis
-                        .iter()
-                        .map(|p| p.as_str())
-                        .collect::<Vec<&str>>()
-                        .join(", ")
-                );
-                let _ = execute!(std::io::stdout(), Print(msg + "\n"));
+        // Check for exact match first
+        if consumption_apis.contains(consumption_name) {
+            // Exact match found, proceed with the request
```
callicles: if I read this right - as is nothing happens no?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -217,25 +217,50 @@ async fn get_consumption_api_res(
...
+            // Check if this looks like a versioned request (contains a slash)
+            if consumption_name.contains('/') {
+                // Split the path to get base name and potential version
+                let parts: Vec<&str> = consumption_name.split('/').collect();
+                if parts.len() >= 2 {
+                    let base_name = parts[0];
+
+                    // Look for any API that starts with the base name followed by a slash
+                    for api in consumption_apis.iter() {
+                        if api.starts_with(&format!("{base_name}/")) {
```
callicles: Is there a case where the api could be nested?

Like `ingest/test/test/2`

callicles (commented):
Thank you for doing the work of splitting this out - much easier to review!

Some small comments to explore or adjust

apps/framework-cli/src/cli/local_webserver.rs
camelCasedAditya: Currently it does not work with nested api versions. Will add this functionality

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1062,7 +1086,9 @@ async fn router(
         (_, &hyper::Method::POST, ["admin", "plan"]) => {
             admin_plan_route(req, &project.authentication.admin_api_key, &redis_client).await
         }
-        (_, &hyper::Method::GET, ["consumption", _rt]) => {
+        (_, &hyper::Method::GET, route_segments)
+            if route_segments.len() >= 2 && route_segments[0] == "consumption" =>
```
callicles, line 1068: Not a blocker - not sure why you changed this

I believe this is strictly equivalent to what it was before but with more lines?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -211,31 +211,55 @@ async fn get_consumption_api_res(
...
-                    "Consumption API {} not found. Available consumption paths: {}",
-                    consumption_name,
-                    consumption_apis
-                        .iter()
-                        .map(|p| p.as_str())
-                        .collect::<Vec<&str>>()
-                        .join(", ")
-                );
-                let _ = execute!(std::io::stdout(), Print(msg + "\n"));
+            // Try to find a versioned match
```
callicles: Did you remove completely the logic for handling versions?

camelCasedAditya: No. It is just in the consumption runners for typescript and python

apps/framework-cli/src/cli/local_webserver.rs
camelCasedAditya: The original does not allow for nested version paths

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -211,31 +211,55 @@ async fn get_consumption_api_res(
...
+
+            // Check if this looks like a versioned request (contains a slash)
+            if consumption_name.contains('/') {
+                // Split the path to get base name and potential version
+                let parts: Vec<&str> = consumption_name.split('/').collect();
+                if parts.len() >= 2 {
+                    let base_name = parts[0];
+
+                    // Look for any API that starts with the base name followed by a slash
+                    for api in consumption_apis.iter() {
```
callicles: To be honest - I am not in love with this `for loop` but I am not sure I have a much better solution from an implementation perspective.

For every request, we will have to find the consumption API that fits. If it is the latest in a 1000 endpoint list then it will have to go to the end. 

In high load scenarios that will eat a bunch of time. The previous lookup was O(1) so really fast.

I am tempted to remove all this code that's in the end just there to return an early 404 and let the node / python service return the 404.

What do you think?

callicles (commented):
You should add an example of the consumption API working with and without the version to the end to end templates for python and ts.

That way we show that this works through end to end tests and that we did not backward compatible break it

apps/framework-cli/src/cli/local_webserver.rs
camelCasedAditya: I agree. Just made the change. Let me know if it is sufficient

apps/framework-cli-e2e/test/templates.test.ts
```diff
@@ -671,6 +770,30 @@ describe("Moose Templates", () => {
         ],
       );
 
+      // Test versioned API (V1)
+      await utils.verifyVersionedConsumptionApi(
```
callicles, line 783: you are missing the not v1 test no?

and how are you going to tell if it is actually hitting v1 or not v1?

apps/framework-cli-e2e/test/templates.test.ts
```diff
@@ -358,6 +358,84 @@ const utils = {
...
+      // Check structure and value validation rather than exact matches
+      expect(json).to.be.an("array");
+      expect(json.length).to.be.at.least(1);
+
+      json.forEach((item: any, index: number) => {
+        const expected = expectedResponse[index] || expectedResponse[0];
+
+        // Verify structure - all expected keys exist
+        Object.keys(expected).forEach((key) => {
+          if (key === "metadata") {
```
callicles: This is hardcoded to what you expect - do you think there is a way to have this logic be generic? since you are passing what's expected as the input of the function

apps/framework-cli-e2e/test/templates.test.ts
camelCasedAditya: No I belive it is right above on line 760. Also currently I am checking if it hits V1 based on if it returns meta data as I set V1 to return metadata and the unversioned consumption API to not return metadata

callicles (approved):

PR #2593, opened 2025-07-30
add task oncancel

DatGuyJonathan:
ufa example:

Screenshot 2025-07-30 at 12 53 09 PM


packages/ts-moose-lib/src/scripts/activity.ts
```diff
@@ -104,11 +105,31 @@ export const activities = {
     task: Task<any, any>,
     inputData: any,
   ): Promise<any[]> {
+    // Get context for heartbeat (required for cancellation detection)
+    const context = Context.current();
+
+    // Periodic heartbeat. Important for cancellation detection.
+    let heartbeatInterval: NodeJS.Timeout | null = null;
```
callicles, line 117: I don't get why you need the heartbeat, can you explain?

packages/ts-moose-lib/src/scripts/activity.ts
```diff
@@ -119,7 +140,31 @@ export const activities = {
...
+          if (fullTask.config.onCancel) {
+            await fullTask.config.onCancel();
+          }
+          return [];
+        } else {
+          throw error;
+        }
+      } finally {
+        stopPeriodicHeartbeat();
+      }
```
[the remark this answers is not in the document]
DatGuyJonathan: Yes, hence the todo to refactor this to use something we can kill.

packages/ts-moose-lib/src/scripts/activity.ts
DatGuyJonathan: Good call, I added more comments & a link to their doc. Seems to imply that the cancellation signal is part of the heartbeat response from temporal server and that tells the `context.cancelled` to reject. The reject didn't work without the heartbeat.

DatGuyJonathan:
> Does python need anything?

Yep, python will need to be mirrored. Hopefully this unblocks ufa first.

PR #2595, opened 2025-07-31
Versioned ingest apis

camelCasedAditya:


apps/framework-cli/src/cli/routines/ls.rs
```diff
@@ -401,15 +401,21 @@ impl StreamInfo {
...
         topic_to_table_sync_processes: &HashMap<String, TopicToTableSyncProcess>,
+        infra_map: &InfrastructureMap,
     ) -> Self {
         let process = topic_to_table_sync_processes
             .values()
             .find(|p| p.source_topic_id == value.id());
 
         Self {
-            name: value.name,
+            name: value.id(),
```
callicles, line 411: Why did you change this?

packages/py-moose-lib/moose_lib/streaming/streaming_function_runner.py
```diff
@@ -160,7 +160,8 @@ def load_streaming_function_dmv2(function_file_dir: str, function_file_name: str
 
     # Find the stream that has a transformation matching our source/destination
     for source_py_stream_name, stream in get_streams().items():
-        if source_py_stream_name != source_topic.topic_name_to_stream_name():
+        # Compare the stream name (without version) with the topic's stream name
+        if stream.name != source_topic.topic_name_to_stream_name():
```
callicles: Why did you change this?

packages/py-moose-lib/moose_lib/internal.py
```diff
@@ -237,7 +244,7 @@ def _map_sql_resource_ref(r: Any) -> InfrastructureSignatureJson:
         if r.kind == "OlapTable":
             # Explicitly cast for type hint checking if needed, though Python is dynamic
             table = r # type: OlapTable
-            res_id = f"{table.name}_{table.config.version}" if table.config.version else table.name
+            res_id = f"{table.name}_{table.config.version.replace('.', '_')}" if table.config.version else table.name
```
callicles: you should be using the method you created above no?

packages/py-moose-lib/moose_lib/internal.py
```diff
@@ -299,7 +306,7 @@ def to_infra_map() -> dict:
         ]
 
         topics[name] = TopicConfig(
-            name=name,
+            name=stream.name,
```
callicles, line 312: Why all the variable changes here?

camelCasedAditya: The key is versioned but stream.name is not. So the registrations would be "Foo_1": {"name": "Foo", [REST OF STUFF]}

camelCasedAditya: This is also why there are similar changes for other registrations like for tables and ingest APIs

apps/framework-cli/src/cli/routines/ls.rs
camelCasedAditya: calling .id() includes the version suffix like "_1" when the streams are printed out. ".name" only contains the name the user specified, not the version attached to the name.

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1067,49 +1067,67 @@ async fn router(
...
+                        let path_str = path.to_str().unwrap();
+                        // Only consider versioned routes that match the pattern
                         if let Some(version) = &meta.version {
-                            if latest_version.is_none() || version > latest_version.unwrap() {
+                            // Check if this is a versioned route for the same base path
+                            // e.g., "ingest/Foo/1" for base path "ingest/Foo"
+                            if path_str.starts_with(&format!("{base_path}/"))
+                                && (latest_version.is_none() || version > latest_version.unwrap())
+                            {
                                 latest_version = Some(version);
```
callicles: Small but important optimization - when found we should break the loop. There is no need to continue. 
Maybe changing this code to use `find` instead would do.

packages/py-moose-lib/moose_lib/internal.py
```diff
@@ -285,7 +292,7 @@ def to_infra_map() -> dict:
         transformation_targets = [
             Target(
                 kind="stream",
-                name=dest_name,
+                name=transform.destination.generate_topic_id(),
```
callicles: Isn't this wrong?

ie the version for the target is already below and it should not be duplicated in the name

packages/py-moose-lib/moose_lib/internal.py
```diff
@@ -313,17 +320,23 @@ def to_infra_map() -> dict:
...
+        dead_letter_queue_id = api.config.dead_letter_queue.generate_topic_id() if api.config.dead_letter_queue else None
+
+        ingest_apis[versioned_key] = IngestApiConfig(
+            name=api.name,  # Use original name from the API object
             columns=_to_columns(api._t),
             version=api.config.version,
             write_to=Target(
                 kind="stream",
-                name=api.config.destination.name
+                name=api.config.destination.generate_topic_id(),
```
callicles: same here

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1067,39 +1067,58 @@ async fn router(
...
-                                latest_version = Some(version);
-                            }
-                        }
-                    }
-                }
 
-                match latest_version {
-                    // If latest version exists, use it
-                    Some(version) => {
+                let incoming_route_str = route.to_str().unwrap_or("");
```
callicles: is it correct to have the empty `.unwrap_or("")`?

it would not panic but would the rest of the logic work?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1067,39 +1067,58 @@ async fn router(
...
+                let incoming_lower = incoming_route_str.to_ascii_lowercase();
+
+                // 1) Try to find an exact (case-insensitive) unversioned match
+                if let Some((matched_path, _)) = route_table_read.iter().find(|(p, _)| {
+                    p.to_str()
```
callicles: Instead of doing all the iterations twice - we should only be doing them once.
it looks like you could put the else content of the find in here as well. so that you can find the route whether it has a version or not

apps/framework-cli/src/framework/core/partial_infrastructure_map.rs
```diff
@@ -180,16 +180,20 @@ struct PartialEgressApi {
...
 
 /// Specifies a transformation target for topic data.
 ///
 /// Used to define where transformed data should be written and optionally specify a version.
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct TransformationTarget {
     pub kind: WriteToKind,
     pub name: String,
     pub version: Option<String>,
```
callicles: What's this version vs the topic_version below?

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1067,50 +1067,63 @@ async fn router(
...
+                let mut versioned_count = 0;
+                let mut final_path: Option<PathBuf> = None;
+
+                for (path, metadata) in route_table_read.iter() {
+                    if let Some(path_str) = path.to_str() {
+                        if path_str.eq_ignore_ascii_case(incoming_route_str) {
+                            final_path = Some(path.clone());
+                            break;
+                        }
+                        if metadata.version.is_some()
```
callicles: Does the algorithm work in this case?

```
/ingest/test/hello/0.0
/ingest/test/something/0.0
```

You probably should see if you can add a bunch of unit test for the router function.

camelCasedAditya: It does work in this specific case. Also added other unit tests too

PR #2598, opened 2025-08-01
Support PNPM, copy lock file into packager, add baseURL modification

georgevanderson:


apps/framework-cli/src/cli/routines/build.rs
```diff
@@ -430,6 +449,7 @@ fn run_moose_check(package_dir: &PathBuf) -> Result<(), BuildError> {
         BuildError::MooseCheckFailed(format!("Failed to get current executable path: {e}"))
     })?);
 
+    println!("Running moose check in directory: {:?}", package_dir);
```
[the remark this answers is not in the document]
callicles: +1

apps/framework-cli/src/cli/routines/build.rs
```diff
@@ -544,3 +564,71 @@ fn create_archive(project: &Project, package_dir: &Path) -> Result<PathBuf, Buil
...
+///
+/// * `Result<(), BuildError>` - Returns Ok(()) on success
+///
+/// # Errors
+///
+/// Returns a `BuildError` if:
+/// - Reading the tsconfig.json file fails
+/// - Parsing the JSON fails
+/// - Writing the modified file fails
+fn modify_tsconfig_baseurl(tsconfig_path: &Path) -> Result<(), BuildError> {
```
callicles, line 620: Why do you need this? aren't we copying all the files in the packager during the build?

apps/framework-cli/src/utilities/package_managers.rs
```diff
@@ -148,6 +148,42 @@ pub fn link_sdk(
...
+/// # Returns
+///
+/// * `PackageManager` - The detected package manager
+pub fn detect_package_manager(project_dir: &PathBuf) -> PackageManager {
+    use crate::utilities::constants::{PACKAGE_LOCK_JSON, PNPM_LOCK};
+
+    debug!("Detecting package manager in directory: {:?}", project_dir);
+
+    // Check for pnpm-lock.yaml first (most specific)
+    if project_dir.join(PNPM_LOCK).exists() {
```
callicles: I believe you also need to look up the tree for `pnpm` the lock file might not be in the project directory but in one of the parent

apps/framework-cli/src/cli/routines/build.rs
georgevanderson: If the tsconfig is using relative path resolution, it won't work, because we copy the files into a subidrectory two levels below

ie
 "paths": {
      "@workspace/models": ["../../packages/models/src/index"],
      "@workspace/models/*": ["../../packages/models/src/*"],

callicles (approved):

PR #2603, opened 2025-08-03
mitchell/moose-build: monorepo support

jmsuzuki:
**chore: moose build supports mono repos.**

### DIFF
--- a/apps/framework-cli/src/cli/routines/docker_packager.rs
+++ b/apps/framework-cli/src/cli/routines/docker_packager.rs
@@ -10,8 +10,10 @@ use crate::utilities::docker::DockerClient;
 use crate::utilities::package_managers::get_lock_file_path;
 use crate::utilities::{constants, system};
 use crate::{cli::display::Message, project::Project};
-use log::{error, info};
+
+use log::{debug, error, info};
 use std::fs;
+use std::path::{Path, PathBuf};
 
 // Adapted from https://github.com/nodejs/docker-node/blob/2928850549388a57a33365302fc5ebac91f78ffe/20/bookworm-slim/Dockerfile
 // and removed yarn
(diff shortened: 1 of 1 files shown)


apps/framework-cli/src/cli/routines/docker_packager.rs
```diff
@@ -376,6 +440,91 @@ pub fn build_dockerfile(
         cli_version = "0.3.810";
     }
 
+    // Check if this is a monorepo build
+    let monorepo_info_path = internal_dir.join("packager/.monorepo-info");
```
phiSgr, line 439: do we have to document this somewhere

jmsuzuki: No, other than a code comment to help the next reader.

Yes, you're right! `.monorepo-info` is a temporary file generated by our `docker_packager` process. Let me explain:
Ran tool
Let me check what this file does:
Read file: apps/framework-cli/src/cli/routines/docker_packager.rs
Read file: apps/framework-cli/src/cli/routines/docker_packager.rs
Now let me check where it gets cleaned up:
Read file: apps/framework-cli/src/cli/routines/docker_packager.rs
Let me find the cleanup section:
Ran tool
Let me check the cleanup section more thoroughly:
Read file: apps/framework-cli/src/cli/routines/docker_packager.rs
Perfect! Now let me check if it exists in your area-code project:
```bash
ls -la .moose/packager/.monorepo-info 2>/dev/null || echo "File does not exist"
```
## Explanation of `.monorepo-info`

`.monorepo-info` is a **temporary file** created by our `docker_packager` process:

1. **Purpose**: It's used to pass information between the two phases of the Docker build process:
   - **Phase 1 (`create_dockerfile`)**: Detects if we're in a monorepo, generates the appropriate Dockerfile, and stores the workspace root path and relative project path in `.monorepo-info`
   - **Phase 2 (`build_dockerfile`)**: Reads `.monorepo-info` to determine if it needs to build from the monorepo root or use standard build process

2. **Contents**: The file contains two lines:
   - Line 1: Absolute path to the workspace root (e.g., `/Users/mitchell/repos/area-code`)
   - Line 2: Relative path from workspace root to the project (e.g., `ufa/services/analytical-moose-foobar`)

3. **Location**: `.moose/packager/.monorepo-info`

4. **Lifecycle**: 
   - Created during `create_dockerfile` if a monorepo is detected
   - Read during `build_dockerfile` to determine build context
   - **Cleaned up automatically** at the end of `build_dockerfile` (line 660)

The reason you don't see it in your area-code project is because it's properly cleaned up after each build! The Docker builds work perfectly - the file is just a temporary communication mechanism between the two build phases and is removed once the Docker image is successfully built.

apps/framework-cli/src/cli/routines/docker_packager.rs
```diff
@@ -436,8 +641,186 @@
...
+        if line.trim() == "packages:" {
+            in_packages = true;
+            continue;
+        }
+        if in_packages && line.starts_with("  - ") {
+            let pattern = line.trim_start_matches("  - ").trim_matches('"');
+            patterns.push(pattern.to_string());
+        }
+    }
+
```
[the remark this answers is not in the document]
jmsuzuki: fixed

PR #2613, opened 2025-08-05
components only

okane16:
Breaking PR into smaller chunks for easier review. this is just the components changes:
- github button
- localstorage language switcher
- some UI things needed for remaining content

### DIFF
--- a/apps/framework-docs/src/components/animate-ui/buttons/github-stars.tsx
+++ b/apps/framework-docs/src/components/animate-ui/buttons/github-stars.tsx
@@ -1,0 +1,254 @@
+"use client";
+
+import * as React from "react";
+import { Star } from "lucide-react";
+import {
+  motion,
+  AnimatePresence,
+  useMotionValue,
+  useSpring,
+  useInView,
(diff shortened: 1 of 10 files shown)


callicles (approved):

apps/framework-docs/src/components/animate-ui/buttons/github-stars.tsx
```diff
@@ -0,0 +1,254 @@
...
+  const [displayParticles, setDisplayParticles] = React.useState(false);
+  const [isLoading, setIsLoading] = React.useState(true);
+
+  const repoUrl = React.useMemo(
+    () => `https://github.com/${username}/${repo}`,
+    [username, repo],
+  );
+
+  React.useEffect(() => {
+    fetch(`https://api.github.com/repos/${username}/${repo}`)
```
DatGuyJonathan, line 78: I think I got rate limited at some point (because this is an unauthenticated request), so the button shows 0 stars.  Maybe we show no number if the api call fails?

Screenshot 2025-08-05 at 7 07 12 PM

PR #2620, opened 2025-08-06
Patch users `paths` file in compilerOptions

georgevanderson:
This is something I spent very long trying to avoid, but in area-code there is no configuration that either @groy-514 or I could use that doesnt rely on using `paths` in tsconfig. When we copy over and pnpm deploy, we lose these references. This only handles one specific case of a hard problem and what to do with more external dependencies

### DIFF
--- a/apps/framework-cli/src/cli/routines/docker_packager.rs
+++ b/apps/framework-cli/src/cli/routines/docker_packager.rs
@@ -12,8 +12,23 @@ use crate::utilities::{constants, system};
 use crate::{cli::display::Message, project::Project};
 
 use log::{debug, error, info};
+use serde_json;
 use std::fs;
 use std::path::{Path, PathBuf};
+
+#[derive(Debug, Clone)]
+struct PackageInfo {
+    name: String,
+    original_path: String,
+}
+
+#[derive(Debug)]
+struct PathMapping {
+    alias: String,
+    original_path: String,
+    package_name: String,
+    package_directory: String, // The directory name from the filesystem path
+}
 
 // Adapted from https://github.com/nodejs/docker-node/blob/2928850549388a57a33365302fc5ebac91f78ffe/20/bookworm-slim/Dockerfile
 // and removed yarn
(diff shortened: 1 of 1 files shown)


apps/framework-cli/src/cli/routines/docker_packager.rs
```diff
@@ -12,9 +12,62 @@ use crate::utilities::{constants, system};
...
 
+#[derive(Debug, Clone)]
+struct PackageInfo {
+    name: String,
+}
+
+type RoutineResult = Result<RoutineSuccess, RoutineFailure>;
+
+/// Helper function to safely create directories
+fn ensure_directory_exists(path: &Path) -> Result<(), std::io::Error> {
```
jmsuzuki, line 26: surprised you needed this.

apps/framework-cli/src/cli/routines/docker_packager.rs
```diff
@@ -785,6 +833,206 @@ fn generate_workspace_copy_commands(workspace_root: &Path, patterns: &[String])
...
+    if let Some(paths) = tsconfig
+        .get("compilerOptions")
+        .and_then(|co| co.get("paths"))
+        .and_then(|p| p.as_object())
+    {
+        for (alias, path_array) in paths {
+            if let Some(paths) = path_array.as_array() {
+                for path in paths {
+                    if let Some(path_str) = path.as_str() {
+                        if let Some((package_info, package_dir)) = read_package_from_typescript_path(
```
jmsuzuki, line 925: some docs around what we're digging in to here might help a future reader

apps/framework-cli/src/cli/routines/docker_packager.rs
```diff
@@ -136,10 +189,10 @@ EXPOSE 4000
...
 "#;
 
-pub fn create_dockerfile(
-    project: &Project,
-    docker_client: &DockerClient,
-) -> Result<RoutineSuccess, RoutineFailure> {
+/// Creates a Dockerfile for the given project, supporting both monorepo and standalone configurations.
+/// For TypeScript projects in monorepos, analyzes workspace structure and TypeScript path mappings
+/// to generate appropriate multi-stage Docker builds with proper dependency resolution.
+pub fn create_dockerfile(project: &Project, docker_client: &DockerClient) -> RoutineResult {
```
jmsuzuki: sounds like you did the thing correctly! multi-stage builds are yum.

some docs up here about the actual process and how you dug in to it would be really great to have here as well.
it's very complex! (ask cursor to do it for you, lol.)

jmsuzuki (approved):

PR #2624, opened 2025-08-07
long running workflows

DatGuyJonathan:
Typescript:
- Removed dmv1 workflow
- Added state between task run & oncancel
- Added timeout never
- Added continue-as-new for long running workflows
- Updated templates & tests

Python:
- Removed dmv1 workflow
- Added oncancel handler
- Added state between task run & oncancel
- Added timeout never
- Added continue-as-new for long running workflows
- Updated templates & tests

### DIFF
--- a/packages/ts-moose-lib/src/scripts/workflow.ts
+++ b/packages/ts-moose-lib/src/scripts/workflow.ts
@@ -134,22 +134,20 @@ async function handleDmv2Task(
     executeDmv2Task(workflow, task, inputData)
       .then((taskResult) => {
         return {
-          type: "task_completed",
+          type: "task_completed" as const,
           data: taskResult,
         };
       })
       .finally(() => {
         taskCompleted = true;
       }),
-    monitorTask().then((continueResult) => ({
-      type: "continue_as_new",
-      data: continueResult,
-    })),
+    monitorTask().then(() => {
+      return { type: "continue_as_new" as const, data: undefined };
+    }),
   ]);
 
-  if (result.type === "continue_as_new") {
-    logger.info(`Workflow continuing as new due to history limits`);
-    return result.data;
+  if (result.type !== "task_completed") {
+    return [];
   }
 
   const results = [result.data];
@@ -168,8 +166,10 @@ async function handleDmv2Task(
     task.name.endsWith("_extract") &&
     result &&
     typeof result === "object" &&
-    "hasMore" in result &&
-    result.hasMore === true
+    result.data &&
+    typeof result.data === "object" &&
+    "hasMore" in result.data &&
+    (result.data as any).hasMore === true
   ) {
     logger.info(`Extract task ${task.name} has more data, restarting chain...`);
 


templates/ads-b-frontend/moose/app/scripts/1.fetch_and_ingest_military_aircraft.ts
```diff
@@ -99,7 +99,10 @@ export const fetchAndIngestMilitaryAircraft = new Task<
   FetchTaskInput,
   FetchTaskOutput
 >("fetch_and_ingest_military_aircraft", {
-  run: async (input: FetchTaskInput): Promise<FetchTaskOutput> => {
+  run: async (
+    taskState: any,
```
callicles: is the new API change required? could we make this optional?

DatGuyJonathan: I tried, but it gets really weird because the workflow input is also optional. If they just have one param, I'm not sure if we can know whether they wanted state or input.

callicles: Did you think about this:

```js
function configureUser(options) {
  const defaults = {
    name: "Guest",
    age: 0,
    email: null,
    isActive: true
  };

  const settings = { ...defaults, ...options }; // Merge provided options with defaults

  console.log(`Configuring user:`);
  console.log(`Name: ${settings.name}`);
  console.log(`Age: ${settings.age}`);
  console.log(`Email: ${settings.email}`);
  console.log(`Active: ${settings.isActive}`);
}

// Usage with optional parameters
configureUser({ name: "Alice", age: 30 });
configureUser({ email: "bob@example.com", isActive: false });
configureUser({}); // All defaults
```

packages/ts-moose-lib/src/scripts/workflow.ts
```diff
@@ -25,145 +38,129 @@ export async function ScriptWorkflow(
...
+      const info = workflowInfo();
+
+      // TODO: remove historyLimitChecks >= 10. This is just to test the continue as new functionality
+      if (
+        info.continueAsNewSuggested ||
+        info.historyLength >= 800 ||
+        info.historySize >= 1048576 ||
+        historyLimitChecks >= 10
+      ) {
+        logger.info(
```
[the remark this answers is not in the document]
DatGuyJonathan: Yes, for testing only. Will remove before merge.

packages/ts-moose-lib/src/scripts/activity.ts
```diff
@@ -1,82 +1,35 @@
...
-        scriptModule = await import(pathToFileURL(scriptPath).href);
-      }
-      const execResult = await scriptModule.default();
-      const result = await execResult.task(processedInput);
-
-      return result;
-    } catch (error) {
-      const rawDetails = error instanceof Error ? error.message : String(error);
-      let hint: string | undefined;
-      if (rawDetails.includes("ERR_REQUIRE_ESM")) {
```
callicles, line 39: Did all the error handling for this move to somewhere else?

callicles: Is that the proxyActivities method on the temporal SDK?

DatGuyJonathan: Yea, the temporal activities call a dmv2 fn to load the objects, and that's where the esm check is so others can use it as well https://github.com/514-labs/moosestack/blob/main/packages/ts-moose-lib/src/dmv2/internal.ts#L401

callicles:
Last comment - did you test the build with area code to make sure we did not break anything?

DatGuyJonathan:
> Last comment - did you test the build with area code to make sure we did not break anything?

Yea, lgtm. I'm ready to merge.

PR #2627, opened 2025-08-08
Track CLI flags and send to PostHog

camelCasedAditya:


apps/framework-cli/src/utilities/capture.rs
```diff
@@ -10,6 +10,19 @@ use serde_json::json;
 use std::collections::HashMap;
 use uuid::Uuid;
 
+fn collect_cli_flags() -> Vec<String> {
```
cjus, line 13: Consider stopping flag collection as soon as you encounter a lone --. Anything after that is positional, even if it starts with -.

```rust
fn collect_cli_flags<I, S>(args: I) -> Vec<String>
where
    I: IntoIterator<Item = S>,
    S: AsRef<str>,
{
    let mut flags = Vec::new();
    // skip program name
    let mut iter = args.into_iter().skip(1);

    for arg in iter.by_ref() {
        let s = arg.as_ref();

        // End-of-options sentinel: stop parsing flags entirely.
        if s == "--" {
            break;
        }

        // Only consider things that look like options before the sentinel.
        if s.starts_with('-') {
            // Keep only the key portion (e.g., "--beta=2" -> "--beta")
            if let Some((key, _val)) = s.split_once('=') {
                flags.push(key.to_string());
            } else {
                flags.push(s.to_string());
            }
        } else {
            // First positional before "--" can also terminate option parsing if you want strict POSIX behavior:
            // break;
            // If you prefer GNU-style "intermixing", just continue.
        }
    }

    flags
}
```

cjus: Also looks like dups are possible which above code doesn't address.

Consider a BTreeSet<String> to dedupe and keep deterministic order, then convert to Vec.

cjus (approved):
Looks good - some comments which might improve this - time permitting.

PR #2651, opened 2025-08-18
ENG-673 MOOSE_MANAGEMENT_PORT env var

phiSgr:


apps/framework-cli/src/framework/python/executor.rs
```diff
@@ -94,6 +96,10 @@ pub fn run_python_command(
 
     Command::new("python3")
         .env(PYTHON_PATH, python_path_with_lib(project_location))
+        .env(
+            "MOOSE_MANAGEMENT_PORT",
+            project.http_server_config.management_port.to_string(),
```
callicles, line 101: could you follow the standard with an optional arg instead?

PR #2656, opened 2025-08-18
execute stored migration plan

phiSgr:


apps/framework-cli/src/cli/routines.rs
```diff
@@ -455,7 +459,62 @@ pub async fn start_production_mode(
...
+    if !execute_migration_yaml {
+        info!("Executing pre-planned migrations");
+
+        // Load and validate the approved migration plan
+        let plan_content = std::fs::read_to_string(MIGRATION_FILE)?;
+        let migration_plan: MigrationPlan = serde_yaml::from_str(&plan_content)?;
+
+        info!("Loaded approved migration plan from {:?}", MIGRATION_FILE);
+        info!("Plan created: {}", migration_plan.created_at);
+        info!("Total operations: {}", migration_plan.total_operations());
```
callicles, line 476: [Optional] maybe adding the sha of the migration plan? we will need it anyway to store it in the db

apps/framework-cli/src/cli/routines.rs
```diff
@@ -455,7 +459,62 @@ pub async fn start_production_mode(
...
+
+        let before_state = std::fs::read_to_string(MIGRATION_BEFORE_STATE_FILE)?;
+        let state_when_planned: InfrastructureMap = serde_json::from_str(&before_state)?;
+
+        if current_state.tables == state_when_planned.tables {
```
callicles, line 481: Love this safety check

apps/framework-cli/src/cli/routines.rs
phiSgr: SHA of the JSON? i think that'll depend on iteration order of `HashMap`, which might change every time

> `HashMap` uses a hashing algorithm selected to provide resistance against HashDoS attacks

phiSgr:
> we should not be running the execute via the old way

right. using mutability to solve the problem is probably not the cleanest
`plan.changes.olap_changes = Vec::new();`

apps/framework-cli/src/cli/routines.rs
callicles: That's a good point - I wonder, does that highlight a flaw in the design? Maybe we pass on that for now but we solve before we put this in prod

phiSgr: > I wonder, does that highlight a flaw in the design?

I think comparing the full object is pretty safe. What flaw are you thinking of?

callicles: I am think about eyeballing comparison when we have things in a table.
ie how does someone quickly know 2 things are equal? a quick sha comparison is quick

phiSgr: i think i can add the check for the desired state in `moose check`.
if infra map from code does not match the desired state of the plan: fail, and print the two JSONs.

PR #2657, opened 2025-08-18
Tracking of copy buttons clicked

camelCasedAditya:


camelCasedAditya:
> Looks like you should address AI's issues.
> 
> Qustion, does this create one copy event type, or is it parameterized by the instantiation?

It is one event type for all copy events. It also includes the code snippet that was copied

PR #2660, opened 2025-08-19
ENG-647 add order_by_fields to generated code

phiSgr:


callicles:
Looks like this doesn't build yet?

callicles:
addressing https://linear.app/514/issue/ENG-647/moose-init-from-remote-should-create-olaptables-with-the-correct

callicles (approved):

PR #2672, opened 2025-08-20
Track template choice of user

camelCasedAditya:


apps/framework-cli/src/utilities/capture.rs
```diff
@@ -92,6 +92,7 @@ pub fn capture_usage(
     project_name: Option<String>,
     settings: &Settings,
     machine_id: String,
+    parameters: Option<HashMap<String, String>>,
```
phiSgr: an empty HashMap does not allocate
so i think you can write `parameters: HashMap<String, String>,` which is cleaner without any performance penalty.

phiSgr (approved):

PR #2673, opened 2025-08-20
ENG-646 fix ReplacingMergeTree recreation

phiSgr:


apps/framework-cli/src/cli/display/infrastructure.rs
```diff
@@ -151,46 +151,41 @@ pub fn show_olap_changes(olap_changes: &[OlapChange]) {
...
             column_changes,
             order_by_change,
-            before,
-            after,
+            before: _,
+            after: _,
         }) => {
-            if after.deduplicate != before.deduplicate {
-                infra_removed(&before.expanded_display());
-                infra_added(&after.expanded_display());
```
phiSgr, line 159: This special case is handled by `ClickHouseTableDiffStrategy`.

apps/framework-cli/src/framework/core/infrastructure/table.rs
```diff
@@ -163,7 +164,10 @@ impl Table {
             order_by: self.order_by.clone(),
             version: self.version.as_ref().map(|v| v.to_string()),
             source_primitive: MessageField::some(self.source_primitive.to_proto()),
-            deduplicate: self.deduplicate,
+            deduplicate: self
```
callicles, line 167: Can't you make this optional and only read from it for backwards compatibility?

apps/framework-cli/src/infrastructure/olap/clickhouse/diff_strategy.rs
```diff
@@ -174,8 +175,17 @@ impl TableDiffStrategy for ClickHouseTableDiffStrategy {
...
+            .engine
+            .as_deref()
+            .and_then(|e| ClickhouseEngine::try_from(e).ok());
+        let after_engine = after
+            .engine
+            .as_deref()
+            .and_then(|e| ClickhouseEngine::try_from(e).ok());
+
+        // Check if engine has changed
+        if after_engine.is_some_and(|e| Some(e) != before_engine) {
```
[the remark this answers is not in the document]
callicles: +1

apps/framework-cli/src/framework/core/infrastructure/table.rs
phiSgr: this is already marked as deprecated. im usually wary of changing a type in protobuf

callicles: fair enough

callicles (approved):

PR #2676, opened 2025-08-20
Consumption api to api rename

camelCasedAditya:


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1168,7 +1168,7 @@ async fn router(
             .await
         }
         (_, &hyper::Method::GET, route_segments)
-            if route_segments.len() >= 2 && route_segments[0] == "consumption" =>
+            if route_segments.len() >= 2 && route_segments[0] == "api" =>
```
phiSgr: maybe check both for backwards compatibility?

apps/framework-docs/llm-docs/python/consumption.md
```diff
@@ -1,45 +1,20 @@
-# Consumption APIs
+# Analytics APIs
 
 ## Overview
 
-Moose provides two main types of APIs for exposing data:
+Moose provides one main type of API for exposing data:
```
phiSgr: "one main type" reads a bit awkward.

PR #2683, opened 2025-08-21
add `ClickHouseDefault` type

phiSgr:


apps/framework-cli/src/infrastructure/olap/clickhouse.rs
```diff
@@ -455,17 +466,12 @@ async fn execute_modify_column_comment(
...
     );
 
-    // Get the ClickHouse column type for the ALTER statement
-    let clickhouse_column = std_column_to_clickhouse_column(column.clone())?;
-    let column_type_string = basic_field_type_to_string(&clickhouse_column.column_type)?;
-
     // Escape single quotes in the comment for SQL safety
     let escaped_comment = comment.replace('\'', "''");
 
-    // ClickHouse requires MODIFY COLUMN with the full column definition when changing comment
```
phiSgr, line 465: hallucination?

apps/framework-cli/src/framework/core/infrastructure/table.rs
```diff
@@ -650,10 +642,11 @@ impl Column {
             required: self.required,
             unique: self.unique,
             primary_key: self.primary_key,
-            default: EnumOrUnknown::new(match &self.default {
-                None => ProtoColumnDefaults::NONE,
-                Some(column_default) => column_default.to_proto(),
-            }),
+            default: 0, // ColumnDefaults enum removed in favor of free-form default string
```
callicles: is this `ProtoColumnDefaults::NONE` ?

if so could we make it explicit? 
No risk here of triggering a change?

apps/framework-cli/src/infrastructure/olap/clickhouse.rs
callicles: Interestingly looks like it should be different:

```
ALTER TABLE visits COMMENT COLUMN browser 'This column shows the browser used for accessing the site.'

```

apps/framework-cli/src/infrastructure/olap/clickhouse.rs
```diff
@@ -413,18 +414,28 @@ async fn execute_modify_table_column(
...
+        .as_ref()
+        .map(|d| format!(" DEFAULT {}", d))
+        .unwrap_or_default();
     let modify_column_query = if let Some(ref comment) = clickhouse_column.comment {
         // Escape single quotes in the comment for SQL safety
         let escaped_comment = comment.replace('\'', "''");
         format!(
-            "ALTER TABLE `{}`.`{}` MODIFY COLUMN IF EXISTS `{}` {} COMMENT '{}'",
-            db_name, table_name, clickhouse_column.name, column_type_string, escaped_comment
+            "ALTER TABLE `{}`.`{}` MODIFY COLUMN IF EXISTS `{}` {}{} COMMENT '{}'",
```
callicles: missing a space for the default value?

https://clickhouse.com/docs/sql-reference/statements/alter/column#modify-column

packages/protobuf/infrastructure_map.proto
```diff
@@ -148,7 +148,10 @@ message Column {
   bool required = 3;
   bool unique = 4;
   bool primary_key = 5;
-  ColumnDefaults default = 6;
+  // was an enum that we never used
+  int32 default = 6 [deprecated = true];
```
callicles, line 152: you are not worried about changing the type on this?
Even though we were not using it there was a default value assigned to `None` right?

callicles (commented):
Can you add some tests around what will happen with the protobuff migration and the diff?

callicles:
+ maybe some around the ALTER TABLE generation with the default

packages/protobuf/infrastructure_map.proto
phiSgr: protobuf enum has the representation as an `int32`

callicles: No addiitonal metadata whatsoever or different decoding on the rust side?

apps/framework-cli/src/infrastructure/olap/clickhouse.rs
```diff
@@ -455,17 +466,12 @@ async fn execute_modify_column_comment(
...
-
     // Escape single quotes in the comment for SQL safety
     let escaped_comment = comment.replace('\'', "''");
 
-    // ClickHouse requires MODIFY COLUMN with the full column definition when changing comment
     let modify_comment_query = format!(
-        "ALTER TABLE `{}`.`{}` MODIFY COLUMN `{}` {} COMMENT '{}'",
-        db_name, table_name, column.name, column_type_string, escaped_comment
+        "ALTER TABLE `{}`.`{}` MODIFY COLUMN `{}` COMMENT '{}'",
+        db_name, table_name, column.name, escaped_comment
```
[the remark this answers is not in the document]
phiSgr: `alter table Bar modify column hasText comment 'stuff'` works locally.

apps/framework-cli/src/infrastructure/olap/clickhouse.rs
phiSgr: The `default_clause` starts a space if it's not `None`. This is less obviously correct but the output will not have ugly double spaces.

apps/framework-cli/src/infrastructure/olap/clickhouse.rs
phiSgr: looks like there are two ways to write the same thing

packages/protobuf/infrastructure_map.proto
phiSgr: yeah they're the same.

https://protobuf.dev/programming-guides/proto3/#:~:text=enum%20is%20compatible%20with%20int32

callicles (approved):

PR #2684, opened 2025-08-21
Olivia/python mv fixes

okane16:
- Prohibits mv and table name from being the same
- allows writing to existing target_table, checking types are consistent
- documents the new API
This does not flatten the configs yet, to fast follow


apps/framework-docs/src/pages/moose/olap/model-materialized-view.mdx
```diff
@@ -567,7 +579,7 @@ from pydantic import BaseModel
 from moose_lib import MaterializedView, MaterializedViewOptions, OlapConfig
 
 class Dest(BaseModel):
-    id: str
+    id: Key[str] ## Do not need to set `order_by_fields` for the materialized view with a Key type
```
phiSgr: we still need `order by` for `ReplacingMergeTree` i think

phiSgr (approved):

apps/framework-docs/src/pages/moose/olap/model-materialized-view.mdx
```diff
@@ -887,11 +899,9 @@ agg_stmt = '''
 mv = MaterializedView[MetricsById](MaterializedViewOptions(
     select_statement=agg_stmt,
     select_tables=[reviews],
-    target_table=OlapConfig(
-        name="metrics_by_id",
-        engine=ClickHouseEngines.AggregatingMergeTree,
-        order_by_fields=["id"],
-    ),
+    table_name="metrics_by_id",
```
callicles, line 910: What happens if the user spcificies both table_name and the target_name?

packages/py-moose-lib/moose_lib/dmv2/materialized_view.py
```diff
@@ -5,36 +5,44 @@
...
-        table_name: The name of the underlying target table storing the materialized data.
+        table_name: (Deprecated in favor of target_table) Optional name of the underlying
+                    target table storing the materialized data.
         materialized_view_name: The name of the MATERIALIZED VIEW object itself.
-        engine: Optional ClickHouse engine for the target table.
+        engine: Optional ClickHouse engine for the target table (used when creating
+                a target table via table_name or inline config).
         order_by_fields: Optional ordering key for the target table (required for
                          engines like ReplacingMergeTree).
+        target_table: Either an existing `OlapTable` instance to use as the destination,
```
callicles: This is actually not an option on this no?

packages/py-moose-lib/moose_lib/dmv2/materialized_view.py
```diff
@@ -62,25 +70,38 @@ class MaterializedView(SqlResource, BaseTypedResource, Generic[T]):
...
+            target_table: Optional[OlapTable[T]] = None,
             **kwargs
     ):
         self._set_type(options.materialized_view_name, self._get_type(kwargs))
 
+        # Resolve target table from options
+        if target_table:
+            self.target_table = target_table
+            if self._t != target_table._t:
+                raise ValueError("Target table must have the same type as the materialized view")
```
callicles, line 79: is this a runtime error for the Table concrete type not matching the concrete type of the MaterailziedView?

okane16: yes- but also have compile time check

packages/py-moose-lib/moose_lib/dmv2/materialized_view.py
okane16: oh good catch i deleted it

apps/framework-docs/src/pages/moose/olap/model-materialized-view.mdx
okane16: target_table takes precedence

apps/framework-docs/src/pages/moose/olap/model-materialized-view.mdx
okane16: oh shit yes

apps/framework-docs/src/pages/moose/olap/model-materialized-view.mdx
phiSgr: we ripped out `deduplicate` for the same problem (allowing specifying conflicting/redundant information) though.

packages/py-moose-lib/moose_lib/dmv2/materialized_view.py
okane16: fixed

packages/py-moose-lib/moose_lib/dmv2/materialized_view.py
okane16: yes! i wanted this since it was a better error message than if the CH operation failed. we still also get static type

apps/framework-docs/src/pages/moose/olap/model-materialized-view.mdx
```diff
@@ -13,7 +13,27 @@ import { Lightbulb } from "lucide-react";
 
 ## Overview
 
-Materialized views are write-time pipelines. They run a static `SELECT` to transform data from one or more source table(s) to a target destination table. You query the destination table like any other table.
+Materialized views are write-time pipelines. They run a static `SELECT` to transform data from one or more source table(s) to a target destination table. You query the destination table like any other table. The MaterializedView class wraps ClickHouse's `MATERIALIZED VIEW` functionality, letting you model these transformations in code rather than raw SQL.
```
03cranec: "in code rather than raw SQL" -- careful here, SQL is code. maybe go with "application code"?

03cranec: Alt version (take it or leave it):

The MaterializedView class wraps ClickHouse's `MATERIALIZED VIEW` functionality [link to CH docs]. In ClickHouse, Materialized views are write-time pipelines. They run a static `SELECT` to transform data from one or more source table(s) to a target destination table. You query the destination table like any other table. This is useful for pre-computing derivative data sets to optimize performance. 

The MooseStack MaterializedView abstraction lets you model your materialized views in type-safe application code, while still giving you access to the actual SQL transformation logic.

apps/framework-docs/src/pages/moose/olap/model-materialized-view.mdx
```diff
@@ -13,7 +13,27 @@ import { Lightbulb } from "lucide-react";
...
+
+<BulletPointsCard title="Key differences from other databases" bulletStyle="number" compact bullets={[
```
03cranec: Confusing headline. Seems like its referring to MooseStack at first. Maybe something like:
"Key difference with ClickHouse MVs, vs other databases"

apps/framework-docs/src/pages/moose/olap/model-materialized-view.mdx
```diff
@@ -13,7 +13,27 @@ import { Lightbulb } from "lucide-react";
...
+    "Requires both a transformation query AND a target table definition",
```
03cranec: "Requires writing/aligning/executing two SQL queries: a transformation query AND a target table definition"

03cranec: this one is a con, while the other are pros, right? might be worth making that clear. aren't triggers a unique CH thing too? this list could be framed as you get XYZ benefits, but that comes with complexity/cost ABC

apps/framework-docs/src/pages/moose/olap/model-materialized-view.mdx
```diff
@@ -13,7 +13,27 @@ import { Lightbulb } from "lucide-react";
...
+    "Automatically handles backfilling historical data",
+    "Manages dependencies between multiple materialized views"
+]} />
+
+<BulletPointsCard title="What Moose handles for you"
+bulletStyle="check"
+compact
+bullets={[
+    "Orchestrates the transformation query and target table setup",
```
03cranec: is it clear to a reader what orchestrating means here, and why its valuable? might want to be more explicit? 
- only write one sql query
- don't have to execute anything
(i think?)

apps/framework-docs/src/pages/moose/olap/model-materialized-view.mdx
```diff
@@ -13,7 +13,27 @@ import { Lightbulb } from "lucide-react";
...
+    "Coordinates cascading dependencies as your pipeline grows",
+    "Automatically manages setup/teardown when you modify pipelines"
```
03cranec: this sounds scary. maybe mention moose migrate? ie. nothing happens magically that won't expect.

03cranec: link to nico's post on this?

apps/framework-docs/src/pages/moose/olap/model-materialized-view.mdx
```diff
@@ -13,7 +13,27 @@ import { Lightbulb } from "lucide-react";
...
+]}
+/>
+
+<Callout type="info" title="Dependency awareness" compact icon={Lightbulb}>
+This dependency awareness becomes crucial as you build more complex data pipelines with multiple materialized views that depend on each other.
```
03cranec: or maybe link to the post here actually

apps/framework-docs/src/pages/moose/olap/model-materialized-view.mdx
```diff
@@ -13,7 +13,27 @@ import { Lightbulb } from "lucide-react";
...
-Materialized views are write-time pipelines. They run a static `SELECT` to transform data from one or more source table(s) to a target destination table. You query the destination table like any other table.
+Materialized views are write-time pipelines. They run a static `SELECT` to transform data from one or more source table(s) to a target destination table. You query the destination table like any other table. The MaterializedView class wraps ClickHouse's `MATERIALIZED VIEW` functionality, letting you model these transformations in code rather than raw SQL.
+
+<BulletPointsCard title="Key differences from other databases" bulletStyle="number" compact bullets={[
+    "Requires both a transformation query AND a target table definition",
+    "Automatically handles backfilling historical data",
+    "Manages dependencies between multiple materialized views"
+]} />
+
+<BulletPointsCard title="What Moose handles for you"
```
03cranec: maybe something about the benefit of having the typed object in your app code? or maybe that is already very clear as a moose core thing by the time a user makes it here

apps/framework-docs/src/pages/moose/olap/model-materialized-view.mdx
03cranec: is this is really were the value is now, i'm not sure its actually hammered home. maybe a little more on what the pain is without moose?

PR #2687, opened 2025-08-21
Scan package.json for engines, and use appropriate docker image

georgevanderson:


apps/framework-cli/src/utilities/nodejs_version.rs
```diff
@@ -0,0 +1,220 @@
...
+/// This should be updated periodically or ideally fetched from Node.js release API
+const NODE_LTS_VERSIONS: &[NodeVersion] = &[
+    NodeVersion {
+        major: 18,
+        is_lts: true,
+    },
+    NodeVersion {
+        major: 20,
+        is_lts: true,
+    },
```
phiSgr: i thought we do not run on 18?

jmsuzuki (approved):

PR #2694, opened 2025-08-22
eagerly load user API

phiSgr:


callicles (approved):

callicles:
@phiSgr did you test changing the api to make sure they get reloaded properly?

phiSgr:
yeah it reloads fine.
reloading is done by rust side restarting the processes.

PR #2698, opened 2025-08-22
Open api spec update and better error messages

camelCasedAditya:


apps/framework-cli/src/framework/core/primitive_map.rs
```diff
@@ -41,7 +41,7 @@ pub enum PrimitiveMapLoadingError {
     #[error("Failed to load functions")]
     FunctionsLoading(#[from] crate::framework::streaming::model::FunctionError),
 
-    #[error("Failed to load consumption")]
+    #[error("Failed to load analytics apis")]
     Consumption(#[from] ConsumptionLoaderError),
```
phiSgr: rename this enum case as well?

phiSgr (approved):

callicles (approved):

PR #2705, opened 2025-09-02
add replacingmergetree(ver)

DatGuyJonathan:


apps/framework-cli/src/framework/core/infrastructure/table.rs
```diff
@@ -84,6 +84,8 @@ pub struct Table {
     pub order_by: Vec<String>,
     #[serde(default)]
     pub engine: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub replacing_merge_tree_dedup_by: Option<String>,
```
phiSgr, line 88: i think "dedup by" should refer to the `ORDER BY` fields. maybe `choose_by` fits the semantics better, or just call it `version_col` like CH does, without inventing a new name?

also we need to think about the params for other engine types, do we put all of them as string fields in `Table`?
or should we just do `pub engine_params: Vec<String>`

PR #2708, opened 2025-09-03
fix hyphens when direct insert or mv

DatGuyJonathan:


packages/ts-moose-lib/src/utilities/sql.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Quote a ClickHouse identifier with backticks if not already quoted.
+ * Backticks allow special characters (e.g., hyphens) in identifiers.
+ */
+export const quoteIdentifier = (name: string): string => {
+  const trimmed = name.trim();
```
phiSgr: i'm not sure why we should trim it

packages/ts-moose-lib/src/utilities/sql.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Quote a ClickHouse identifier with backticks if not already quoted.
+ * Backticks allow special characters (e.g., hyphens) in identifiers.
+ */
+export const quoteIdentifier = (name: string): string => {
```
phiSgr: maybe put this in `sqlHelpers.ts`?

PR #2710, opened 2025-09-04
default docs v1

okane16:


apps/framework-docs/src/pages/moose/olap/model-table.mdx
```diff
@@ -234,6 +234,60 @@ class Order(BaseModel):
...
+
+interface Event {
+  id: Key<string>;
+  // Static defaults (ClickHouse expression as a string literal)
+  status: string & ClickHouseDefault<"'pending'">;   // DEFAULT 'pending'
+  retries: number & ClickHouseDefault<"0">;          // DEFAULT 0
+  // Server-side timestamps
+  createdAt: Date & ClickHouseDefault<"now()">;      // DEFAULT now()
+  // Decimal with default
+  amount: (string & ClickHouseDecimal<10, 2> & ClickHouseDefault<"0">);
```
DatGuyJonathan, line 253: I'm not sure how to read this. The amount is a string (not a number?), but also a decimal, and whats 10 and 2 and 0 (as a string)?

okane16: its confusing ish the value is a sql string

PR #2714, opened 2025-09-04
add truncate and batching for seed in case of large limit or ALL

okane16:
- Adds a new moose truncate cli command that clears contents of the specified table(s), either ALL or a certain number
- Adds ability to seed -all rows of a table, using batching for large tables. Also updates regular seed with limit to use the batching so that you can do large seeds

### DIFF
--- a/apps/framework-cli/src/cli/routines/truncate_table.rs
+++ b/apps/framework-cli/src/cli/routines/truncate_table.rs
@@ -1,6 +1,8 @@
 use crate::cli::display::Message;
 use crate::cli::routines::{RoutineFailure, RoutineSuccess};
-use crate::infrastructure::olap::clickhouse::{check_ready, create_client, run_query};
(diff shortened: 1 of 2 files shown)


apps/framework-cli/src/cli/routines/seed_data.rs
```diff
@@ -198,18 +217,266 @@ pub async fn seed_clickhouse_tables(
     let remote_host_and_port = format!("{remote_host}:{remote_port}");
 
     for table_name in tables {
-        let sql = format!(
-            "INSERT INTO `{local_db}`.`{table_name}` SELECT * FROM remoteSecure('{remote_host_and_port}', '{remote_db}', '{table_name}', '{remote_user}', '{remote_password}') LIMIT {limit}"
```
phiSgr, line 202: I think we can do batching within this SQL, without the data going through rust side.

PR #2726, opened 2025-09-06
mitchell/chore-print: available routes

jmsuzuki:
**chore: print available moose routes on startup**


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1575,6 +1575,149 @@ pub struct Webserver {
...
+                .map_or("latest".to_string(), |v| v.to_string())
+        );
+        ingest_routes.push((method, endpoint, description));
+    }
+
+    // Collect consumption/API routes
+    let mut consumption_routes = Vec::new();
+    for api_path in consumption_apis_read.iter() {
+        let method = "GET";
+        let endpoint = format!("/consumption/{}", api_path);
```
callicles: we have renamed those I believe with `/api`

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -1575,6 +1575,149 @@ pub struct Webserver {
...
+    let mut static_routes = vec![
+        (
+            "GET",
+            "/admin/inframap".to_string(),
+            "Admin: Get infrastructure map".to_string(),
+        ),
+        (
+            "GET",
+            "/admin/reality-check".to_string(),
+            "Admin: Reality check".to_string(),
```
jmsuzuki: this could use a better description

apps/framework-cli/src/cli/local_webserver.rs
jmsuzuki: ah! great call here.
consumption still works for backwards compatibility. i assume.

callicles: yeah

jmsuzuki:
> Hum available routes are dynamic. Do you only want the routes at start up?
> 
> There is probably a way to do this without re-implementing some of the logic?

how are routes dynamic?
minimally, I wanted to see the routes at startup. I was unaware that they could change in a production setting.

and yes, running `moose dev`, the routes can be changed and an update isn't posted regarding it.
nor is the table reprinted. should it?

callicles:
Yeah I meant in dev

callicles (approved):

PR #2734, opened 2025-09-08
fix direct insert in boreal

DatGuyJonathan:


DatGuyJonathan:
> why not pass those coming from the rust side so that we're not parsing those values in 3 languages?

These clickhouse clients are standalone, no? It's not like a streaming function or api process that we spawn and pass configs to.

phiSgr:
ah true, should we use a library instead of rolling our own implementation?

https://www.dynaconf.com/envvars/

callicles:
I believe this is a blocker for a customer - we should probably get this through and then think how we want to iterate.
The tricky part with the third party lib is that we will want the lib to work exactly the same in the 3 languages

PR #2738, opened 2025-09-09
handle array nullability in ClickHouseTableDiffStrategy

phiSgr:


DatGuyJonathan (approved):
We can probably add a quick test or two in `diff_strategy.rs`

PR #2739, opened 2025-09-09
show queries from query client

DatGuyJonathan:


packages/ts-moose-lib/src/consumption-apis/runner.ts
```diff
@@ -169,6 +169,7 @@ const apiHandler = async (
           }
 
           modulesCache.set(pathName, userFuncModule);
+          console.log(`[QueryClient] | Executing API: ${apiName}`);
```
phiSgr: a bit weird to write `QueryClient` here

PR #2744, opened 2025-09-10
overwrite external file with db pull

phiSgr:


apps/framework-cli/src/cli/routines/code_generation.rs
```diff
@@ -350,3 +426,89 @@ pub async fn db_to_dmv2(remote_url: &str, dir_path: &Path) -> Result<(), Routine
...
+        debug!("Loading InfrastructureMap from user code (DMV2)");
+        InfrastructureMap::load_from_user_code(&project)
+            .await
+            .map_err(|e| {
+                RoutineFailure::error(Message::new(
+                    "Failure".to_string(),
+                    format!("loading infra map: {e:?}"),
+                ))
+            })?
+    } else {
```
DatGuyJonathan: We probably don't need this (no more dmv1 :P)

PR #2747, opened 2025-09-10
Nicolas/eng 617 no support for ver in replacingmergetree

callicles:


phiSgr (approved):
> `is_deleted` can only be enabled when `ver` is used.

would be great if we can enforce this at the type level

BTW, i don't find it obvious that `is_deleted` is a field name

callicles:
Yeah I just reused the clickhouse names since that what people will be looking at in the docs

apps/framework-cli/src/infrastructure/olap/clickhouse/queries.rs
```diff
@@ -361,7 +361,9 @@ impl ClickhouseEngine {
             }
             if let Some(d) = is_deleted {
                 // Only add is_deleted if ver is present (validated elsewhere)
-                params.push(format!("'{}'", d));
+                if ver.is_some() {
```
phiSgr, line 364: if this is validated elsewhere, then when ver is `None`, we should crash and expose the bug

PR #2754, opened 2025-09-11
transform should use dead letter queue in ingest pipeline by default

phiSgr:


phiSgr:
> How is the default deadletter queue configured?

In `StreamConfig`, or automatically set if the IngestApi has the dlq configured.

PR #2755, opened 2025-09-11
add partition_by

phiSgr:


apps/framework-cli/src/framework/core/partial_infrastructure_map.rs
```diff
@@ -149,6 +149,8 @@ struct PartialTable {
     pub columns: Vec<Column>,
     #[serde(alias = "order_by")]
     pub order_by: Vec<String>,
+    #[serde(default)]
+    pub partition_by: Option<String>,
```
callicles, line 153: Add alias "partition_by" for consistency

apps/framework-cli/src/infrastructure/olap/clickhouse.rs
```diff
@@ -1371,6 +1376,31 @@ pub fn extract_order_by_from_create_query(create_query: &str) -> Vec<String> {
     Vec::new()
 }
 
+/// Extracts PARTITION BY expression from a CREATE TABLE query, if present
+pub fn extract_partition_by_from_create_query(create_query: &str) -> Option<String> {
```
callicles: To remove

apps/framework-cli/src/framework/core/partial_infrastructure_map.rs
phiSgr: maybe delete `#[serde(alias = "order_by")]` for consistency :P

PR #2759, opened 2025-09-14
documentation on pydantic alias

phiSgr:


apps/framework-docs/src/pages/moose/data-modeling.mdx
okane16, line 1: make sure to add <Python></Python> tags around this

apps/framework-docs/src/pages/moose/getting-started/from-clickhouse.mdx
okane16, line 1: same here

03cranec:
@phiSgr i moved the location of the documentation to another page. i also refactored it a bit to make it generic advice on dealing with irregular column names, instead of just documenting how `init from remote` does it. can you confirm this is still accurate?

apps/framework-docs/src/pages/moose/olap/model-table.mdx
```diff
@@ -612,6 +612,23 @@ s3_events = OlapTable[S3Event]("s3_events", OlapConfig(
 S3Queue requires ClickHouse 24.7+ and proper ZooKeeper/ClickHouse Keeper configuration for coordination between replicas. Files are processed exactly once across all replicas.
 </Callout>
 
+<Python>
```
03cranec, line 615: is there any similar behavior for TS @phiSgr ?

phiSgr: Not really. Only quotation marks needed.

okane16 (approved):

PR #2771, opened 2025-09-18
fix: allow destination to be specified in py StreamConfig

phiSgr:


phiSgr:
~~Other types might also be affected, will have a followup PR.~~

turns out this is the only one

callicles (approved):

PR #2772, opened 2025-09-18
fix docs

okane16:


apps/framework-docs/src/pages/moose/streaming/create-stream.mdx
```diff
@@ -65,18 +65,18 @@ export const rawIngestionStream = new IngestPipeline<RawData>("raw_data", {
...
+from moose_lib import IngestPipeline, IngestPipelineConfig, Key
 from pydantic import BaseModel
 
 class RawData(BaseModel):
     id: Key[str]
     value: int
 
-raw_ingestion_stream = IngestPipeline[RawData]("raw_data", {
+raw_ingestion_stream = IngestPipeline[RawData]("raw_data", IngestPipelineConfig(
   ingest: True, # Creates an ingestion API endpoint at `POST /ingest/raw_data`
```
03cranec: need " = " instead of ": " --> this is an issue all over this docs page

apps/framework-docs/src/pages/moose/streaming/create-stream.mdx
```diff
@@ -109,20 +109,20 @@ export const rawIngestApi = new IngestApi<RawData>("raw_data", {
...
     value: int
 
 raw_table = OlapTable[RawData]("raw_data")
-raw_stream = Stream[RawData]("raw_data", {
+raw_stream = Stream[RawData]("raw_data", StreamConfig(
   destination: raw_table # Optional: Specify a destination table for the stream, sets up a process to sync data from the stream to the table
-})
-raw_ingest_api = IngestApi[RawData]("raw_data", {
+))
+raw_ingest_api = IngestApi[RawData]("raw_data", IngestConfig(
```
03cranec, line 123: this only works for me with "IngestConfigWithDestination". not sure if that is a bug, or intended?

03cranec: also homepage docs says "IngestAPIConfig" which does NOT work. we really need to test docs code better.

03cranec (changes requested):

apps/framework-docs/src/pages/moose/streaming/create-stream.mdx
03cranec: cc @phiSgr on the potential bug

apps/framework-docs/src/pages/moose/streaming/create-stream.mdx
```diff
@@ -209,18 +209,18 @@ class TransformedData(BaseModel):
     transformedAt: Date    # Timestamp of transformation
 
 # Create pipeline for raw data - only for ingestion and streaming
-raw_data = IngestPipeline[RawData]("raw_data", {
+raw_data = IngestPipeline[RawData]("raw_data", IngestPipelineConfig(
   ingest: True,   # Enable API endpoint
```
03cranec: i thought we changed this specific setting from begin called "ingest" to "api" or "ingest_api" like weeks ago?? seems like maybe that never happened? that is urgent papercut to be fixed, because "ingest = true/false" on an "ingest pipeline" makes no sense. cc @phiSgr this needs to be updated on python and TS ingest pipeline config objects. my vote is just calling it "api". its clear its an ingest api since its a setting on an ingest pipeline

okane16: i believe its actually ingest_api now, and this was missed whenever that got updated

apps/framework-docs/src/pages/moose/streaming/create-stream.mdx
okane16: seems like a bug

apps/framework-docs/src/pages/moose/streaming/create-stream.mdx
okane16: ingest still works since its backwards compatible

apps/framework-docs/src/pages/moose/streaming/create-stream.mdx
okane16: fixed homepage

apps/framework-docs/src/pages/moose/streaming/create-stream.mdx
okane16: fixed

PR #2776, opened 2025-09-18
allow params dict to include tables and columns in python

phiSgr:


packages/py-moose-lib/moose_lib/main.py
```diff
@@ -193,12 +196,16 @@ def execute(self, input, variables, row_type: Type[BaseModel] = None):
...
-                values[f'p{i}'] = value
+                if isinstance(value, Column) or isinstance(value, OlapTable):
+                    params[variable_name] = f'{{p{i}: Identifier}}'
+                    values[f'p{i}'] = value.name
+                else:
+                    t = 'String' if isinstance(value, str) else \
+                        'Int64' if isinstance(value, int) else \
+                            'Float64' if isinstance(value, float) else \
+                                "String"  # unknown type
+                    params[variable_name] = f'{{p{i}: {t}}}'
```
callicles: Don't we need to add support for all the types here?

PR #2786, opened 2025-09-23
handle paths in tsconfig.json

phiSgr:


phiSgr:
Was monkey-patching to test, somehow doing
`require("tsconfig-paths/register")` (in object passed to `register`), 
`require: ["tsconfig-paths/register"]`, and
`experimentalResolver: true`
all 3 made it fail.
We need only the first.

PR #2788, opened 2025-09-23
fix Nullable(Nullable(...)) inside Nested

phiSgr:


apps/framework-cli/src/infrastructure/olap/clickhouse/queries.rs
```diff
@@ -1034,8 +1034,10 @@ pub fn basic_field_type_to_string(
                 .map(|col| {
                     let field_type_string = basic_field_type_to_string(&col.column_type)?;
                     match col.required {
-                        true => Ok(format!("{} {}", col.name, field_type_string)),
-                        false => Ok(format!("{} Nullable({})", col.name, field_type_string)),
+                        false if !matches!(col.column_type, Nullable(_)) => {
```
callicles, line 1038: I don't get it - why is this `!matches!(col.column_type, Nullable(_))` ? If this is nullable we want it nullable no?

PR #2807, opened 2025-09-26
strip nullability from array/nested

phiSgr:


apps/framework-cli/src/infrastructure/olap/clickhouse/queries.rs
```diff
@@ -1056,8 +1056,14 @@ pub fn basic_field_type_to_string(
         }
         ClickHouseColumnType::Nullable(inner_type) => {
             let inner_type_string = basic_field_type_to_string(inner_type)?;
-            // <column_name> String NULL is equivalent to <column_name> Nullable(String)
-            Ok(format!("Nullable({inner_type_string})"))
+            match inner_type.as_ref() {
+                ClickHouseColumnType::Array(_) | ClickHouseColumnType::Nested(_) => {
+                    info!("Nullability stripped from array/nested field as this is not allowed in ClickHouse.");
```
okane16, line 1069: Maybe specify add default value? or only let the nullability be stripped if there is a default? or add a default as empty array/object?

lean towards 1/2 3 feels too magical

phiSgr: clickhouse has empty arrays by default

callicles (approved):

PR #2813, opened 2025-09-30
RFD to enable BYOF for moose APIs

onelesd:
### Summary

This RFD proposes an opt-in approach for MooseStack users to use their own HTTP frameworks (Express, Fastify, Koa, etc.) for Consumption APIs, while maintaining full backward compatibility with the existing batteries-included `Api` class.

### Problem

The current `Api` class automatically instantiates a Node.js HTTP server, which limits users to a basic implementation without access to mature framework features like middleware, plugins, and advanced HTTP capabilities. This creates friction when integrating MooseStack into existing applications or when users need production-grade features.

### Proposed Solution

Export a `getMooseClients()` function that initializes and returns ClickHouse, Temporal, and Redis clients. Users can then leverage these clients in their own HTTP framework handlers:

```typescript
import express from 'express';
import { getMooseClients, sql } from '@514labs/moose-lib';

const app = express();
const { clickhouse, temporal } = await getMooseClients();

app.get('/api/data', async (req, res) => {
  const result = await clickhouse.execute(
    sql`SELECT * FROM MyTable LIMIT ${req.query.limit || 10}`
  );
  res.json(await result.json());
});
```

### Key Features

- **Opt-in**: Existing `Api` class continues to work unchanged
- **Framework agnostic**: Works with any Node.js HTTP framework
- **Full utility access**: Preserves `client`, `sql`, `MooseCache`, and workflow capabilities
- **Type-safe**: Maintains TypeScript validation support
- **Flexible configuration**: Supports explicit config, environment variables, or moose.config.ts

### Scope

- TypeScript implementation (Python parity planned separately)
- Consumption APIs only (IngestApi requires separate consideration due to Rust proxy complexity)
- Maintains all existing functionality and deployment paths

---

> [!NOTE]
> Adds an RFD detailing a BYOF approach for Consumption APIs, introducing getMooseClients, optional framework helpers, and backwards-compatible usage.
> 
> - **RFD: Bring Your Own Framework (BYOF) for Consumption APIs** (`rfd/0008/README.mdx`)
>   - Proposes exporting `getMooseClients()` to initialize `clickhouse`, `temporal`, and optional `cache` clients, with config via args/env/moose.config.
>   - Preserves existing utilities (`sql`, `MooseCache`, `QueryClient`, `WorkflowClient`) and compatibility with `Api`/`runApis()`.
>   - Optional framework helpers: `mooseMiddleware` for Express, `moosePlugin` for Fastify.
>   - Outlines configuration, lifecycle with `moose dev`, OpenAPI considerations (deferred for BYOF), and exclusion of `IngestApi` (future work).
>   - Provides concise Express/Fastify integration examples and migration guidance.
>   - Defines rollout/testing plan, environments, and implementation checklist.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 102660f426c026821fe4b20b0b33f8ce89b8da6b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/rfd/0008/README.mdx
+++ b/rfd/0008/README.mdx
@@ -127,6 +127,14 @@ app.listen(3000);
 
 #### 1. Client Initialization Function
 
+The configuration will be compiled by the `getMooseClients()` function. It's
+logic will follow:
+
+1. read configuration object passed to it
+2. if any fields are missing, look for them in `moose.config.toml`
+3. if any fields are missing, look for them, in the environment variables
+4. if configuration is invalid, throw an error
+
 Export a function to initialize MooseStack clients from configuration:
 
 ```typescript
@@ -839,4 +847,3 @@ Once this RFD is approved, implementation will follow this checklist:
 - [ ] `moose dev` compatibility
 - [ ] Boreal/OSS deployment compatibility
 - [ ] Plan Python parity implementation
-
--- a/rfd/0008/README.mdx
+++ b/rfd/0008/README.mdx
@@ -28,7 +28,6 @@ mature, battle-tested frameworks.
 - Replacing or deprecating the existing `Api` class approach
 - Refactoring IngestApi in this RFD (requires separate consideration due to
   Rust proxy complexity)
-- Addressing production deployment implications (to be revisited if needed)
 - Modifying type validation or OpenAPI generation mechanisms
 
 ## User Benefit
@@ -576,9 +575,9 @@ End-to-end tests in `apps/framework-cli-e2e`:
 
 ### Production Deployment Considerations
 
-**Deferred**: This RFD focuses on API design and local development. Production
-deployment implications (containerization, process management, configuration
-management) will be addressed separately if needed.
+This RFD focuses on API design, local development and deployments to Boreal.
+Since deploying to Boreal uses the same path OSS users of moosestack will use,
+all current paths to production should be covered.
 
 **Initial guidance**: Users deploying BYOF should:
 
@@ -837,5 +836,7 @@ Once this RFD is approved, implementation will follow this checklist:
 - [ ] Update main README
 - [ ] Add migration guide from `Api` class to BYOF
 - [ ] Add end-to-end tests in `framework-cli-e2e`
+- [ ] `moose dev` compatibility
+- [ ] Boreal/OSS deployment compatibility
 - [ ] Plan Python parity implementation
 
--- a/rfd/0008/README.mdx
+++ b/rfd/0008/README.mdx
@@ -1,0 +1,861 @@
+---
+authors: Dave Seleno <dave@fiveonefour.com>
+title: Bring Your Own Framework (BYOF) for Consumption APIs
+state: discussion
+---
+
+## Objective
+
+Enable MooseStack library consumers to use their own HTTP server frameworks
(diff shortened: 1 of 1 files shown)


phiSgr:
maybe add a paragraph to address any complication from the compiler plugin that is needed to instantiate moose objects (e.g. tables)?

rfd/0008/README.mdx
```diff
@@ -0,0 +1,842 @@
...
+  clickhouse: QueryClient;
+  temporal?: WorkflowClient;
+  cache?: MooseCache;
+}
+
+export async function getMooseClients(
+  config?: MooseClientsConfig
+): Promise<MooseClients> {
+  // If no config provided, read from environment or moose.config
+  const finalConfig = config ?? await loadMooseConfig();
```
DatGuyJonathan, line 175: How do we know where their moose config is?

onelesd: this is yet to be fleshed out, but we're thinking to follow an order of precedence when `getMooseClients()` is called:

1. object passed to `getMooseClients({...})`
2. moose toml
3. environment vars
4. error if not found or incomplete

onelesd: added this logic: https://github.com/514-labs/moosestack/pull/2813/commits/c216289433644fd9d931d3d46432240cf906cd92

DatGuyJonathan: Okay cool. The moose toml file provides defaults mainly for local dev, checked in to git, and gets overridden in prod through env vars, so I wonder if env vars also take precedence over moose toml in this new client.

I also wasn't sure where they could put their byof framework code. If it's strictly in the moose app folder, then we could walk the tree and find the toml file. But libifying moose makes me think it can be in a separate folder (monorepo or even separate repo), so maybe they need a way to tell us exactly where their moose app is.

okane16: > so maybe they need a way to tell us exactly where their moose app is
 i think this would be a pretty big improvement actually. let the user configure this

onelesd: > env vars also take precedence over moose toml

got it. it makes sense to do the same here then. updated the logic to reflect that: https://github.com/514-labs/moosestack/pull/2813/commits/d281a520ad4d163a8c7a9bdd94f4f3423bc79295

rfd/0008/README.mdx
```diff
@@ -0,0 +1,859 @@
...
+    port: 9000,
+    username: 'default',
+    password: 'password',
+  },
+});
+```
+
+### Lifecycle Management with `moose dev`
+
+When users choose BYOF:
```
okane16, line 282: how does the developer opt into BYO APIs? Is there a feature in the moose config where you tell the CLI "don't start consumption api server"?

onelesd: tl;dr: we shouldn't need new config to control this

the current behavior is to start the node consumption api server no matter what. if no `Api` were created its an empty server with no routes. since creating `Api` instances populates a registry it would be trivial to check that registry before starting the node server and bailing out if the registry is empty of consumption apis. this approach would allow users to mix and match `Api` with their BYOF apis if desired.

PR #2851, opened 2025-10-10
fix: shutdown order of connected clients

callicles:
> [!NOTE]
> Gracefully shuts down HTTP servers and sync processes, flushes Kafka producer, waits for librdkafka destruction, and refactors registries to support coordinated process cleanup.
> 
> - **Shutdown Orchestration**:
>   - Graceful HTTP shutdown with environment-specific timeouts; user-facing shutdown messages.
>   - Flush Kafka producer after HTTP drain; then drop services and add brief grace periods.
>   - In dev, stop workflows and optionally Docker containers; final shutdown status message.
> - **Kafka/Librdkafka**:
>   - Add `rdkafka-sys` and call `rd_kafka_wait_destroyed` with timeout to await client teardown.
>   - Keep/clone `ConfiguredProducer` for post-HTTP flush; import `Producer` where needed.
> - **Sync Processes (Kafka→CH, Topic→Topic)**:
>   - Introduce cancelable tasks via oneshot channels; graceful stop with timeout and optional producer flush.
>   - New structures (`TableSyncProcess`, `TopicSyncProcess`) and `SyncingProcessesRegistry::stop_all()`.
>   - Loops now `select!` on cancel signals.
> - **Process Registry**:
>   - Embed `syncing` registry into `ProcessRegistries`; `stop()` now halts functions, consumption, orchestration, and syncing.
>   - Update constructors and call sites; remove separate syncing registry threading through APIs.
> - **API/Execution Refactors**:
>   - `execute_initial_infra_change` now returns only `ProcessRegistries`; `execute_online_change` no longer takes syncing registry.
>   - Adjust `routines.rs` and `watcher.rs` to new signatures.
> - **System Utilities**:
>   - `kill_child` sends SIGTERM, waits up to 10s, escalates to SIGKILL if needed; `RestartingProcess::stop` awaits clean exit.
> - **TypeScript Cluster Utils**:
>   - On worker shutdown, skip `workerStop` if startup not completed to avoid errors.
> - **Misc**:
>   - Tweak health/ready checks and log levels; minor cleanups.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dce6f96c494e0227a6ccdb7e1a50c10e148eab59. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/cli/local_webserver.rs
+++ b/apps/framework-cli/src/cli/local_webserver.rs
@@ -2494,6 +2494,7 @@ async fn shutdown(
 
     // Step 1: Stop all managed processes (functions, syncing, consumption, orchestration workers)
     // This sends termination signals and waits with timeouts for all processes to exit
+    // Note: This happens in BOTH dev and production - workers must always be stopped gracefully
     let stop_result = with_spinner_completion_async(
         "Stopping managed processes (functions, syncing, consumption, workers)",
         "Managed processes stopped",
@@ -2501,7 +2502,7 @@ async fn shutdown(
             let mut process_registry = process_registry.write().await;
             process_registry.stop().await
         },
-        !project.is_production,
+        !project.is_production, // Show spinner in dev only
     )
     .await;
 
@@ -2549,12 +2550,17 @@ async fn shutdown(
     .await;
 
     // Step 2: Shutdown workflows (if enabled)
+    // Note: This only runs in development (!project.is_production) because:
+    // - In dev: We want a clean slate - terminate workflow executions so they don't resume on next start
+    // - In production: Workers are already stopped (Step 1), but workflow executions should continue
+    //   running so other workers or future restarts can pick them up. Terminating production workflows
+    //   should be an explicit operational decision, not automatic on every deployment.
     if !project.is_production && project.features.workflows {
         let termination_result = with_spinner_completion_async(
             "Stopping workflows",
             "Workflows stopped",
             async { terminate_all_workflows(project).await },
-            true,
+            true, // Always show spinner in dev (this code only runs in dev)
         )
         .await;
 
@@ -2632,7 +2638,11 @@ async fn shutdown(
         MessageType::Success,
         Message {
             action: "Shutdown".to_string(),
-            details: "Dev server shutdown complete".to_string(),
+            details: if project.is_production {
+                "Server shutdown complete".to_string()
+            } else {
+                "Dev server shutdown complete".to_string()
+            },
         },
     );
 


apps/framework-cli/src/utilities/system.rs
```diff
@@ -56,10 +57,50 @@ pub async fn kill_child(child: &Child) -> Result<(), KillProcessError> {
...
+                    warn!("Process {} killed with SIGKILL", id);
+                    Ok(())
+                }
+                _ => {
+                    error!("Process {} could not be killed even with SIGKILL", id);
+                    Err(KillProcessError::ProcessNotFound)
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
callicles: This is aleady in the back up - I believe that's acceptable risk

callicles:
In a container env we won't have the disk. But we only ack to the source if the data was written.
So that guarantees that if things did not make it we will retry. We have at least once delivery so duplictaes are ok

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -2216,13 +2225,16 @@ impl Webserver {
             .unwrap_or_else(|e| handle_listener_err(proxy_socket.port(), e));
 
         let producer = if project.features.streaming_engine {
-            Some(kafka::client::create_producer(
+            Some(Arc::new(kafka::client::create_producer(
```
phiSgr: the kafka producer has an `Arc` internally so it's cheaply clone-able, no need for an extra `Arc`

apps/framework-cli/src/infrastructure/processes/kafka_clickhouse_sync.rs
```diff
@@ -53,33 +53,25 @@ const VERSION_SYNC_GROUP_ID: &str = "version_sync_flow_sync";
...
+    /// Map of topic-table processes by their combined key with cancellation senders
+    to_table_registry: HashMap<
+        String,
+        (
+            JoinHandle<anyhow::Result<()>>,
+            tokio::sync::oneshot::Sender<()>,
+        ),
+    >,
+    /// Map of topic-to-topic processes by their target topic name with cancellation senders
+    to_topic_registry: HashMap<String, (JoinHandle<()>, tokio::sync::oneshot::Sender<()>)>,
```
phiSgr: maybe make the pair into a struct so that we don't have to look at the comment for the meaning of the channel?

apps/framework-cli/src/infrastructure/processes/kafka_clickhouse_sync.rs
```diff
@@ -233,10 +350,11 @@ fn spawn_sync_process_core(
     target_table_name: String,
     target_table_columns: Vec<ClickHouseColumn>,
     metrics: Arc<Metrics>,
-) -> TableSyncingProcess {
+    cancel_rx: tokio::sync::oneshot::Receiver<()>,
```
phiSgr: maybe make this function create the channel
and return the `Sender` as part of the output?
i think that'll be cleaner

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -2388,7 +2416,58 @@ impl Webserver {
...
+        // Use different timeouts for dev vs production:
+        // - Dev: 2 seconds (balance between fast shutdown and allowing in-flight requests to complete)
+        // - Production: 10 seconds (allow load balancers time to drain connections during rolling deployments)
+        let shutdown_timeout = if project.is_production {
+            std::time::Duration::from_secs(10)
+        } else {
+            std::time::Duration::from_secs(2)
+        };
+
+        let shutdown_future = graceful.shutdown();
```
DatGuyJonathan, line 2432: Do we need to flush the producer after shutting down http connections? Wondering if there might be data going through within the 10/2 timeouts above.

apps/framework-cli/src/infrastructure/processes/kafka_clickhouse_sync.rs
```diff
@@ -189,27 +200,131 @@ impl SyncingProcessesRegistry {
...
+            .map(|(sync_id, handle)| async move {
+                match tokio::time::timeout(grace_period, handle).await {
+                    Ok(Ok(Ok(()))) => {
+                        debug!("Topic-to-table sync process {} completed successfully", sync_id);
+                    }
+                    Ok(Ok(Err(e))) => {
+                        debug!("Topic-to-table sync process {} completed with error: {:?}", sync_id, e);
+                    }
+                    Ok(Err(join_err)) => {
+                        debug!("Topic-to-table sync process {} panicked: {:?}", sync_id, join_err);
```
DatGuyJonathan: Should these be error messages instead of debugs?

apps/framework-cli/src/infrastructure/processes/kafka_clickhouse_sync.rs
```diff
@@ -313,7 +431,15 @@ async fn sync_kafka_to_kafka(
     let target_topic_name = &target_topic_name;
 
     loop {
-        match subscriber.recv().await {
+        select! {
+            // Check for cancellation signal
+            _ = &mut cancel_rx => {
+                info!("Received cancellation signal for topic-to-topic sync: {} -> {}", source_topic_name, target_topic_name);
+                break;
```
DatGuyJonathan, line 474: Why does sync_kafka_to_clickhouse flush the inserter on cancellation but this one can just exit the loop?

apps/framework-cli/src/cli/local_webserver.rs
callicles: Makes sense - updated

apps/framework-cli/src/cli/local_webserver.rs
callicles: Good catch!

apps/framework-cli/src/infrastructure/processes/kafka_clickhouse_sync.rs
callicles: No Good reason - updated

phiSgr (approved):

PR #2872, opened 2025-10-17
moose migrate

DatGuyJonathan:
> [!NOTE]
> Introduce ClickHouse-backed state storage with planned migrations and new `moose migrate`, plus `--clickhouse-url` support across plan/generate, docs, tests, and templates.
> 
> - **CLI**:
>   - Add `migrate` command to execute planned migrations against ClickHouse with distributed lock and drift detection.
>   - Extend `plan` and `generate migration` with `--clickhouse-url` (serverless flow); support Moose server or direct ClickHouse.
>   - Telemetry: add `MigrateCommand` activity.
> - **Core/State**:
>   - Introduce `StateStorage` abstraction (Redis or ClickHouse) and project `state_config`.
>   - Refactor planning, dev/prod startup, watcher, and admin endpoints to use `StateStorage` and persist infra state in ClickHouse (`_MOOSE_STATE`).
> - **Migrations**:
>   - Implement migration executor with operation descriptions, partial-failure reporting, and schema drift checks (Current vs Expected vs Target).
>   - ClickHouse KeeperMap-based state table and migration lock.
> - **ClickHouse integration**:
>   - Add connection-string parser; enhance table introspection (LowCardinality) and operation descriptions.
>   - Docker compose: configure `keeper_map_path_prefix` for KeeperMap.
> - **Docs**:
>   - Update guides for serverless migrations, KeeperMap requirements, CLI usage, and troubleshooting (lock handling).
> - **Testing/Templates**:
>   - Add E2E tests for migration flow and new `typescript-migrate-test` template.
>   - Test constants include migration timeout.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 44957c4d1ac73459d9289d098abbab86778747e2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/cli/commands.rs
+++ b/apps/framework-cli/src/cli/commands.rs
@@ -62,12 +62,12 @@ pub enum Commands {
     /// to production, considering the current state of the project
     Plan {
         /// URL of the remote Moose instance (default: http://localhost:4000)
-        #[arg(long, conflicts_with = "clickhouse_url", requires = "token")]
+        #[arg(long, conflicts_with = "clickhouse_url")]
         url: Option<String>,
 
         /// API token for authentication with the remote Moose instance
         /// This token will be sent as a Bearer token in the Authorization header
-        #[arg(long, requires = "url")]
+        #[arg(long)]
         token: Option<String>,
 
         /// ClickHouse connection URL for serverless deployments
@@ -205,14 +205,13 @@ pub enum GenerateCommand {
         #[arg(
             long,
             conflicts_with = "clickhouse_url",
-            required_unless_present = "clickhouse_url",
-            requires = "token"
+            required_unless_present = "clickhouse_url"
         )]
         url: Option<String>,
 
         /// API token for authentication with the remote Moose instance
         /// This token will be sent as a Bearer token in the Authorization header
-        #[arg(long, requires = "url")]
+        #[arg(long)]
         token: Option<String>,
 
         /// ClickHouse connection URL for serverless deployments


apps/framework-cli/src/cli/routines/migrate.rs
```diff
@@ -0,0 +1,663 @@
...
+
+    Ok(MigrationFiles {
+        plan,
+        state_before,
+        state_after,
+    })
+}
+
+/// Check for drift between current, expected, and target states
+fn detect_drift(
```
callicles, line 114: Is this reproduciung the `check_reality` functionality we have?

apps/framework-cli/src/cli/routines/migrate.rs
```diff
@@ -0,0 +1,663 @@
...
+    }
+}
+
+/// Execute migration operations with detailed error handling
+async fn execute_operations(
+    project: &Project,
+    migration_plan: &MigrationPlan,
+    client: &ConfiguredDBClient,
+) -> Result<()> {
+    if !project.features.olap || migration_plan.operations.is_empty() {
```
callicles: If OLAP is not activated it looks like we should have a different error message

apps/framework-cli/src/framework/core/state_storage.rs
```diff
@@ -0,0 +1,200 @@
...
+    async fn ensure_state_table(&self) -> Result<()> {
+        let create_table_sql = format!(
+            r#"
+            CREATE TABLE IF NOT EXISTS `{}`.`{}`
+            (
+                key String,
+                value String,
+                updated_at DateTime DEFAULT now()
+            )
+            ENGINE = ReplacingMergeTree(updated_at)
```
callicles: Why did you choose ReplacingMergeTree? That looks like this is eventually consistent?

apps/framework-cli/src/framework/core/state_storage.rs
```diff
@@ -0,0 +1,200 @@
...
+            .context("Failed to deserialize infrastructure map from protobuf")?;
+
+        info!("✓ Loaded infrastructure map from ClickHouse");
+
+        Ok(Some(infra_map))
+    }
+}
+
+/// Creates the appropriate state storage implementation based on project configuration
+pub async fn create_state_storage(
```
callicles: This looks a bit weird.

Looks like input could contracdtic one another, in serverless migrations there is no redis, so this should not have a redis client no?

apps/framework-cli/src/project.rs
```diff
@@ -207,6 +229,9 @@ pub struct Project {
     /// Temporal workflow configuration
     #[serde(default)]
     pub temporal_config: TemporalConfig,
+    /// State storage configuration
+    #[serde(default)]
+    pub state_config: StateConfig,
```
callicles, line 237: shouldn't that be derieved if no redis config?

hh2110:
I've just had a thought... our platform allows our customers to edit their TTL expressions according to their needs. These edits can happen any time which means that for some of our customers, in between `moose plan`, the PR being reviewed and finally `moose migrate` a customer may have changed their TTL and so when `moose migrate` is run, it will abort because the remote state of the db would have changed ... could there an option to ignore TTL changes please with maybe a flag? ... I'm trying to think of other scenarios in which this sort of thing can happen but TTLs feel quite unique 

I'm thinking I may have shot myself in the foot with the ask around ttl-expressions ... if i don't set these in the olap config, then will moose realise this and try to remove ttl expressions on tables in a migration -- I guess I can always remove such changes from the plan. But i think the issue around aborting a migrate command due to changes in the remote db still apply

apps/framework-cli/src/cli/routines/migrate.rs
DatGuyJonathan: Good catch, updated!

apps/framework-cli/src/framework/core/state_storage.rs
DatGuyJonathan: I was trying to keep one infra map but didn't want to delete then insert since we could be left without an infra map. I added "final" to use with replacingmergetree which should fix the eventual consistency. 

Would a history of infra maps be useful someday (and we just query the latest for this)?

apps/framework-cli/src/framework/core/state_storage.rs
DatGuyJonathan: Agreed, I did some refactoring

apps/framework-cli/src/cli/routines/migrate.rs
DatGuyJonathan: We're using the existing check_reality through reconcile_with_reality - same flow for moose prod & moose migrate. The adds extra safety checks because it's applying a pre-generated plan.

apps/framework-cli/src/project.rs
DatGuyJonathan: Good question. The Redis config is always present (has defaults), so we'd need to check if Redis is actually reachable to derive this. That's risky though - if Redis is temporarily down or auth fails, we'd silently fall back to ClickHouse when the user expected Redis. Felt safer to keep it explicit so there's no ambiguity about which backend is being used.

apps/framework-cli/src/framework/core/state_storage.rs
callicles: History would definitely be useful, that would be the audit trail we have been talking about

callicles: But the biggest Issue I have with this is the insert are async. ie when you run insert you are not guaranteed to get a select right after.

So if multiple moose start at the same time, then we don't have a lock on who needs to apply the migration

callicles: https://clickhouse.com/docs/engines/table-engines/special/keeper-map

DatGuyJonathan: Nice, that looked like a good fit. I updated the pr with that (needed an extra config for local ch but the engine worked out of the box in clickhouse cloud)

apps/framework-cli/src/framework/core/state_storage.rs
```diff
@@ -0,0 +1,390 @@
...
+            .await
+            .context("Failed to create state table")?;
+
+        Ok(())
+    }
+
+    /// Try to acquire migration lock
+    /// Must be manually released with release_migration_lock()
+    /// Lock automatically expires after 5 minutes as a safety fallback
+    pub async fn acquire_migration_lock(&self) -> Result<()> {
```
callicles, line 113: Ohhhh - Would be so good to have that on the redis side too. 

Not a blocker but that would enable to make the Redis Client a lot more stateless. No more subscribing to redis and keep alive.

apps/framework-cli/src/framework/core/state_storage.rs
```diff
@@ -0,0 +1,390 @@
...
+                    existing_lock.machine_id, existing_lock.expires_at
+                );
+            } else {
+                // Active lock held by someone else
+                let time_remaining = existing_lock.expires_at - Utc::now();
+                let minutes = time_remaining.num_minutes();
+                let seconds = time_remaining.num_seconds() % 60;
+
+                anyhow::bail!(
+                    "Migration already in progress on machine {}. Started at {}. Expires in {}m {}s.",
```
callicles, line 180: In the docs we need a way for someone to manually delete a lock - it happens with terraform when the lock is fudged. You can manually unlock

Can be a follow up

DatGuyJonathan:
> I've just had a thought... our platform allows our customers to edit their TTL expressions according to their needs. These edits can happen any time which means that for some of our customers, in between `moose plan`, the PR being reviewed and finally `moose migrate` a customer may have changed their TTL and so when `moose migrate` is run, it will abort because the remote state of the db would have changed ... could there an option to ignore TTL changes please with maybe a flag? ... I'm trying to think of other scenarios in which this sort of thing can happen but TTLs feel quite unique
> 
> I'm thinking I may have shot myself in the foot with the ask around ttl-expressions ... if i don't set these in the olap config, then will moose realise this and try to remove ttl expressions on tables in a migration -- I guess I can always remove such changes from the plan. But i think the issue around aborting a migrate command due to changes in the remote db still apply

@hh2110  Thanks for sharing! Love that use case. I'm going to put up another PR for ignoring TTL diffs since this PR is getting massive :grin: I'll tag on you on that one for viz.

PR #2881, opened 2025-10-20
deep clone requests when stripping the urls for proxied requests

onelesd:
> [!NOTE]
> Use prototype-preserving cloning of `req` when rewriting URLs so headers and stream methods are retained for web app and API handlers.
> 
> - **Backend Router (`packages/ts-moose-lib/src/consumption-apis/runner.ts`)**:
>   - **Request cloning**: Replace shallow `{ ...req }` with `Object.assign(Object.create(Object.getPrototypeOf(req)), req, { url: ... })` when rewriting URLs.
>     - Applies to WebApp routing before calling `webApp.handler(modifiedReq, res)`.
>     - Applies to API routing before calling `apiRequestHandler(modifiedReq, res)`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3834f69cb3698f621292f94be853d82133dd6418. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli-e2e/test/utils/file-utils.ts
+++ b/apps/framework-cli-e2e/test/utils/file-utils.ts
@@ -4,11 +4,22 @@ import * as os from "os";
 import { randomUUID } from "crypto";
 
 /**
- * Removes a test project directory
+ * Removes a test project directory with retry logic for race conditions
  */
 export const removeTestProject = (dir: string): void => {
   console.log(`Deleting ${dir}`);
-  fs.rmSync(dir, { recursive: true, force: true });
+
+  try {
+    fs.rmSync(dir, {
+      recursive: true,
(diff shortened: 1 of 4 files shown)


callicles (approved):

packages/ts-moose-lib/src/consumption-apis/runner.ts
```diff
@@ -328,13 +328,15 @@ const createMainRouter = async (
         }
 
         try {
-          await webApp.handler(
+          // Create a modified request preserving all properties including headers
```
phiSgr, line 331: i think a comment here should explain why we clone in this particular way.

PR #2888, opened 2025-10-22
mitchell/chore-ci: builds

jmsuzuki:
**chore: updated release scripts for ci builds**
ci builds are not published as latest

**chore: release-cli gha workflow supports ci builds**
the long awaited feature... drum roll please!
.
.
.
ci builds!

---

> [!NOTE]
> Adds CI build flow with prerelease versioning, routes artifacts to dev channel, and publishes npm packages with a dev tag instead of latest; release builds remain unchanged.
> 
> - **GHA Workflow (`.github/workflows/release-cli.yaml`)**
>   - Trigger on all branches; full fetch with tags.
>   - New versioning step outputs `VERSION`, `PYTHON_VERSION` (PEP 440), and `IS_CI_BUILD`.
>   - CI builds: derive version via `git describe` (exclude `-ci-` tags); mark GitHub releases as prerelease without notes; upload version to `dev/<version>/` and use dev cache headers.
>   - Release builds: unchanged tagging and stable uploads; update `stable/latest/`.
>   - Python lib publish uses `python_version`; only runs for release builds.
>   - Binary build: set Cargo to base version, inject full version for binary via `CARGO_PKG_VERSION`, set Python `pyproject` version, and use `TAG_LATEST` env.
>   - GCS paths, cache headers, and index channel selection now depend on `is_ci_build`.
>   - `release-python` and `cleanup-pypi` gated to release builds only.
>   - Fix waits for correct npm package names.
> - **Publish Scripts**
>   - `apps/create-moose-app`, `apps/moose-cli-npm` (bin/cli), and multiple `packages/*/scripts/release*.sh` publish with `--tag dev` when `TAG_LATEST=false`; otherwise publish normally.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d258bde8de48467518fe959361585917cd6f3739. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/release-cli.yaml
+++ b/.github/workflows/release-cli.yaml
@@ -615,7 +615,7 @@ jobs:
 
       - name: Wait for @514labs/moose-cli-darwin-x64 to be available
         shell: bash
-        run: ./scripts/wait-for-npm-package.sh @514labs/moose-cli-linux-x64 ${{ needs.version.outputs.version }}
+        run: ./scripts/wait-for-npm-package.sh @514labs/moose-cli-darwin-x64 ${{ needs.version.outputs.version }}
 
       - name: Wait for @514labs/moose-cli-darwin-arm64 to be available
         shell: bash
@@ -623,7 +623,7 @@ jobs:
 
       - name: Wait for @514labs/moose-cli-linux-arm64 to be available
         shell: bash
-        run: ./scripts/wait-for-npm-package.sh @514labs/moose-cli-linux-x64 ${{ needs.version.outputs.version }}
+        run: ./scripts/wait-for-npm-package.sh @514labs/moose-cli-linux-arm64 ${{ needs.version.outputs.version }}
 
       - name: Wait for @514labs/moose-cli-linux-x64 to be available
         shell: bash


.github/workflows/release-cli.yaml
```diff
@@ -32,23 +32,66 @@ jobs:
       id-token: "write"
     outputs:
       version: ${{ steps.version.outputs.VERSION }}
+      python_version: ${{ steps.version.outputs.PYTHON_VERSION }}
```
jmsuzuki, line 35: python versions must follow pep440 vs semver.

.github/workflows/release-cli.yaml
```diff
@@ -3,7 +3,7 @@ name: Release CLI
 on:
   push:
     branches:
-      - main
+      - '**'  # All branches
```
DatGuyJonathan, line 6: Could anyone create a branch and get a preview release to distribute? Should we check if they're part of an org?

.github/workflows/release-cli.yaml
DatGuyJonathan: I think our normal python releases have a limit of 100 versions. Do these python review releases contribute to that count?

jmsuzuki: Most likely.

.github/workflows/release-cli.yaml
jmsuzuki: Yes, anyone who creates a branch will get CI builds. As an OSS, I think we want that for anyone.

.github/workflows/release-cli.yaml
callicles: Could we send dev version to another registry? Google artifact? so that it doesn't count?

.github/workflows/release-cli.yaml
callicles: Yeah that's a good call - since this is a public repo we should have that not distribute things to a easily public release - could be fine to a public Google artefact registry since someone would need to configure that expliciely to get access

.github/workflows/release-cli.yaml
```diff
@@ -543,15 +615,15 @@ jobs:
 
       - name: Wait for @514labs/moose-cli-darwin-x64 to be available
         shell: bash
-        run: ./scripts/wait-for-npm-package.sh @514labs/moose-cli-linux-x64 ${{ needs.version.outputs.version }}
+        run: ./scripts/wait-for-npm-package.sh @514labs/moose-cli-darwin-x64 ${{ needs.version.outputs.version }}
```
callicles, line 618: Lol good catch!
Copy pasta....

jmsuzuki: great catch by one of the bots. 🍷

.github/workflows/release-cli.yaml
jmsuzuki: so... why make the ci builds private? this seems counterintuitive towards getting community involvement.

as an OSS moose developer, I'd like to have access to these builds as well so that I can test them in my projects.
as an OSS moose developer, I'm probably making updates to the moose-cli, so that I can have the shiny new thing in my moose project.

these are CI / pre-lease builds. they don't tie to "latest".

callicles: I am not saying make them private. 
I am saying push them to another repo so that end users don't see them and can't install them by mistake and we don't use more of our quota for hisotry which is already quite small

jmsuzuki: users should be unable to install by mistake. they will have to specify the version they want.
fuzzy match on versions and/or latest will not result in a pre-release version being installed.

if the real concern here is our quota, that limit can be increased.
I have an open support issue to have our limit increased from 10GB to 50GB.
https://github.com/pypi/support/issues/7991

callicles: If they can increase then awesome.
At some point they wanted to allow orgs to pay for space which we are happy to do. But we are still on the waiting list I believe

jmsuzuki: okay. cool. so, let's hold on to this for a minute to see what happens with the request.
if they don't approve it in the next week or so, I'll look at using Google Artifact Registry. I believe there's a viable path there

.github/workflows/release-cli.yaml
jmsuzuki: we can... will look in to that solution if pypi support doesn't respond timely.

.github/workflows/release-cli.yaml
jmsuzuki: we could also do this in 2 pieces.
1. add support for TS CI builds today.
2. add support for python CI builds in the future

callicles: YEah I think splitting it makes sense.
Ship what works and wait on what we don't have

.github/workflows/release-cli.yaml
```diff
@@ -474,8 +544,8 @@ jobs:
   release-python:
     name: Release Python Wheels
     runs-on: ubuntu-latest
-    needs: [build-and-publish-binaries, cleanup-pypi]
-    if: ${{ !inputs.dry-run }}
+    needs: [version, build-and-publish-binaries, cleanup-pypi]
+    if: ${{ !inputs.dry-run && needs.version.outputs.is_ci_build == 'false' }}
```
jmsuzuki, line 548: @callicles python releases are disabled

.github/workflows/release-cli.yaml
jmsuzuki: cool. done.

PR #2901, opened 2025-10-24
config to ignore certain migration operations

DatGuyJonathan:
> [!NOTE]
> Adds project-level config to ignore TTL-related migration operations and drift, updates planning/migration logic to honor it, and documents usage.
> 
> - **Migration planning/execution**:
>   - Add `IgnorableOperation` enum (`ModifyTableTtl`, `ModifyColumnTtl`) and matching logic.
>   - Extend `MigrationPlan` with `filter_ignored_operations(...)` and apply it in `remote_gen_migration` based on `project.migration_config.ignore_operations`.
>   - Add `strip_ignored_fields(...)` and update drift detection to strip metadata and ignored fields; adjust tests and add TTL-specific cases.
> - **Project config**:
>   - Introduce `MigrationConfig { ignore_operations: Vec<IgnorableOperation> }` on `Project`; wire defaults in plan/reality checker constructors.
> - **Docs**:
>   - Document `[migration_config]` with `ignore_operations = ["ModifyTableTtl", "ModifyColumnTtl"]` in configuration and migration guides (incl. TTL management section).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9204d467cecd84c90c807fbab6376b68a9c94fad. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/infrastructure/olap/clickhouse.rs
+++ b/apps/framework-cli/src/infrastructure/olap/clickhouse.rs
@@ -193,6 +193,28 @@ pub enum SerializableOlapOperation {
     },
 }
 
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[serde(rename_all = "PascalCase")]
+pub enum IgnorableOperation {
+    ModifyTableTtl,
+    ModifyColumnTtl,
+}
+
+impl IgnorableOperation {
+    pub fn matches(&self, op: &SerializableOlapOperation) -> bool {
+        matches!(
+            (self, op),
+            (
+                Self::ModifyTableTtl,
+                SerializableOlapOperation::ModifyTableTtl { .. }
+            ) | (
+                Self::ModifyColumnTtl,
+                SerializableOlapOperation::ModifyColumnTtl { .. }
+            )
+        )
+    }
+}
+
 /// Executes a series of changes to the ClickHouse database schema
 ///
 /// # Arguments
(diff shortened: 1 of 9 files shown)


apps/framework-cli/src/infrastructure/olap/clickhouse.rs
```diff
@@ -193,6 +193,28 @@ pub enum SerializableOlapOperation {
...
 
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[serde(rename_all = "PascalCase")]
+pub enum IgnorableOperation {
+    ModifyTableTtl,
+    ModifyColumnTtl,
+}
+
+impl IgnorableOperation {
+    pub fn matches(&self, op: &SerializableOlapOperation) -> bool {
```
callicles, line 204: Why do you want to limit what the user wants to do?

callicles (commented):
I would also add the filtering at the input of the ordering of opertations before the execution so that no matter what they get filtered out

phiSgr (approved):

apps/framework-cli/src/infrastructure/olap/clickhouse.rs
DatGuyJonathan: Mainly because there's specific implementations for ignoring different operations, so this was a way to get ignoring ttl out then we can iteratively add others.

DatGuyJonathan:
> I would also add the filtering at the input of the ordering of opertations before the execution so that no matter what they get filtered out

Yea, I believe `remote_gen_migration` is already filtering it out so the plan file won't have it either. Lmk if that's not what you meant

DatGuyJonathan:
Going to merge so @phiSgr can rebase his pr. Will do follow-up prs if there's anything else.

fyi @callicles

PR #2911, opened 2025-10-28
ENG-1122: fix mv not populating new mv on update

LucioFranco:
> [!NOTE]
> Materialized view population now runs only in dev (with optional truncate) and the production flag is threaded through diffing/planning; endpoints, DDL generation, and tests updated accordingly.
> 
> - **Olap/Infra Diffing**:
>   - Extend `InfrastructureMap::diff_with_table_strategy` to accept `is_production` and propagate from callers (`local_webserver` admin plan, planning, routines).
>   - Update `diff_sql_resources` to accept `is_production` and invoke `ClickHouseTableDiffStrategy::check_materialized_view_population` with it.
> - **ClickHouse Strategy**:
>   - `check_materialized_view_population` now skips population in production; triggers population for new/updated MVs; emits `OlapChange::PopulateMaterializedView { should_truncate }`.
> - **DDL Generation**:
>   - `AtomicOlapOperation::PopulateMaterializedView` and serialization now support `should_truncate`, emitting `TRUNCATE TABLE` before `INSERT` when set.
> - **Planning/CLI**:
>   - Pass `project.is_production` into diffing in `plan.rs` and admin plan route; local diff (`calculate_plan_diff_local`) updated likewise.
> - **Tests**:
>   - Adjust tests for new method signatures and new MV population behavior; add tests ensuring truncate + insert order and MV update triggers backfill.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a6860eddf3082fce1f5681330f4f110a90e9e52f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/cli/routines.rs
+++ b/apps/framework-cli/src/cli/routines.rs
@@ -864,7 +864,7 @@ fn calculate_plan_diff_local(
 
     let clickhouse_strategy = ClickHouseTableDiffStrategy;
     // planning about action on prod env, respect_life_cycle is true
-    current_map.diff_with_table_strategy(target_map, &clickhouse_strategy, true)
+    current_map.diff_with_table_strategy(target_map, &clickhouse_strategy, true, true)
 }
 
 /// Legacy implementation of remote_plan using the existing /admin/plan endpoint
--- a/apps/framework-cli/src/framework/core/infrastructure_map.rs
+++ b/apps/framework-cli/src/framework/core/infrastructure_map.rs
@@ -382,6 +382,8 @@ pub enum OlapChange {
         select_statement: String,
         /// Source tables that data is pulled from
         source_tables: Vec<String>,
+        /// Whether to truncate the target table before populating
+        should_truncate: bool,
     },
 }
 
@@ -885,6 +887,7 @@ impl InfrastructureMap {
         target_map: &InfrastructureMap,
         table_diff_strategy: &dyn TableDiffStrategy,
         respect_life_cycle: bool,
+        is_production: bool,
     ) -> InfraChanges {
         let mut changes = InfraChanges::default();
 
@@ -934,6 +937,7 @@ impl InfrastructureMap {
             &self.sql_resources,
             &target_map.sql_resources,
             &target_map.tables, // Pass target tables for MV analysis
+            is_production,
             &mut changes.olap_changes,
         );
         let sql_resource_changes = changes.olap_changes.len() - olap_changes_len_before;
@@ -1556,11 +1560,13 @@ impl InfrastructureMap {
     /// * `self_sql_resources` - HashMap of source SQL resources to compare from
     /// * `target_sql_resources` - HashMap of target SQL resources to compare against
     /// * `target_tables` - Target tables for MV population analysis
+    /// * `is_production` - Whether running in production environment
     /// * `olap_changes` - Mutable vector to collect the identified changes
     pub fn diff_sql_resources(
         self_sql_resources: &HashMap<String, SqlResource>,
         target_sql_resources: &HashMap<String, SqlResource>,
         target_tables: &HashMap<String, Table>,
+        is_production: bool,
         olap_changes: &mut Vec<OlapChange>,
     ) {
         log::info!(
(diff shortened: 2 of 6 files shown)


apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -1583,6 +1564,15 @@ impl InfrastructureMap {
                         before: Box::new(sql_resource.clone()),
                         after: Box::new(target_sql_resource.clone()),
                     }));
+
+                    // Check if updated SQL resource is a materialized view that needs population
+                    use crate::infrastructure::olap::clickhouse::diff_strategy::ClickHouseTableDiffStrategy;
```
callicles, line 1594: Isn't this going to have the consequence of duplicatign data in the target if the data has already got through and you are iterating on the MV?

I believe we should have it as if we re-populate then we need to also truncate the target table.

LucioFranco: Yeah, you are right. One idea could be to always truncate the target table so that would noop for a new MV and for an updated MV it would truncate and refill. Otherwise, I imagine we need to find a way to reconcile the two tables.

callicles: Yeah that's what I am thinking

LucioFranco: Should be solved, we now insert truncates before the insert commands. Ive also added some unit tests.

apps/framework-cli/src/infrastructure/olap/ddl_ordering.rs
```diff
@@ -267,17 +274,23 @@ impl AtomicOlapOperation {
...
+                        format!("TRUNCATE TABLE `{}`.`{}`", database, target_table),
+                        format!(
+                            "INSERT INTO `{}`.`{}` {}",
+                            database, target_table, select_statement
+                        ),
                     )
                 } else {
-                    format!("INSERT INTO `{}` {}", target_table, select_statement)
+                    (
+                        format!("TRUNCATE TABLE `{}`", target_table),
```
callicles: ok so that works great for dev, but I don't think we can truncate data in prod. I think we need to tests if dev or prod. 

Also we need to make sure this gets displayed in the plan when a user runs it. As such we have been trying to stay away from adding operations in DDL ordering, it should be mostly ordering operations, Not adding some. diff strategy in clickhouse has been I believe where we want to add that

LucioFranco: Ok updated to check if we are in dev/prod, added when tables are going to be truncated to display in the plan but I did not totally move the truncate out of the ddl_ordering, not 100% clear on what this means but I think the current state is good enough to merge now and fix in a follow up. We are mostly covered with the dev check like you suggested.

apps/framework-cli/src/infrastructure/olap/clickhouse/diff_strategy.rs
```diff
@@ -185,10 +186,11 @@ impl ClickHouseTableDiffStrategy {
                         .is_some_and(Self::is_s3queue_table)
                 });
 
-                // Only populate new MVs with non-S3Queue sources
-                if is_new && !has_s3queue_source {
+                // Skip population in production (user must handle manually)
```
phiSgr, line 225: i'm not sure about this assumption

apps/framework-cli/src/cli/routines.rs
```diff
@@ -872,18 +872,20 @@ pub(crate) async fn get_remote_inframap_protobuf(
...
     current_map: &InfrastructureMap,
     target_map: &InfrastructureMap,
+    is_production: bool,
 ) -> crate::framework::core::infrastructure_map::InfraChanges {
     use crate::infrastructure::olap::clickhouse::diff_strategy::ClickHouseTableDiffStrategy;
 
     let clickhouse_strategy = ClickHouseTableDiffStrategy;
     // planning about action on prod env, respect_life_cycle is true
-    current_map.diff_with_table_strategy(target_map, &clickhouse_strategy, true)
+    current_map.diff_with_table_strategy(target_map, &clickhouse_strategy, true, is_production)
```
phiSgr: it should probably also be true?

PR #2915, opened 2025-10-29
feat: add infra diagnostics tool to the Moose Dev MCP

callicles:
> [!NOTE]
> Adds a new MCP tool `diagnose_infrastructure` with multiple ClickHouse diagnostics and wires it into the server alongside tests and docs updates.
> 
> - **MCP Server**:
>   - Registers new tool `diagnose_infrastructure` in `list_tools` and routes in `call_tool`.
>   - Updates server instructions to mention diagnostics and source browsing.
>   - Updates tests to account for 6 tools and validates tool names.
> - **New Tool: `infra_issues` (ClickHouse diagnostics)**:
>   - Core module `infra_issues/mod.rs` with params parsing, severity filtering, Redis-backed infra map usage, and aggregated JSON output with summaries.
>   - Diagnostic providers:
>     - `MutationDiagnostic` (`system.mutations`): stuck/failed mutations.
>     - `PartsDiagnostic` (`system.parts`): excessive parts per partition.
>     - `MergeDiagnostic` (`system.merges`): long-running merges.
>     - `ErrorStatsDiagnostic` (`system.errors`): top recurring errors (system-wide).
>     - `S3QueueDiagnostic` (`system.s3queue_log`): ingestion failures (S3Queue tables).
>     - `ReplicationDiagnostic` (`system.replication_queue`/`system.replicas`): lag and health issues (replicated tables).
>     - `MergeFailureDiagnostic` (`system.metrics`): background merge failures (system-wide).
>     - `StoppedOperationsDiagnostic`: merges/replication possibly stopped.
> - **Tool Registry**:
>   - Exposes `infra_issues` in `tools/mod.rs`.
> - **Misc**:
>   - `get_source.rs`: remove unused `SourceLocationNotAvailable` error variant.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7c40c3662778b8bc0e9aa91e77a62dd3944e851e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
+++ b/apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
@@ -1,0 +1,858 @@
+//! # Infrastructure Issues Diagnostic Tool
+//!
+//! This module implements a proactive infrastructure diagnostics MCP tool that intelligently
+//! surfaces errors by using the infrastructure map to determine what to check based on
+//! infrastructure type (ClickHouse, Kafka/Redpanda, Temporal, etc.).
+//!
+//! Initial implementation focuses on ClickHouse diagnostics with extensible architecture
+//! for future infrastructure types.
+//!
+//! ## ClickHouse Diagnostic Providers
+//!
+//! The tool automatically runs multiple diagnostic providers based on table engine types:
+//!
+//! ### 1. MutationDiagnostic
+//! Detects stuck or failing mutations (ALTER operations) that can block table maintenance.
+//! - **Source**: `system.mutations`
+//! - **Detection**: Mutations not done (is_done = 0) or with non-empty failure reasons
+//! - **Thresholds**:
+//!   - Error: Mutation has a failure reason (latest_fail_reason not empty)
+//!   - Warning: Mutation in progress but not completed (is_done = 0)
+//! - **Suggested Action**: Cancel stuck mutations with KILL MUTATION
+//!
+//! ### 2. PartsDiagnostic
+//! Identifies excessive data parts per partition that impact query performance.
+//! - **Source**: `system.parts`
+//! - **Detection**: Active parts count per partition > 100
+//! - **Thresholds**:
+//!   - Error: part_count > 300
+//!   - Warning: 100 < part_count ≤ 300
+//! - **Suggested Action**: Run OPTIMIZE TABLE to merge parts
+//!
+//! ### 3. MergeDiagnostic
+//! Monitors long-running background merges.
+//! - **Source**: `system.merges`
+//! - **Detection**: Merges running > 300 seconds
+//! - **Thresholds**:
+//!   - Error: elapsed_time > 1800s (30 minutes)
+//!   - Warning: 300s < elapsed_time ≤ 1800s
+//! - **Note**: Progress is tracked and reported but not used in severity determination
+//! - **Suggested Action**: Monitor merge progress and check server resources (CPU, disk I/O, memory)
+//!
+//! ### 4. ErrorStatsDiagnostic
+//! Aggregates errors from ClickHouse system.errors to surface recurring issues.
+//! - **Source**: `system.errors`
+//! - **Detection**: All errors with count > 0 (reports top 10 by occurrence)
+//! - **Thresholds**:
+//!   - Error: error_count > 100
+//!   - Warning: error_count > 10
+//!   - Info: 0 < error_count ≤ 10
+//! - **Suggested Action**: Review error messages and recent system changes
+//!
+//! ### 5. S3QueueDiagnostic (S3Queue tables only)
+//! Detects S3Queue ingestion failures and processing issues.
+//! - **Source**: `system.s3queue_log`
+//! - **Detection**: Failed or ProcessingFailed status entries in S3Queue log
+//! - **Threshold**: All failed entries trigger Error severity
+//! - **Suggested Action**: Check S3 credentials, permissions, and file formats
+//!
+//! ### 6. ReplicationDiagnostic (Replicated* tables only)
+//! Monitors replication health, queue backlogs, and stuck replication entries.
+//! - **Sources**: `system.replication_queue`, `system.replicas`
+//! - **Detection**:
+//!   - Large queue backlogs (queue_size > 10 or > 100 for replicas health)
+//!   - Stuck entries (num_tries > 3 or has exceptions)
+//!   - Replica health issues (readonly, session_expired, high delay > 300s)
+//! - **Thresholds**:
+//!   - Error: queue_size > 50, num_tries > 10, session_expired, delay > 600s
+//!   - Warning: queue_size > 10, 3 < num_tries ≤ 10, readonly, 300s < delay ≤ 600s
+//! - **Suggested Action**: Check ZooKeeper connectivity, restart replication queues
+//!
+//! ### 7. MergeFailureDiagnostic
+//! Detects system-wide background merge failures that may affect multiple tables.
+//! - **Source**: `system.metrics`
+//! - **Detection**: FailedBackgroundMerges metric > 0
+//! - **Thresholds**:
+//!   - Error: failed_merges > 10
+//!   - Warning: failed_merges > 0
+//! - **Suggested Action**: Check system.errors for merge failure details, review disk space
+//!
+//! ### 8. StoppedOperationsDiagnostic
+//! Identifies manually stopped or stalled merge/replication operations.
+//! - **Sources**: `system.parts`, `system.merges`, `system.replicas`
+//! - **Detection**:
+//!   - Many parts (>100) but no active merges
+//!   - Replica readonly with pending queue items
+//! - **Thresholds**:
+//!   - Error: Replica readonly with queue items (replication stopped)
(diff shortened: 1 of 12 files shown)


onelesd:
can you explain a bit the tools added and how they interact, perhaps with an example use case?

actually, just pasting the output of a session with tool calls would be illuminating

callicles:
@onelesd Did you take a look at the issue? there is a bunch of context there

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
```diff
@@ -0,0 +1,849 @@
+//! # Infrastructure Issues Diagnostic Tool
+//!
+//! This module implements a proactive infrastructure diagnostics MCP tool that intelligently
+//! surfaces errors by using the infrastructure map to determine what to check based on
+//! infrastructure type (ClickHouse, Kafka/Redpanda, Temporal, etc.).
+//!
+//! Initial implementation focuses on ClickHouse diagnostics with extensible architecture
+//! for future infrastructure types.
+//!
+//! ## ClickHouse Diagnostic Providers
```
LucioFranco, line 10: @callicles what did you do to get this list? Did you have the llm crawl the ch docs or did you curate these?

callicles: I prompted my way through this with problems I saw happening in the past

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
```diff
@@ -0,0 +1,849 @@
...
+//! Aggregates errors from ClickHouse system.errors to surface recurring issues.
+//! - **Source**: `system.errors`
+//! - **Detection**: All errors with count > 0 (reports top 10 by occurrence)
+//! - **Thresholds**:
+//!   - Error: error_count > 100
+//!   - Warning: error_count > 10
+//!   - Info: 0 < error_count ≤ 10
+//! - **Suggested Action**: Review error messages and recent system changes
+//!
+//! ### 5. S3QueueDiagnostic (S3Queue tables only)
```
LucioFranco, line 52: We probably want this for s3cluster as well? And/or drop s3 queue all together?

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
LucioFranco: Okay cool def a really good starting point, I'd like to expand on these since I think there are some missing from what I saw, but those can be added later.

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
callicles: s3Cluster is a synchronous function, it doesn't have the same caveats.
The S3 Table engine probably has some similar caveats alhough I don't think it will do anything untill you use it in a query so the feedback is actually synchronous there.

callicles: ie to test

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
```diff
@@ -0,0 +1,849 @@
...
+pub use merges::MergeDiagnostic;
+pub use mutations::MutationDiagnostic;
+pub use parts::PartsDiagnostic;
+pub use replication::ReplicationDiagnostic;
+pub use s3queue::S3QueueDiagnostic;
+pub use stopped_operations::StoppedOperationsDiagnostic;
+
+/// Maximum number of log lines to retrieve from Docker
+#[allow(dead_code)] // Will be used when DockerLogsDiagnostic is implemented
+const MAX_DOCKER_LOG_LINES: u32 = 100;
```
LucioFranco: This is unused we can just drop this for now?

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
```diff
@@ -0,0 +1,849 @@
...
+    /// Minimum severity level to report
+    pub severity: Severity,
+    /// Optional time filter (e.g., "-1h" for last hour)
+    pub since: Option<String>,
+}
+
+/// Component information for issue context
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Component {
+    #[serde(rename = "type")]
```
LucioFranco: don't think this rename is worth it imo

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
```diff
@@ -0,0 +1,849 @@
...
+    pub component_type: String,
+    pub name: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub database: Option<String>,
```
LucioFranco: this seems pretty specific to databases while we may have providers down the line that are not databases. Maybe we actually just want some sort of grab bag like a `HashMap<String, String>` that can carry any additional metadata that the providers can try to extract and use.

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
```diff
@@ -0,0 +1,849 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DiagnosticOutput {
+    pub infrastructure_type: InfrastructureType,
+    pub issues: Vec<Issue>,
+    pub summary: IssueSummary,
+}
+
+impl DiagnosticOutput {
+    /// Create a new diagnostic output and compute summary statistics
+    fn new(infrastructure_type: InfrastructureType, issues: Vec<Issue>) -> Self {
```
LucioFranco: nit: kinda odd to me that the fields are public but its constructor is not public...maybe the fields need to be pub(crate)?

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
```diff
@@ -0,0 +1,849 @@
...
+
+/// Trait for diagnostic providers
+/// Each provider implements checks for a specific aspect of infrastructure health
+#[async_trait::async_trait]
+pub trait DiagnosticProvider: Send + Sync {
+    /// Name of this diagnostic provider
+    fn name(&self) -> &str;
+
+    /// Check if this provider is applicable to the given component
+    fn applicable_to(&self, component: &Component, engine: Option<&ClickhouseEngine>) -> bool;
```
LucioFranco, line 328: is this meant to hardcode to clickhouse? Reading the doc comments it shouldn't be

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
```diff
@@ -0,0 +1,849 @@
...
+    }
+
+    /// Run diagnostics and return list of issues found
+    async fn diagnose(
+        &self,
+        component: &Component,
+        engine: Option<&ClickhouseEngine>,
+        config: &ClickHouseConfig,
+        since: Option<&str>,
+    ) -> Result<Vec<Issue>, DiagnoseError>;
```
LucioFranco, line 343: I wonder if making this more stream based would be good for the ux say with a UI (webapp) or cli where this becomes `next_diagnostic` and it would then allow us to stream them rather than try to collect them all then return them. This is just an idea don't need this to be done here.

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
```diff
@@ -0,0 +1,849 @@
...
+                "examples": ["-1h", "-30m", "-1d", "2024-01-01T00:00:00Z"]
+            }
+        },
+        "required": ["infrastructure_type"]
+    });
+
+    Tool {
+        name: "diagnose_infrastructure".into(),
+        description: Some(
+            "Proactively diagnose infrastructure issues by intelligently checking relevant diagnostic sources based on infrastructure type. For ClickHouse, automatically checks: stuck mutations, S3Queue ingestion errors (for S3Queue tables), replication health (for replicated tables), data parts issues, background merge problems, system errors, and Docker container logs. Returns structured, actionable information about errors and warnings with suggested remediation steps.".into()
```
LucioFranco, line 390: This is a good start but the prompt feels almost too specific and I feel like we may get better results engineering this one better.

LucioFranco (approved):
LGTM, looking forward to giving this a try and improving on this! This is a really great start and I am quite happy because I think what is in here aligns really well with what we've been talking about!

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
callicles: Yep!

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
callicles: YEah we could explore that in an iteration

apps/framework-cli/src/mcp/tools/infra_issues/mod.rs
callicles: Let's iterate as we go

PR #2918, opened 2025-10-29
fix: make table diffing backwards compatible

phiSgr:
handle 
- old infra map where the key (table ID) does not contain db name
- `remote_plan` where the remote ID
  - is old (point 1)
  - contains another DB name

---

> [!NOTE]
> Make infra map diff/plan compatible with old table IDs (no DB prefix) and cross-DB IDs, add default_database to proto, and add E2E backward-compat tests in CI.
> 
> - **Core/Infra Map & Planning**:
>   - Use stable table IDs via `Table::id(default_database)` across diff/reality checks; add helper `find_table_from_infra_map` to match legacy IDs without DB prefix.
>   - `InfrastructureMap.diff_tables(_with_strategy)` now keys by IDs and uses target `default_database` for comparisons; filters adds/removes via new matching.
>   - `reconcile_with_reality` normalizes legacy map keys (rehashes IDs with DB) and updates `default_database` from project.
>   - Add `default_database` to proto (`InfrastructureMap`) and round-trip it in serialization.
> - **Partial Map (DMv2)**:
>   - Clear `primary_key` flags when engine doesn’t support ORDER BY; propagate table TTL/index settings; compute IDs with DB.
> - **Reality Checker**:
>   - Map actual tables by new ID format, compare with map after normalizing DB; compute missing/unmapped/mismatched using ID-aware logic; update tests accordingly.
> - **ClickHouse**:
>   - Include `database` in `TableWithUnsupportedType`; richer logging for column modify; warn (not debug) on drop+create cases in diff strategy; refine PK/engine/partition handling.
> - **Tests & CI**:
>   - Add E2E backward-compatibility tests (`backward-compatibility.test.ts`) and new GitHub Actions jobs for TS and Python; include them in final status gate.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7a2eb27a68560cfdb2b0143ca9afd456adfd2783. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-docs/src/pages/release-notes/2025-11-01.mdx
+++ b/apps/framework-docs/src/pages/release-notes/2025-11-01.mdx
@@ -13,7 +13,7 @@ import { Callout } from "@/components";
   - [S3 table engine](/moose/olap/model-table#direct-s3-access-s3): keep data in S3 while ClickHouse reads and writes it on demand ([ClickHouse docs](https://clickhouse.com/docs/en/engines/table-engines/integrations/s3))
   - Beta (self-hosted only): [Distributed table engine](/moose/olap/model-table#distributed-tables-distributed): serve cluster-wide queries through Moose-managed tables ([ClickHouse docs](https://clickhouse.com/docs/en/engines/table-engines/special/distributed))
 - **Improved:** Serverless migrations gain Redis-backed state storage plus per-table databases.
-- **Feature that will make a small number of people very happy:** `flake.nix` now ships a Moose devshell.
+- **Feature that will make a small number of people very happy:** Moose now has a flake.nix to let users install the cli via nix run github:514-labs/moosestack (also provides a dev shell to make contributing easier!)
 </Callout>
 
 ## Buffer tables for burst protection
@@ -128,6 +128,12 @@ PR: [#2876](https://github.com/514-labs/moosestack/pull/2876) | Docs: [Multi-dat
 - Fixed the redirect loop after deleting an organization so users land back on the create-org screen instead of bouncing between routes.
 
 ## Nix development environment
-`flake.nix` now bootstraps a full MooseStack toolchain (`nix develop` drops you into Rust, Node.js, and Python with shared test commands). If you use Nix, let us know!
+`flake.nix` now bootstraps a full MooseStack build environment (`nix develop` drops you into a development shell that will have everything you need to build all the components of moosestack). If you use Nix, let us know!
+
+Give it a go if you have nix installed on your machine with:
+
+```bash
+nix run github:514-labs/moosestack # -- to pass additional arguments to the cli
+```
 
 PR: [#2920](https://github.com/514-labs/moosestack/pull/2920)


apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -2228,7 +2235,7 @@ impl InfrastructureMap {
         let proto = ProtoInfrastructureMap::parse_from_bytes(&bytes)?;
 
         Ok(InfrastructureMap {
-            default_database: default_database_name(),
+            default_database: proto.default_database,
```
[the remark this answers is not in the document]
phiSgr: the only time we load and old version and use it it's through `reconcile_with_reality`, which sets the correct value

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -1711,6 +1715,8 @@ impl InfrastructureMap {
                             }
                         }
 
+                        // TODO: PARTITION BY is not checked!
```
callicles, line 1718: Is that left on purpose?

apps/framework-cli/src/framework/core/plan.rs
```diff
@@ -90,18 +91,36 @@ pub async fn reconcile_with_reality<T: OlapOperations>(
...
+    let mut reconciled_map = current_infra_map.clone();
+    reconciled_map.default_database = project.clickhouse_config.db_name.clone();
+
+    if current_infra_map
+        .tables
+        .iter()
+        .any(|(id, t)| id != &t.id(&project.clickhouse_config.db_name))
+    {
+        // fix up IDs where in the old version it does not contain the DB name
+        let existing_tables = mem::take(&mut reconciled_map.tables);
```
callicles, line 104: Why do you need this?

callicles (commented):
Can we get a test that shows that migrating from version n-1 to version n is not backward incompatible?

I am thinking we need to spin up the template test with the n-1 version of the cli before the commit and then spin it up after the commit and make sure the plan doesn't have changes

apps/framework-cli/src/framework/core/infrastructure_map.rs
phiSgr: yes, as a "by the way"

apps/framework-cli/src/framework/core/plan.rs
phiSgr: do you mean `mem::take` or the logic?

phiSgr:
See https://github.com/514-labs/moosestack/pull/2926 for test catching the existing bug.

apps/framework-cli/src/framework/core/plan.rs
callicles: mem::take

apps/framework-cli-e2e/test/backward-compatibility.test.ts
```diff
@@ -0,0 +1,467 @@
...
+  projectDir: string,
+  templateName: string,
+  appName: string,
+): Promise<void> {
+  console.log(`Initializing Python project with latest pypi moose-cli...`);
+
+  // Initialize project with latest CLI via npx
+  try {
+    const result = await execAsync(
+      `npx -y @514labs/moose-cli@latest init ${appName} ${templateName} --location "${projectDir}"`,
```
callicles, line 136: This doesn't look like using pypi, is that on purpose?

BACKWARD_COMPATIBILITY_TESTING.md
callicles, line 1: We should not commt this right?

apps/framework-cli/src/framework/core/plan.rs
phiSgr: to move contents of the table out of `reconciled_map.tables`

.github/workflows/test-framework-cli.yaml
```diff
@@ -554,6 +554,132 @@ jobs:
...
+          MOOSE_TELEMETRY_ENABLED: false
+          # Add test credentials for S3Queue tests
+          TEST_AWS_ACCESS_KEY_ID: "test-access-key-id"
+          TEST_AWS_SECRET_ACCESS_KEY: "test-secret-access-key"
+
+      - name: Inspect Logs
+        if: always()
+        run: |
+          cat ~/.moose/*-cli.log
+
```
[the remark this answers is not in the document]
phiSgr: nah, it's built in `pretest` script

callicles:
Screenshot 2025-11-02 at 11 50 03
We should merge but it looks like there is still some things we need to look into for the table diffing

PR #2926, opened 2025-10-30
Test backward compat 0.6.157

phiSgr:
the test should fail


phiSgr:
test failed successfully

PR #2939, opened 2025-11-03
ENG-902: Add `FixedString(N)` data type

LucioFranco:
> [!NOTE]
> Introduces end-to-end FixedString(N) type support (core, TS/Python SDKs, mappers, DDL, generators, validators) with new tests and template coverage.
> 
> - **Core/Type System**:
>   - Add `ColumnType::FixedString { length }` with serde, display, proto roundtrip, and type parsing (`FixedString(N)`).
> - **ClickHouse Mapping/DDL**:
>   - Map `FixedString` to `ClickHouseColumnType::FixedString`; generate `FixedString(N)` (incl. `Nullable`), and parse from CH types.
>   - Update enum-mapping and basic type printers to handle `FixedString`.
> - **Codegen/Runtime**:
>   - TS: add `FixedString<N>` type/tag; emit `string & FixedString<N>`; typeConvert detects `_clickhouse_fixed_string_size` and maps to `FixedString(N)`.
>   - Python: add `FixedString(size)` annotation (`ClickhouseFixedStringSize`); generator emits `Annotated[str, FixedString(N)]`.
>   - Streaming defaults treat `FixedString` like `String`.
> - **Validation/IO**:
>   - Passthrough validator enforces length for `FixedString`; CSV import and Kafka sync map as strings; Map key handling supports `FixedString`.
> - **Tests/Templates**:
>   - Add TS/Python unit tests for `FixedString` mappings and DDL; e2e template adds `FixedStringTest` table and DDL assertions (incl. type alias and arrays).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e309490f42035ecce9782d1819f7712464b2d2fa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/utilities/validate_passthrough.rs
+++ b/apps/framework-cli/src/utilities/validate_passthrough.rs
@@ -397,9 +397,7 @@ impl<'de, S: SerializeValue> Visitor<'de> for &mut ValueVisitor<'_, S> {
                 }
                 Err(Error::invalid_type(serde::de::Unexpected::Str(v), &self))
             }
-            ColumnType::String => {
-                self.write_to.serialize_value(v).map_err(Error::custom)
-            }
+            ColumnType::String => self.write_to.serialize_value(v).map_err(Error::custom),
             ColumnType::FixedString { length } => {
                 let byte_length = v.len() as u64;
                 if byte_length > *length {


apps/framework-cli/src/framework/python/generate.rs
```diff
@@ -58,6 +58,9 @@ fn map_column_type_to_python(
 ) -> String {
     match column_type {
         ColumnType::String => "str".to_string(),
+        ColumnType::FixedString { length } => {
+            format!("Annotated[bytes, FixedString({})]", length)
```
phiSgr: both `FixedString` and `String` in ClickHouse can store arbitrary bytes. If we are saying CH `FixedString` has to correspond to Python `bytes`, then the same should apply to CH `String` as well.

LucioFranco: Good call, I just moved fixedstring back to `str` to match `string`

apps/framework-cli/src/utilities/validate_passthrough.rs
```diff
@@ -395,7 +397,9 @@ impl<'de, S: SerializeValue> Visitor<'de> for &mut ValueVisitor<'_, S> {
                 }
                 Err(Error::invalid_type(serde::de::Unexpected::Str(v), &self))
             }
-            ColumnType::String => self.write_to.serialize_value(v).map_err(Error::custom),
+            ColumnType::String | ColumnType::FixedString { .. } => {
+                self.write_to.serialize_value(v).map_err(Error::custom)
```
phiSgr: maybe validate the length

phiSgr (commented):
add deserialization in `ColumnTypeVisitor`

phiSgr (approved):

PR #2944, opened 2025-11-04
on cluster

DatGuyJonathan:
> [!NOTE]
> Introduce full ClickHouse cluster support: new table `cluster` config, ON CLUSTER DDL across migrations/ops, validations, TS/Py SDK updates, new templates and E2E tests, CI jobs, and docs updates.
> 
> - **OLAP/CLI core**:
>   - Add `cluster_name` to `Table`/inframap (proto, serde) and show in CLI.
>   - Generate `ON CLUSTER` in CREATE/ALTER/DROP/SETTINGS/SAMPLE/INDEX DDL and pass cluster through all operations.
>   - Do not treat `cluster_name` as schema-diff; preserve from infra map during reconcile.
>   - Validate databases and clusters in plans/migrations; improved error messages.
>   - Handle replicated engines: auto-inject params in dev when no cluster; support cluster-aware replication.
>   - Update migration schema to allow structured `engine` objects.
> - **SDKs (TypeScript/Python)**:
>   - Add optional `cluster` to OLAP table config; serialize to internal model.
>   - Validate that `cluster` cannot be combined with explicit `keeperPath/replicaName` (TS) or `keeper_path/replica_name` (Py).
>   - Add unit tests for cluster validation.
> - **E2E & Templates**:
>   - New templates `templates/typescript-cluster` and `templates/python-cluster` with clustered/non-clustered tables.
>   - Add E2E tests (`apps/framework-cli-e2e/test/cluster.test.ts`) for infra map, CH clusters, XML generation, and table creation.
>   - CI: add `test-e2e-cluster-typescript` and `test-e2e-cluster-python` jobs; adjust path filters.
> - **Infra & Misc**:
>   - Docker compose supports mounting `clickhouse_clusters.xml` when provided.
> - **Docs**:
>   - Update configuration and model-table docs to document clusters, replication modes, and deployment guidance; note in planned migrations.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a201fdd6f29a05cec6f1fda437730c03a1046e86. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/infrastructure/olap/clickhouse/queries.rs
+++ b/apps/framework-cli/src/infrastructure/olap/clickhouse/queries.rs
@@ -2463,7 +2463,7 @@ pub fn create_table_query(
 }
 
 pub static DROP_TABLE_TEMPLATE: &str = r#"
-DROP TABLE IF EXISTS `{{db_name}}`.`{{table_name}}`{{#if cluster_name}} ON CLUSTER {{cluster_name}}{{/if}} SYNC;
+DROP TABLE IF EXISTS `{{db_name}}`.`{{table_name}}`{{#if cluster_name}} ON CLUSTER {{cluster_name}} SYNC{{/if}};
 "#;
 
 pub fn drop_table_query(
@@ -4623,8 +4623,11 @@ ENGINE = S3Queue('s3://my-bucket/data/*.csv', NOSIGN, 'CSV')"#;
             "DROP query should contain ON CLUSTER clause"
         );
 
-        // Should have SYNC (always present)
-        assert!(query.contains("SYNC"), "DROP query should contain SYNC");
+        // Should have SYNC (when using ON CLUSTER)
+        assert!(
+            query.contains("SYNC"),
+            "DROP query should contain SYNC with ON CLUSTER"
+        );
 
         // Should have DROP TABLE
         assert!(query.contains("DROP TABLE"));
@@ -4641,8 +4644,11 @@ ENGINE = S3Queue('s3://my-bucket/data/*.csv', NOSIGN, 'CSV')"#;
             "DROP query should not contain ON CLUSTER clause when cluster_name is None"
         );
 
-        // Should still have SYNC (for replicated tables)
-        assert!(query.contains("SYNC"), "DROP query should contain SYNC");
+        // Should NOT have SYNC (only needed with ON CLUSTER)
+        assert!(
+            !query.contains("SYNC"),
+            "DROP query should not contain SYNC without ON CLUSTER"
+        );
 
         // Should still have DROP TABLE
         assert!(query.contains("DROP TABLE"));


.github/workflows/test-framework-cli.yaml
```diff
@@ -716,6 +716,138 @@ jobs:
         run: |
           cat ~/.moose/*-cli.log
 
+  test-e2e-cluster-typescript:
```
LucioFranco, line 719: I'd like to find a better approach then having to add jobs each time we add new tests. Just throwing this out there, no need to solve this now.

apps/framework-cli/src/framework/core/plan_validator.rs
```diff
@@ -10,11 +10,72 @@ pub enum ValidationError {
...
+    ClusterValidation(String),
+}
+
+/// Validates that all tables with cluster_name reference clusters defined in the config
+fn validate_cluster_references(project: &Project, plan: &InfraPlan) -> Result<(), ValidationError> {
+    let defined_clusters = project.clickhouse_config.clusters.as_ref();
+
+    // Get all cluster names from the defined clusters
+    let cluster_names: Option<Vec<String>> =
+        defined_clusters.map(|clusters| clusters.iter().map(|c| c.name.clone()).collect());
```
LucioFranco: Nit (suggestion no need to actually do this): Representing this state as option<vec> is a bit hard to follow but instead I would suggest leveraging the empty vec state as the none state and then adjusting the match below to be a set of if statements that each run through the validations. I find that match pretty hard to follow, I'd like to see a `if names.contains(...) { continue }` since I think that reads super clear to me.

apps/framework-cli/src/infrastructure/olap/clickhouse/mod.rs
```diff
@@ -1767,6 +1935,10 @@ impl OlapOperations for ConfiguredDBClient {
                 indexes,
                 database: Some(database),
                 table_ttl_setting,
+                // cluster_name is always None from introspection because ClickHouse doesn't store
+                // the ON CLUSTER clause - it's only used during DDL execution and isn't persisted
+                // in system tables. Users must manually specify cluster in their table configs.
+                cluster_name: None,
```
LucioFranco, line 1939: would it make sense to provide some sort of escape hatch say let the user set an obscure env var in-case the on cluster stuff changes and we were not able to pick up that change etc since its not tied to our normal reconciliation process?

DatGuyJonathan: Hey! I think the cluster field in their table config is like an escape hatch - it doesn't trigger drop+create so it can be modified for future ddls. 

Could you walk me through the scenario you're thinking of?

LucioFranco: Ok cool if the config file provides the hatch that makes sense. I was thinking worst case scenario you can provide an env var they can set that would pick up the change etc, but I think with the config file it fines.

apps/framework-cli/src/cli/routines/migrate.rs
```diff
@@ -218,7 +218,11 @@ fn validate_table_databases(
             SerializableOlapOperation::CreateTable { table } => {
                 validate_db(&table.database, &table.name);
             }
-            SerializableOlapOperation::DropTable { table, database } => {
+            SerializableOlapOperation::DropTable {
+                table,
+                database,
+                cluster_name: _,
+            } => {
                 validate_db(database, table);
```
callicles: How come this is not used here?

DatGuyJonathan: Good catch. I had validation during plan generation, but this is during plan execution and the plan files could be modified in between. Added the validation here as well.

.github/workflows/test-framework-cli.yaml
```diff
@@ -43,8 +43,8 @@ jobs:
             "^\.github/workflows/test-framework-cli\.yaml"
             "^apps/framework-cli-e2e/"
             "^apps/framework-cli/"
-            "^templates/python/"
-            "^templates/typescript/"
+            "^templates/python"
+            "^templates/typescript"
```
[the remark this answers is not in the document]
DatGuyJonathan: Hmm, should be okay. There are other typescript & python templates too that should trigger tests if they change.

callicles (approved):

DatGuyJonathan:
> This is missing documentation. Is the plan to follow up with documentation?

I think I already have some, but lemme see if I need to migrate it to docs v2

PR #2982, opened 2025-11-13
Moose Plan backwards compatability

georgevanderson:
Old versions of moose might have engine null stored in redis, in which case we should use the default engine

Keys stored in old versions did not have the database information for the tables

---

> [!NOTE]
> Makes `Table.engine` non-optional with default `MergeTree`, adds infra map normalization (e.g., order_by from primary keys, arrays required), and updates diff/diagnostics/generators to ensure stable backward-compatible plans.
> 
> - **Core Infrastructure**:
>   - Make `Table.engine` non-optional (`ClickhouseEngine`, default `MergeTree`); always display/serialize engine and default on decode.
>   - Update `compute_non_alterable_params_hash`, `order_by_equals`, and proto conversions to reflect mandatory engine.
> - **Infrastructure Map & Diff**:
>   - Add `InfrastructureMap::normalize()` (fill `order_by` from primary keys for MergeTree, force array columns `required=true`).
>   - Normalize remote/local maps before diff; compare using HashMap keys; adjust engine/order_by comparisons to non-optional engine.
>   - Resolve S3 credentials and recalc `engine_params_hash` accordingly.
> - **ClickHouse Integration**:
>   - Parse engine from `CREATE TABLE`, default to `MergeTree`; mapper uses engine directly.
>   - Diff strategy: engine change detection simplified; S3Queue column changes force drop+create.
>   - Diagnostics API now carries a concrete engine (no `Option`).
> - **Code Generation & Models**:
>   - TS/Python generators emit concrete engine configs; data model maps deduplicate to `ReplacingMergeTree` else `MergeTree`.
> - **Tests/E2E**:
>   - Update tests/fixtures for mandatory engine; minor e2e log noise reduction.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 92bcb265ea3ac3c1d002d91c9378614cd84183b5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli-e2e/test/backward-compatibility.test.ts
+++ b/apps/framework-cli-e2e/test/backward-compatibility.test.ts
@@ -15,7 +15,7 @@
  * 5. Asserts that no (or minimal expected) changes are detected
  *
  * This is critical for catching breaking changes in infrastructure map format,
- * particularly changes like table ID prefixes with database names.
+ * particularly changes like table ID prefixes with database names and upgrades
  */
 
 import { spawn, ChildProcess } from "child_process";


apps/framework-cli/src/cli/routines.rs
```diff
@@ -857,6 +857,30 @@ pub(crate) async fn get_remote_inframap_protobuf(
...
+/// # Arguments
+/// * `infra_map` - The infrastructure map to normalize
+///
+/// # Returns
+/// * `InfrastructureMap` - The normalized infrastructure map with default engines applied
+fn normalize_inframap_engines(mut infra_map: InfrastructureMap) -> InfrastructureMap {
+    use crate::infrastructure::olap::clickhouse::queries::ClickhouseEngine;
+
+    for table in infra_map.tables.values_mut() {
+        if table.engine.is_none() {
```
phiSgr: I think we have other code that does "`None` means `MergeTree`", is this normalization done so that we can use `==`?

phiSgr (approved):

PR #2996, opened 2025-11-14
add moose query cli command

onelesd:
adds `moose query`

Showing how it mimics `moose peek`:

Screenshot 2025-11-14 at 3 39 48 PM

---

> [!NOTE]
> Introduces `moose query` to run SQL from arg/file/stdin, stream NDJSON with row limits, plus telemetry, infra checks, tests, and docs.
> 
> - **CLI**:
>   - **New command** `moose query`: execute SQL against ClickHouse from `query` arg, `-f/--file`, or stdin; streams newline-delimited JSON; `--limit` caps rows.
>   - Integrated into `top_command_handler` and command parsing (`Commands::Query`).
>   - Validates Redis/infrastructure state; requires `moose dev` running.
> - **Routines**:
>   - Added `routines/query.rs` implementing query execution, JSON streaming, and limit enforcement.
>   - Exposes `row_to_json` in `clickhouse_alt_client` for reuse.
> - **Telemetry**:
>   - Added `ActivityType::QueryCommand` and usage capture in query flow.
> - **Tests (E2E)**:
>   - `cli-query.test.ts`: covers argument, file, stdin, multi-row, `--limit`, and error handling.
> - **Docs**:
>   - Updated CLI references to document `moose query` usage, options, requirements, and output in `apps/framework-docs*`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c33bbd52d446300f10c8fbd33960659263f54817. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


apps/framework-cli/src/cli/routines/query.rs
```diff
@@ -0,0 +1,183 @@
...
+                    )
+                })?;
+
+                println!("{}", json);
+                info!("{}", json);
+                success_count += 1;
+
+                // Check limit to avoid unbounded queries
+                if success_count >= limit {
+                    info!("Reached limit of {} rows", limit);
```
onelesd: we do it this way because the rust client does not support max_count_row client options. we'd have to switch to the http client to get those options.

apps/framework-cli/src/cli/routines/query.rs
```diff
@@ -0,0 +1,183 @@
...
+        .map_err(|_| {
+            RoutineFailure::error(Message::new(
+                "Failed".to_string(),
+                "Error retrieving current state".to_string(),
+            ))
+        })?
+        .ok_or_else(|| {
+            RoutineFailure::error(Message::new(
+                "Failed".to_string(),
+                "No state found".to_string(),
```
LucioFranco: error message could be a bit more descriptive

apps/framework-cli/src/cli/routines/query.rs
```diff
@@ -0,0 +1,183 @@
...
+            ))
+        })?;
+
+    // Execute query and stream results
+    let mut stream = client.query(&sql_query).stream();
+
+    let mut success_count = 0;
+    let mut enum_mappings: Option<Vec<Option<Vec<&str>>>> = None;
```
LucioFranco: feel like this type could be improved its a bit hard to grok, I think we can leverage empty vec state as Option::none?

apps/framework-cli/src/cli/routines/query.rs
```diff
@@ -0,0 +1,183 @@
...
+                // Reuse peek's row_to_json with enum mappings
+                let value = row_to_json(&row, enum_mappings.as_ref().unwrap()).map_err(|e| {
+                    RoutineFailure::new(
+                        Message::new(
+                            "Query".to_string(),
+                            "Failed to convert row to JSON".to_string(),
+                        ),
+                        e,
+                    )
+                })?;
```
LucioFranco: I think using a if let here would improve this code

docs/plans/2025-01-14-moose-query-command.md
LucioFranco, line 1: Do we need this? Feel like this is the content that should live in linear?

onelesd: we don't need it - it's the llm implementation plan

apps/framework-cli/src/cli/routines/query.rs
onelesd: needed this because `moose query` doesn't know what the tables in the query are, and the shared `row_to_json()` function needs to know how many columns there are.

could you suggest a patch?

LucioFranco:
```diff
diff --git a/apps/framework-cli/src/cli/routines/query.rs b/apps/framework-cli/src/cli/routines/query.rs
index a5928f967e..5af692decc 100644
--- a/apps/framework-cli/src/cli/routines/query.rs
+++ b/apps/framework-cli/src/cli/routines/query.rs
@@ -123,18 +123,18 @@
     let mut stream = client.query(&sql_query).stream();

     let mut success_count = 0;
-    let mut enum_mappings: Option<Vec<Option<Vec<&str>>>> = None;
+    let mut enum_mappings: Vec<Option<Vec<&str>>> = Vec::new();

     while let Some(row_result) = stream.next().await {
         match row_result {
             Ok(row) => {
                 // Create enum mappings on first row (one None entry per column)
-                if enum_mappings.is_none() {
-                    enum_mappings = Some(vec![None; row.len()]);
+                if enum_mappings.is_empty() {
+                    enum_mappings = vec![None; row.len()];
                 }

                 // Reuse peek's row_to_json with enum mappings
-                let value = row_to_json(&row, enum_mappings.as_ref().unwrap()).map_err(|e| {
+                let value = row_to_json(&row, &enum_mappings).map_err(|e| {
                     RoutineFailure::new(
                         Message::new(
                             "Query".to_string(),
```

Something like this maybe?

PR #3027, opened 2025-11-21
add express api key auth middleware

onelesd:
> [!NOTE]
> Add Express API key authentication middleware with PBKDF2, integrate into templates and e2e tests, and document setup/usage.
> 
> - **Library (`packages/ts-moose-lib`)**:
>   - **Express API key auth**: Add `expressApiKeyAuthMiddleware()` in `src/consumption-apis/webAppHelpers.ts` using PBKDF2 HMAC SHA256, constant-time compare, and `MOOSE_WEB_APP_API_KEYS`; export helpers (`constantTimeCompare`, `validateAuthToken`, `getValidApiKeys`).
>   - **Tests**: Add `tests/webAppAuth.test.ts` covering helpers and middleware (missing/malformed headers, valid/invalid tokens, multi-key support).
> - **Templates**:
>   - **Protected APIs**: Add authenticated Express apps in `templates/typescript-express/app/apis/protected.ts` and `templates/typescript-tests/src/apis/protectedApiKey.ts`; export via `app/index.ts` and `src/index.ts`.
> - **E2E**:
>   - **Auth flow**: Update `apps/framework-cli-e2e/test/templates.test.ts` to generate hash token via CLI, set `MOOSE_WEB_APP_API_KEYS`, and verify protected routes (`/protected-api-key/health`, `/echo`) enforce 401/200 behaviors.
> - **Docs**:
>   - **Guides**: Add "API Key Authentication" sections with setup, usage, and security notes to `apps/framework-docs-v2/.../express.mdx` and `apps/framework-docs/.../express.mdx`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 887c747df4093fcea7aa6a6af19f14f8779c8a85. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/ts-moose-lib/src/consumption-apis/webAppHelpers.ts
```diff
@@ -1,6 +1,201 @@
...
+ * which can leak information about the correct token through timing measurements.
+ *
+ * Uses utf16le encoding because it's the only Node.js encoding that doesn't
+ * lose information (utf8 normalizes lone surrogates, latin1 truncates multi-byte
+ * characters, base64 drops whitespace, etc.)
+ */
+function constantTimeCompare(a: string, b: string): boolean {
+  try {
+    // Early return if lengths differ (length is typically not secret for API keys)
+    if (a.length !== b.length) {
```
callicles: That defeats the purpose of the timing function below - that would tell an attacker how long the password is. we should not have an early return

packages/ts-moose-lib/src/consumption-apis/webAppHelpers.ts
```diff
@@ -1,6 +1,201 @@
...
+ * Matches the Rust implementation in framework-cli/src/cli/routines/auth.rs
+ */
+function validateAuthToken(token: string, expectedHash: string): boolean {
+  const tokenParts = token.split(".");
+  if (tokenParts.length !== 2) {
+    return false;
+  }
+
+  const tokenHex = tokenParts[0];
+  const saltHex = tokenParts[1];
```
callicles, line 47: is that the same order as in moose - I thought I recalled the other way around

packages/ts-moose-lib/src/consumption-apis/webAppHelpers.ts
```diff
@@ -1,6 +1,201 @@
...
+ * - Authorization header is missing or malformed
+ * - API key doesn't match any configured hash
+ */
+export function expressApiKeyAuthMiddleware(): (
+  req: any,
+  res: any,
+  next: any,
+) => void {
+  return (req: any, res: any, next: any): void => {
+    const validApiKeys = getValidApiKeys();
```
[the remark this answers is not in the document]
callicles: Valid comment we should load it only once at start up

packages/ts-moose-lib/src/consumption-apis/webAppHelpers.ts
onelesd: crypto.timingSafeEqual() throws if the buffers aren't the same length. whats the idiomatic way to do this?

callicles: try catch it I believe

onelesd: just rely on the throw? maybe the implementation is accounting for this internally

callicles: Yep

packages/ts-moose-lib/src/consumption-apis/webAppHelpers.ts
onelesd: matches here: https://github.com/514-labs/moosestack/blob/dseleno/feat/ENG-1495-express-api-key-auth-middleware/apps/framework-cli/src/cli/routines/auth.rs#L57-L59

packages/ts-moose-lib/src/consumption-apis/webAppHelpers.ts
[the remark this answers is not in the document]
onelesd: agree - problem is having it in the global scope causes deployments to fail because env vars aren't present when the image is created. we'd need to make env vars available during the build phase.

onelesd:
> Why didn't we reuse the implementation in the CLI proxy?
> 
> Could work for both python and ts at the same time?

**tl;dr: we thought it better to separate the concerns**

the proxy relies on the `MOOSE_CONSUMPTION_API_KEY` env var for native consumption APIs and we wanted to use different env vars for byo api because the scope of a byo api app can be more than just consumption apis - it's why we chose `MOOSE_WEB_APP_API_KEYS` for the env var name

callicles:
Wouldn't it be easy enough to just reuse the same logic and load a different env for that code path?

onelesd:
closing in favor of:

- not supporting an express middleware in moosestack
- choosing an existing express middleware on npm and showcasing that in the express template

PR #3033, opened 2025-11-25
Nico/eng 1134 implement toon compression for mcp query results

callicles:
> [!NOTE]
> MCP tools now return TOON-formatted results with a new compressed infrastructure map; the diagnose tool is renamed, embedded docs and get_source are removed, and SQL/flow resources capture source_file across TS/Py/Rust/Proto.
> 
> - **MCP server/tools**:
>   - Add `compressed_map` with lineage-focused `CompressedInfraMap` and TOON output in `get_infra_map` (fuzzy search support).
>   - Rename `diagnose_infrastructure` → `get_issues` and emit TOON.
>   - Make `query_olap` default to `toon` (also supports `table`).
>   - Make `get_stream_sample` default to TOON (also supports JSON).
>   - Remove `get_source` tool and embedded docs (resources API removed); update server routing and tests.
> - **Infra modeling**:
>   - Track `source_file` for `SqlResource` end-to-end (TS/Py SDKs, Rust structs, ClickHouse introspection default, protobuf schema); include in infra map and tests.
>   - Capture source locations for transforms/consumers/functions; expose `InfrastructureSignature::id()`.
> - **Build/docs**:
>   - Drop build-time docs generator; update docs to new tools/flows with TOON examples.
> - **Deps**:
>   - Add `toon-format` and related libraries; refresh lockfiles.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 75bf9613615d4737a68de50c91363b0eeea14efb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/mcp/compressed_map.rs
+++ b/apps/framework-cli/src/mcp/compressed_map.rs
@@ -6,6 +6,7 @@
 
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
+use std::path::Path;
 
 use crate::framework::core::infrastructure::table::Metadata;
 use crate::framework::core::infrastructure_map::InfrastructureMap;
@@ -170,21 +171,18 @@ impl Default for CompressedInfraMap {
 /// Convert an absolute file path to a relative path from the project root
 /// Strips the project directory prefix to get a path like "app/datamodels/User.ts"
 fn make_relative_path(absolute_path: &str) -> String {
-    // Find the last occurrence of "/app/" or "\app\" and take everything from "app/" onwards
-    // This works for paths like "/Users/name/project/app/datamodels/User.ts" -> "app/datamodels/User.ts"
-    if let Some(pos) = absolute_path.rfind("/app/") {
-        return absolute_path[pos + 1..].to_string();
-    }
-    if let Some(pos) = absolute_path.rfind("\\app\\") {
-        return absolute_path[pos + 1..].replace('\\', "/");
+    let path = Path::new(absolute_path);
+
+    // Walk through ancestors to find one ending with "app", then get relative path
+    for ancestor in path.ancestors() {
+        if ancestor.file_name().is_some_and(|name| name == "app") {
+            if let Ok(relative) = path.strip_prefix(ancestor.parent().unwrap_or(ancestor)) {
+                return relative.display().to_string();
+            }
+        }
     }
 
-    // If path already starts with "app/", return as-is
-    if absolute_path.starts_with("app/") || absolute_path.starts_with("app\\") {
-        return absolute_path.replace('\\', "/");
-    }
-
-    // Fallback: return the original path if we can't find "app/"
+    // Fallback: return the original path
     absolute_path.to_string()
 }
 
@@ -528,30 +526,16 @@ mod tests {
     }
 
     #[test]
-    fn test_make_relative_path_with_absolute_unix() {
+    fn test_make_relative_path_with_absolute() {
         let absolute = "/Users/nicolas/code/514/test-projects/ts-test-tests/app/ingest/models.ts";
         let relative = make_relative_path(absolute);
         assert_eq!(relative, "app/ingest/models.ts");
     }
 
     #[test]
-    fn test_make_relative_path_with_absolute_windows() {
-        let absolute = "C:\\Users\\nicolas\\code\\project\\app\\datamodels\\User.ts";
-        let relative = make_relative_path(absolute);
-        assert_eq!(relative, "app/datamodels/User.ts");
-    }
-
-    #[test]
     fn test_make_relative_path_already_relative() {
         let already_relative = "app/ingest/models.ts";
         let result = make_relative_path(already_relative);
-        assert_eq!(result, "app/ingest/models.ts");
-    }
-
-    #[test]
-    fn test_make_relative_path_with_backslashes() {
-        let windows_relative = "app\\ingest\\models.ts";
-        let result = make_relative_path(windows_relative);
         assert_eq!(result, "app/ingest/models.ts");
     }
 
@@ -565,7 +549,7 @@ mod tests {
 
     #[test]
     fn test_make_relative_path_nested_app_directories() {
-        // Should use the last occurrence of "/app/"
+        // Should use the innermost (last) "app" directory
         let path = "/Users/app/project/app/datamodels/User.ts";
         let result = make_relative_path(path);
         assert_eq!(result, "app/datamodels/User.ts");


apps/framework-cli/src/mcp/tools/infra_map.rs
```diff
@@ -349,242 +93,73 @@ async fn execute_get_infra_map(
...
-                &infra_map.workflows,
-                search,
-                component_type_filter,
-                show_all,
-            )?,
-        ];
+        // Filter connections: keep if source OR destination is in filtered components
+        filtered.connections.retain(|conn| {
+            component_ids.contains(conn.from.as_str()) || component_ids.contains(conn.to.as_str())
+        });
```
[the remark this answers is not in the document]
callicles: That's on purpose - we want to be able to zoom in and that means dangling references. Those refrences should have alreeady been received by the LLM as the initial call to the get infra map

apps/framework-cli/src/mcp/server.rs
```diff
@@ -142,19 +134,40 @@ impl ServerHandler for MooseMcpHandler {
...
-            message: format!("Resource not found: {}", request.uri).into(),
-            data: None,
-        })
+        // Route based on URI scheme
+        if request.uri.starts_with("moose://infra/") {
+            // Handle infrastructure resources
+            infra_resources::read_infra_resource(&request.uri, self.redis_client.clone())
+                .await
+                .ok_or_else(|| ErrorData {
+                    code: ErrorCode::INVALID_PARAMS,
```
LucioFranco: am I missing something or this could error in a way that is not actually invalid params? Like say a transient network error?

apps/framework-cli/src/mcp/tools/toon_serializer.rs
```diff
@@ -0,0 +1,166 @@
...
+/// # Returns
+/// * `Ok(String)` - TOON-formatted string
+/// * `Err(ToonSerializationError)` - If encoding fails
+///
+/// # Example
+/// ```rust
+/// let data = json!({"users": [{"id": 1, "name": "Alice"}]});
+/// let toon_str = serialize_to_toon_compressed(&data)?;
+/// ```
+pub fn serialize_to_toon_compressed(value: &Value) -> Result<String, ToonSerializationError> {
```
LucioFranco: Is this old? Your issue got fixed for this https://github.com/toon-format/toon-rust/issues/21

apps/framework-cli/Cargo.toml
```diff
@@ -116,6 +116,7 @@ keyring = { version = "3.6", features = ["apple-native", "linux-native"] }
 rmcp = { version = "0.8.1", features = ["server", "transport-streamable-http-server"] }
 percent-encoding = "2.3.2"
 dotenvy = "0.15"
+toon-format = "0.3.2"
```
LucioFranco: Looks like we are behind as 0.4 is out now. This crate def feels VERY vibe coded and the maintainer doesn't seem to know rust that well from what I can see. That said I am not super concerned since we are using this as an encoding output. We should 100% allow outputing both formats (didnt get that far in the review so maybe its already done).

I've also opened an issue for adding feature flag support https://github.com/toon-format/toon-rust/issues/39

apps/framework-cli/src/mcp/compressed_map.rs
```diff
@@ -0,0 +1,659 @@
...
+
+impl Default for CompressedInfraMap {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Convert an absolute file path to a relative path from the project root
+/// Strips the project directory prefix to get a path like "app/datamodels/User.ts"
+fn make_relative_path(absolute_path: &str) -> String {
```
LucioFranco, line 173: Why not use `Path` here? iirc it should have a lot of the pathing logic built in like the direction of the slash

apps/framework-cli/src/mcp/compressed_map.rs
```diff
@@ -0,0 +1,659 @@
...
+        compressed.add_connection(Connection {
+            from: key.clone(),
+            to: sync.target_topic_id.clone(),
+            connection_type: ConnectionType::Produces,
+        });
+    }
+}
+
+/// Build a compressed infrastructure map from the full InfrastructureMap
+pub fn build_compressed_map(infra_map: &InfrastructureMap) -> CompressedInfraMap {
```
LucioFranco, line 434: could just be a `CompressedInfraMap::from`? I even think theoreticlly (thought his may be hard to actually do) you could make the compressed view a "view" type ontop of the infra map such that it is something like `CompressedInfraMap<'a>` where `'a` is the lifetime of `&'a InfrastructureMap` and save a bunch of allocations but I don't think that is worth it here :smile: thought Id just give that a shout.

apps/framework-cli/src/mcp/compressed_map.rs
```diff
@@ -0,0 +1,659 @@
...
+    let prefix = "moose://infra/";
+    if !uri.starts_with(prefix) {
+        return None;
+    }
+
+    let path = &uri[prefix.len()..];
+    let parts: Vec<&str> = path.splitn(2, '/').collect();
+    if parts.len() != 2 {
+        return None;
+    }
```
LucioFranco: Nit: I would keep it as an iterator and just call next twice like `let component_type = split.next()?;`.

apps/framework-cli/src/mcp/infra_resources.rs
```diff
@@ -0,0 +1,604 @@
...
+use super::compressed_map::{build_resource_uri, parse_resource_uri, ComponentType};
+use super::tools::toon_serializer::serialize_to_toon_compressed;
+use crate::framework::core::infrastructure_map::InfrastructureMap;
+use crate::infrastructure::redis::redis_client::RedisClient;
+
+/// List all available infrastructure resources
+pub async fn list_infra_resources(redis_client: Arc<RedisClient>) -> ListResourcesResult {
+    let infra_map = match InfrastructureMap::load_from_redis(&redis_client).await {
+        Ok(Some(map)) => map,
+        _ => {
```
LucioFranco: Why are we throwing away a potential error here?

apps/framework-cli/src/mcp/infra_resources.rs
```diff
@@ -0,0 +1,604 @@
...
+            raw: RawResource {
+                uri: build_resource_uri(ComponentType::Topic, key),
+                name: topic.name.clone(),
+                title: Some(format!("Topic: {}", topic.name)),
+                description: topic.metadata.as_ref().and_then(|m| m.description.clone()),
+                mime_type: Some("application/x-toon".to_string()),
+                size: None,
+                icons: None,
+            },
+            annotations: None,
```
LucioFranco: Suggestions: I think this code would be better to sit closer to the struct definition, aka there should be some file that contains the impl of an annotated topic and within that file there is a `Into` implementation for `RawResource` that way this code here that converts a generic infra map into an mcp specific one is just calling `.into()` and its easy to find the struct def and conversion logic.

apps/framework-cli/src/mcp/infra_resources.rs
```diff
@@ -0,0 +1,604 @@
...
+            "columns": table.columns.iter().map(|col| json!({
+                "name": col.name,
+                "data_type": format!("{:?}", col.data_type),
+                "required": col.required,
+                "primary_key": col.primary_key,
+                "unique": col.unique,
+            })).collect::<Vec<_>>(),
+        },
+    })
+}
```
LucioFranco: imo this code should live in `infrastructure::table` and/or maybe we should just define a struct for this json struct and have a struct conversion (into/from) then call serialie on that struct rather than do big inline json defs which imo are quite hard to read.

apps/framework-cli/src/mcp/infra_resources.rs
```diff
@@ -0,0 +1,604 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_serialize_topic_basic() {
```
LucioFranco: one test :) maybe we should add some more for the custom code above?

apps/framework-cli/src/mcp/tools/sample_stream.rs
```diff
@@ -53,7 +57,7 @@ struct GetStreamSampleParams {
     stream_name: String,
     /// Number of messages to retrieve (default: 10, max: 100)
     limit: u8,
-    /// Output format: "json" or "pretty" (default: "json")
+    /// Output format: "toon" or "pretty" (default: "toon")
```
LucioFranco: pretty is json? not super clear

apps/framework-cli/src/mcp/tools/infra_map.rs
```diff
@@ -1,37 +1,16 @@
 //! # Infrastructure Map Tool
```
LucioFranco, line 1: This diff was really hard for me to follow, could you help explain what actually changed here? There seems to be a lot removed.

apps/framework-cli/src/framework/core/partial_infrastructure_map.rs
```diff
@@ -371,20 +371,28 @@ pub struct WriteTo {
 ///
 /// Used to define where transformed data should be written and optionally specify a version.
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
```
LucioFranco, line 374: is this going to break anything downstream?

apps/framework-cli/src/mcp/tools/toon_serializer.rs
callicles: Updated

apps/framework-cli/Cargo.toml
callicles: Updated

apps/framework-cli/src/mcp/compressed_map.rs
callicles: Yeah - probably.
Not worth the time though here 😄

apps/framework-cli/src/framework/core/partial_infrastructure_map.rs
callicles: I don't think so - none of the fields required snake case

apps/framework-cli/src/mcp/tools/infra_map.rs
callicles: We used to have on this tool:
* a search bar
* a component type filter
* a full vs summary view of the infra map

Now we only have search. Simplifies the inputs a lot for the llm and with the compression no need for the various drop downs anymore

apps/framework-cli/src/mcp/infra_resources.rs
callicles: Hum - I think  should just remove that code then.
I realized resources are [application controlled ](https://modelcontextprotocol.info/docs/concepts/resources/) so the LLM will not necessarily request them. Since all the config is in code anyway I am tempted to just remove all of this

callicles: I will just remove

apps/framework-cli/src/mcp/infra_resources.rs
callicles: will be removed

apps/framework-cli/src/mcp/infra_resources.rs
callicles: will be removed

apps/framework-cli/src/mcp/tools/sample_stream.rs
callicles: Updated

apps/framework-cli/src/mcp/compressed_map.rs
callicles: Changed it within the function - source_file is using a string because it is coming from the ts / py

LucioFranco (approved):

PR #3050, opened 2025-12-01
rfd: add improved moose errors

LucioFranco:
[Rendered](https://github.com/514-labs/moosestack/blob/e9eb2a698dc86531708def9383bf80c6dce32d12/rfd/0009/README.mdx)

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit e9eb2a698dc86531708def9383bf80c6dce32d12. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/rfd/0009/README.mdx
+++ b/rfd/0009/README.mdx
@@ -1,0 +1,656 @@
+---
+authors: Lucio <lucio@fiveonefour.com>
+title: Improved Error Handling with Contextual Metadata
+state: prediscussion
+---
+
+# Improved Error Handling with Contextual Metadata
+
+## Summary
+
+This RFD proposes adopting `color-eyre` and `tracing` as the foundation for error handling in Moose CLI. This approach provides rich contextual error messages with span traces, suggestions, and structured metadata without requiring custom error infrastructure.
+
+## Problem Statement
+
+The current error handling in Moose CLI has several limitations:
+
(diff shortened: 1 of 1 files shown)


onelesd (commented):
solid proposal

we should add:

1. creation of agent instructions on how/when to use the available eyre methods: .note(), .suggestion(), etc.
2. error codes - we show bubbling up clickhouse error codes, for example, but our own stable error codes would make tracing easier. it can be a both situation.
3. an example of nested errors. in this RFD you mention handling errors at the edge/high-level and let errors bubble up, which is the right way. but does that avoid nested errors entirely?
4. setting max verbosity automatically in CI

final thought, an expansion of #1: error handling approaches are only as good as the discipline engineers/agents apply when using them. a few examples for engineers/agents of nuanced cases would go a long way to encouraging proper usage.

rfd/0009/README.mdx
```diff
@@ -0,0 +1,656 @@
...
+# Standard output (default)
+moose dev
+
+# Verbose output with SpanTrace
+moose dev --verbose
+moose dev -v
+
+# Extra verbose with Backtrace
+moose dev --verbose --verbose
+moose dev -vv
```
jmsuzuki, line 506: does this apply to prod?

oatsandsugar (approved):

PR #3067, opened 2025-12-03
ENG-1641 chore: clean up streaming function implementation with new lib

callicles:
> [!NOTE]
> Centralizes Kafka producer config, simplifies streaming runner to rely on Confluent batching with DLQ fallback, adds topic name/version utilities, bumps @confluentinc/kafka-javascript to 1.7.0, and introduces targeted tests.
> 
> - **Kafka/Config**:
>   - Add `createProducerConfig` in `src/commons.ts` (idempotent=false, `acks`=`-1`, retries, `linger.ms`=0, optional `message.max.bytes`).
>   - Update `getKafkaProducer` and runner producer creation to use `createProducerConfig` and sync with `targetTopic.max_message_bytes`.
>   - Bump `@confluentinc/kafka-javascript` to `1.7.0`.
> - **Streaming Runner (`src/streaming-functions/runner.ts`)**:
>   - Remove custom message-size batching/splitting; rely on client batching and retries; set consumer `js.consumer.max.batch.size`.
>   - Add `handleDLQForFailedMessages` and route failed sends to per-message DLQs; count metrics only on successful sends.
>   - Add topic utilities and validation: `formatVersionSuffix`, `topicNameToStreamName`, `validateTopicConfig`.
> - **Tests**:
>   - Add `tests/producer-config.test.ts` for producer config and constants.
>   - Add `tests/streaming-runner.test.ts` for topic naming/validation and `TopicConfig` expectations.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0ca620540657f158654b2400278e13fecb477ae2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/ts-moose-lib/src/commons.ts
+++ b/packages/ts-moose-lib/src/commons.ts
@@ -135,7 +135,7 @@ export function createProducerConfig(maxMessageBytes?: number) {
         maxRetryTime: MAX_RETRY_TIME_MS,
       },
     },
-    "linger.ms": 0, // Send immediately - batching happens at application level
+    "linger.ms": 0, // This is to make sure at least once delivery with immediate feedback on the send
     ...(maxMessageBytes && { "message.max.bytes": maxMessageBytes }),
   };
 }
--- a/packages/ts-moose-lib/src/streaming-functions/runner.ts
+++ b/packages/ts-moose-lib/src/streaming-functions/runner.ts
@@ -46,6 +46,8 @@ const MAX_RETRIES_CONSUMER = 150;
 const SESSION_TIMEOUT_CONSUMER = 30000;
 const HEARTBEAT_INTERVAL_CONSUMER = 3000;
 const DEFAULT_MAX_STREAMING_CONCURRENCY = 100;
+// Max messages per eachBatch call - Confluent client defaults to 32, increase for throughput
+const CONSUMER_MAX_BATCH_SIZE = 1000;
 
 /**
  * Data structure for metrics logging containing counts and metadata
@@ -923,6 +925,7 @@ export const runStreamingFunctions = async (
         logger,
       );
 
+      // Note: "js.consumer.max.batch.size" is a librdkafka native config not in TS types
       const consumer: Consumer = kafka.consumer({
         kafkaJS: {
           groupId: streamingFuncId,
@@ -935,6 +938,7 @@ export const runStreamingFunctions = async (
           autoCommitInterval: AUTO_COMMIT_INTERVAL_MS,
           fromBeginning: true,
         },
+        "js.consumer.max.batch.size": CONSUMER_MAX_BATCH_SIZE,
       });
 
       // Sync producer message.max.bytes with topic config
--- a/packages/ts-moose-lib/package.json
+++ b/packages/ts-moose-lib/package.json
@@ -42,7 +42,7 @@
     "csv-parse": "^5.5.5",
     "fastq": "1.17.1",
     "jose": "5.9.2",
-    "@confluentinc/kafka-javascript": "1.5.0",
+    "@confluentinc/kafka-javascript": "1.7.0",
     "@kafkajs/confluent-schema-registry": "3.9.0",
     "redis": "^4.6.13",
     "toml": "3.0.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -736,8 +736,8 @@ importers:
         specifier: 1.5.0
         version: 1.5.0
       '@confluentinc/kafka-javascript':
-        specifier: 1.5.0
-        version: 1.5.0
+        specifier: 1.7.0
+        version: 1.7.0
       '@kafkajs/confluent-schema-registry':
         specifier: 3.9.0
         version: 3.9.0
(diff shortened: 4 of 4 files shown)


packages/ts-moose-lib/src/streaming-functions/runner.ts
```diff
@@ -623,68 +457,30 @@ const sendMessages = async (
...
-      );
-      logger.log(
-        `Sent final ${chunk.length} transformed data to ${targetTopic.name}`,
-      );
-    }
+  try {
+    // Library handles batching and retries internally
+    await producer.send({
+      topic: targetTopic.name,
+      messages: messages,
```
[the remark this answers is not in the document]
callicles: I double checked here: https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md that the parameter is respected

packages/ts-moose-lib/src/streaming-functions/runner.ts
```diff
@@ -623,68 +457,30 @@ const sendMessages = async (
...
+    });
+    logger.log(`Sent ${messages.length} messages to ${targetTopic.name}`);
   } catch (e) {
+    // Library already retried - this is a permanent failure
```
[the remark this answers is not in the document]
callicles: The library under the hood already handles the retries

packages/ts-moose-lib/src/streaming-functions/runner.ts
```diff
@@ -600,21 +371,107 @@ const handleMessage = async (
...
+  logger: Logger,
+  producer: Producer,
+  messages: KafkaMessageWithLineage[],
+  error: unknown,
+): Promise<boolean> => {
+  let messagesHandledByDLQ = 0;
+  let messagesWithoutDLQ = 0;
+  let dlqErrors = 0;
+
+  for (const msg of messages) {
```
onelesd, line 394: would concurrency make sense here?

```typescript
import pLimit from 'p-limit';

const limit = pLimit(MAX_CONCURRENCY);

const results = await Promise.all(
  items.map(item =>
    limit(() => asyncFunction(item))
  )
);
```

packages/ts-moose-lib/src/streaming-functions/runner.ts
```diff
@@ -623,69 +480,41 @@ const sendMessages = async (
...
+    // Handle DLQ for failed messages
+    // Only throw if not all messages were successfully routed to DLQ
+    const allHandledByDLQ = await handleDLQForFailedMessages(
+      logger,
+      producer,
+      messages,
+      e,
+    );
+    if (!allHandledByDLQ) {
+      throw e;
```
onelesd, line 518: i see we used to throw for retries but since that is handled by the kafka library do we still need to?

packages/ts-moose-lib/src/commons.ts
```diff
@@ -118,6 +118,28 @@ export const RETRY_FACTOR_PRODUCER = 0.2;
...
+    kafkaJS: {
+      idempotent: false, // Not needed for at-least-once delivery
+      acks: ACKs,
+      retry: {
+        retries: MAX_RETRIES_PRODUCER,
+        maxRetryTime: MAX_RETRY_TIME_MS,
+      },
+    },
+    "linger.ms": 0, // Send immediately - batching happens at application level
+    ...(maxMessageBytes && { "message.max.bytes": maxMessageBytes }),
```
onelesd, line 139: the logic for maxMessageBytes is kinda spread out

the way it currently is it will always be a number > 0 because of how maxMessageBytes is set in the call site at packages/ts-moose-lib/src/streaming-functions/runner.ts line 941

you could remove the `?` from maxMessageBytes in this function signature and then you don't need this object spread

onelesd: this is just a code simplification suggestion - it works as is

packages/ts-moose-lib/src/commons.ts
```diff
@@ -143,23 +165,19 @@ export type KafkaClientConfig = {
...
-    kafkaJS: {
-      idempotent: true,
-      acks: ACKs,
-      retry: {
-        retries: MAX_RETRIES_PRODUCER,
-        maxRetryTime: MAX_RETRY_TIME_MS,
-      },
-    },
-  });
+  const producer = kafka.producer(createProducerConfig(maxMessageBytes));
```
onelesd, line 180: didn't see this, which makes the above comment i left questionable

however i don't see where `getKafkaProducer()` is ever called with a maxMessageBytes, so why is it added as a parameter in this function signature?

packages/ts-moose-lib/src/streaming-functions/runner.ts
callicles: Yes - for at least once delivery

packages/ts-moose-lib/src/streaming-functions/runner.ts
callicles: Not here, this is for the deadletter queue. There is no async processing - checking the code for the batch processing if it is in there.

callicles: In the processing of streaming we are already doing something like this

```ts
      const processedMessages: (KafkaMessageWithLineage[] | undefined)[] =
        await readableStream
          .map(
            async (message) => {
              index++;
              if (
                (batch.messages.length > DEFAULT_MAX_STREAMING_CONCURRENCY &&
                  index % DEFAULT_MAX_STREAMING_CONCURRENCY) ||
                index - 1 === batch.messages.length
              ) {
                await heartbeat();
              }
              return handleMessage(
                logger,
                streamingFunctions,
                message,
                producer,
                fieldMutations,
              );
            },
            {
              concurrency: MAX_STREAMING_CONCURRENCY,
            },
          )
          .toArray();
```

packages/ts-moose-lib/src/commons.ts
callicles: It looks like we should have a Kafka producer / topic and not a generic one, if a topic is configured with smaller messages we should be able to adapt and the way the kafka client is instanciated for streams is generic. That's preparing to iterate on that part of the code

phiSgr (approved):

PR #3069, opened 2025-12-03
ENG-1633: update-error-message-for-failed pypi cleanup

jmsuzuki:
**ENG-1633: updated cleanup-pypi step with instructions on resolving its failures**
fixing this was a pain in the rear.
as such, instructions! so that the next rear is safe.

---

> [!NOTE]
> Refactors the PyPI cleanup workflow into a reusable function, adds detailed failure guidance, and standardizes retention to 80 days for both moose-cli and moose-lib.
> 
> - **GitHub Actions (`.github/workflows/release-cli.yaml`)**:
>   - **PyPI cleanup step**:
>     - Refactor into `cleanup_package` function using `expect` to DRY repeated blocks.
>     - Enhance error output with device authentication instructions and help link.
>     - Run for both `moose-cli` and `moose-lib` via function calls.
>     - Standardize retention to 80 days for `moose-lib` (was 100).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ed6d803744a13b48fd9ef0290bcfc4488d7f14ff. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/release-cli.yaml
+++ b/.github/workflows/release-cli.yaml
@@ -735,8 +735,13 @@ jobs:
 
       - name: Clean up PyPI and only keep the latest 80 days of moose CLI and lib
         run: |
-          expect << 'EOF'
-          spawn pypi-cleanup -u 514 -p moose-cli -d 80 -r ".*" --do-it --yes
+          # Function to run pypi-cleanup with error handling
+          cleanup_package() {
+            local package_name=$1
+            local days=$2
+
+            expect << EOF
+          spawn pypi-cleanup -u 514 -p ${package_name} -d ${days} -r ".*" --do-it --yes
           expect {
             "Authentication code:" {
               send "$env(PYPI_OTP)\r"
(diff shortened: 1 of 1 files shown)


.github/workflows/release-cli.yaml
```diff
@@ -750,34 +755,24 @@ jobs:
...
-          catch wait result
-          set exit_code [lindex $result 3]
-          if {$exit_code != 0} {
-            puts "ERROR: pypi-cleanup for moose-lib failed with exit code $exit_code"
-            exit $exit_code
-          }
-          EOF
+          # Clean up both packages
+          cleanup_package "moose-cli" 80
+          cleanup_package "moose-lib" 100
```
DatGuyJonathan: Should they have the same ttl because they versions go hand-in-hand?

jmsuzuki: I love that question. I'm inclined to agree. there's no need for moose-lib to have older packages if the cli must be pinned to the version...

jmsuzuki: fixed.

.github/workflows/release-cli.yaml
```diff
@@ -750,34 +755,24 @@ jobs:
 
           # Check exit status
           catch wait result
-          set exit_code [lindex $result 3]
-          if {$exit_code != 0} {
-            puts "ERROR: pypi-cleanup for moose-cli failed with exit code $exit_code"
-            exit $exit_code
+          set exit_code [lindex \$result 3]
+          if {\$exit_code != 0} {
```
DatGuyJonathan: The backslash before dollar sign might treat this as literal text instead of expanding it to the actual value.

jmsuzuki: removed it... no need for the extra churn whether or not it works...

PR #3071, opened 2025-12-03
add sql validation to moose native components at infra loading time

onelesd:
this introduces sql validation at infrastructure loading time. it uses sqlparser to catch sql syntax errors before sending them to ClickHouse to execute. it is a "first-pass" that will only catch syntax errors, and not semantic errors.

it also makes the semantic errors returned by ClickHouse a little friendlier by adding the sql resource name to the error.

a future goal is to hook into this syntax validation in an LSP.

screenshot showing "first-pass" syntax validation:
Screenshot 2025-12-03 at 11 11 17 AM

screenshot showing improved ClickHouse semantic error w/ resource name:
Screenshot 2025-12-03 at 11 11 31 AM

---

> [!NOTE]
> Adds first-pass SQL syntax validation for infra SQL resources and enhances ClickHouse errors with resource names and full SQL.
> 
> - **Core/Validation**:
>   - **New Module**: Introduces `framework/core/validation` with `validate_sql_statement`, `validate_sql_resource`, `validate_infrastructure_sql`, and CLI-friendly `format_diagnostics_for_cli` plus tests.
> - **Infrastructure Loading**:
>   - **SQL Validation Gate**: In `InfrastructureMap::load_from_user_code`, validates `infra_map.sql_resources` and fails early with formatted diagnostics on invalid SQL.
> - **CLI**:
>   - **Query Formatting**: `format_query::validate_sql` now delegates to `validate_sql_statement`.
> - **ClickHouse Backend**:
>   - **Error Type**: Adds `ClickhouseChangesError::SqlExecutionFailed { resource_name, message, sql }`.
>   - **Raw SQL Execution**: `execute_raw_sql` detects syntax errors (e.g., `Code: 62`) and returns rich context; includes resource extraction from `description` for both syntax and non-syntax errors.
>   - **Tests**: Adds coverage for new error formatting.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 54c2b20ed92e8c57049d6a9f557eebe8fd46bdb1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/infrastructure/olap/clickhouse/mod.rs
+++ b/apps/framework-cli/src/infrastructure/olap/clickhouse/mod.rs
@@ -100,6 +100,18 @@ pub enum ClickhouseChangesError {
         resource: Option<String>,
     },
 
+    /// Error executing raw SQL (views, materialized views, etc.)
+    /// This error provides additional context about the SQL that failed.
+    #[error("SQL syntax error in '{resource_name}'\n\n{message}\n\nSQL:\n{sql}")]
+    SqlExecutionFailed {
+        /// Name of the resource (view, materialized view, etc.)
+        resource_name: String,
+        /// Human-readable error message
+        message: String,
+        /// The SQL statement that failed
+        sql: String,
+    },
+
     /// Error for unsupported operations
     #[error("Not Supported {0}")]
     NotSupported(String),
@@ -1288,6 +1300,10 @@ async fn execute_rename_table_column(
 }
 
 /// Execute raw SQL statements
+///
+/// This function executes SQL statements against ClickHouse and provides
+/// clear error messages when SQL is rejected, including the resource name
+/// and the full SQL statement that failed.
 async fn execute_raw_sql(
     sql_statements: &[String],
     description: &str,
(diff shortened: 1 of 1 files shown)


jmsuzuki (approved):

apps/framework-cli/src/cli/routines/format_query.rs
```diff
@@ -56,8 +57,12 @@ fn parse_sql(sql: &str) -> Result<Vec<Statement>, RoutineFailure> {
 ///
 /// * `Result<(), RoutineFailure>` - Ok if valid, error with helpful message if invalid
 pub fn validate_sql(sql: &str) -> Result<(), RoutineFailure> {
-    parse_sql(sql)?;
-    Ok(())
+    validate_sql_statement(sql).map_err(|e| {
```
phiSgr, line 60: I believe `sqlparser` rejects some CH SQL statements. Maybe use its error message when CH rejects it?

onelesd: there seem to be a few specific CH syntax not supported. here's one: https://github.com/apache/datafusion-sqlparser-rs/issues/1514

the syntax validation gate in this PR would block those from ever getting to CH which is probably Not Good.

i wonder if it'd be better to not syntax check at all...

PR #3072, opened 2025-12-03
Move Lifecycle protection to the end

georgevanderson:
Move lifecycle protection filtering to the end. Previously, clickhouse strategy could upgrade an Update to a DROP/CREATE after lifecycle filtering

---

> [!NOTE]
> Lifecycle protection is enforced after table diff strategies run, blocking DROP/CREATE for protected tables and adding tests for these scenarios.
> 
> - **Core (table diffing)**:
>   - Move lifecycle filtering to occur after `strategy.diff_table_update`, not before.
>   - Block `TableChange::Removed` and corresponding orphan `Added` when target table is `DeletionProtected` or `ExternallyManaged`.
>   - For `Updated` on `DeletionProtected`, filter out `ColumnChange::Removed` only.
>   - Use target (`after`) lifecycle for decisions; only count/apply non-empty filtered changes.
> - **Tests**:
>   - Add `DropCreateStrategy` mock and comprehensive tests covering: blocking strategy-generated DROP/CREATE for `DeletionProtected` and `ExternallyManaged`, allowing for `FullyManaged`, lifecycle transition to protected, and disabling protection via flag.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c5e1f06c0dce16d1e9d5e804bb61d3a71fb5863f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/framework/core/infrastructure_map.rs
+++ b/apps/framework-cli/src/framework/core/infrastructure_map.rs
@@ -1871,20 +1871,25 @@ impl InfrastructureMap {
 
                             // Filter strategy changes to respect lifecycle constraints
                             // This handles both:
-                            // 1. Strategies that convert updates to drop+create (blocks Removed for protected tables)
+                            // 1. Strategies that convert updates to drop+create (blocks both Removed AND Added for protected tables)
                             // 2. Strategies that return Updated (filters out column removals for DeletionProtected tables)
                             let filtered_changes: Vec<OlapChange> = if respect_life_cycle {
+                                let mut blocked_tables = std::collections::HashSet::new();
+
                                 strategy_changes
                                     .into_iter()
                                     .filter_map(|change| {
                                         match change {
                                             OlapChange::Table(TableChange::Removed(removed_table)) => {
-                                                match removed_table.life_cycle {
+                                                // CRITICAL: Use target_table lifecycle (AFTER state), not removed_table lifecycle (BEFORE state)
+                                                // This handles transitions TO protected lifecycles (e.g., FullyManaged -> DeletionProtected)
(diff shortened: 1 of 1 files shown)


phiSgr (approved):

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -1895,10 +1869,110 @@ impl InfrastructureMap {
                                 default_database,
                             );
 
+                            // Filter strategy changes to respect lifecycle constraints
+                            // This handles both:
+                            // 1. Strategies that convert updates to drop+create (blocks both Removed AND Added for protected tables)
+                            // 2. Strategies that return Updated (filters out column removals for DeletionProtected tables)
+                            let filtered_changes: Vec<OlapChange> = if respect_life_cycle {
```
callicles, line 1876: Hum this is a better place - but I think we should pull it out of the diff completely and have it applied in another phase of the infra map compiler. 

@LucioFranco Woudl you have a suggestion? The issue here is we produce the plan for changes as the set of diff for the infrastructure. That Plan should be sanitized with lifecycle rules which enfore what kind of operations are allowed on a Table / Object. 

Right now it is embeeded in the diff logic and it is prone for developper mistakes and doesn't get enforced no matter what. I am thinking it should wrap the internal planning methods which should become private to the inner module to make sure other modules don't call it without the filtering.

Another idea is triggered in which we could use a visitor pattern to prune the plan? or and build it maybe? But that could be another iteration on the refactor

georgevanderson:
Closed due to complete lifecycle overhaul

PR #3090, opened 2025-12-05
Add Dockerfile customization support for Docker builds

cjus:
Add ability for users to expose and customize the Dockerfile for their Moose projects, enabling installation of system packages in Docker containers.

New CLI flags:
  - `--expose-dockerfile`: Generates Dockerfile to project root for customization
  - `--no-expose-dockerfile`: Reverts to managed mode, deletes custom Dockerfile

Behavior:
  - Custom Dockerfile at ./Dockerfile is auto-detected and preserved
  - "Generate once, preserve forever" - existing customizations are never overwritten
  - Custom Dockerfile is copied to .moose/packager/ during build for compatibility and use in Boreal
  - Config file and env var support: `dev.expose_dockerfile` / `MOOSE_DEV__EXPOSE_DOCKERFILE`

---

> [!NOTE]
> Adds flags and config to expose a project-root Dockerfile for customization, preserves existing files, and updates build to use/copy it (incl. monorepos).
> 
> - **CLI (build)**:
>   - Add `--expose-dockerfile` and `--no-expose-dockerfile` flags (require `--docker`, mutually exclusive).
>   - Build flow can skip image build when exposing only; otherwise builds for selected arches.
> - **Docker packager**:
>   - `create_dockerfile` now respects `settings.dev.expose_dockerfile` and CLI flags; resolves Dockerfile target (custom vs managed), preserves existing custom files, and can remove custom on `--no-expose-dockerfile`.
>   - Records custom mode via marker; shows informative messages; adds helpers for path resolution and safe directory creation.
>   - `build_dockerfile` detects custom Dockerfile, copies/uses it for build context; supports safe temp Dockerfile names; improved cleanup.
>   - Refactors/fallbacks: centralized content writing, standard-build context setup; removes old write-to-managed-only helper.
> - **Settings**:
>   - Add `dev.expose_dockerfile` (env: `MOOSE_DEV__EXPOSE_DOCKERFILE`) to control exposure by default.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 364fb0e37fb955671b8f9593e8be0b35d3780faf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


cjus:
Working on addressing reported bugs above.

apps/framework-cli/src/cli/commands.rs
```diff
@@ -52,6 +52,13 @@ pub enum Commands {
         /// Build for arm64 architecture
         #[arg(long)]
         arm64: bool,
+        /// Expose Dockerfile to project root for customization
+        /// generates Dockerfile at ./Dockerfile instead of hidden .moose/packager/
+        #[arg(long, requires = "docker", conflicts_with = "no_expose_dockerfile")]
+        expose_dockerfile: bool,
+        /// Revert to hidden, managed Dockerfile (default behavior)
+        #[arg(long, requires = "docker", conflicts_with = "expose_dockerfile")]
+        no_expose_dockerfile: bool,
```
LucioFranco: I find these arguments to be a bit strange to use and since we are adding this to our public interface I am hesitant since any changes here would be breaking. 

I wonder if we let users configure it via moose.config.toml where they can enable and set the path of the dockerfile for their project and then provide a way to generate it via moose init? Essentially, overriding the config would move moose from using the file in .moose/packager to the one specified by the moose config for each moose app.

cjus: @LucioFranco I was seeing this as permanent approach and not something that we would actually remove but I see your point, and I like the idea of having it be on a per project basis. So I'll go ahead and refactor this a bit.

PR #3095, opened 2025-12-06
Rework operations

georgevanderson:
> [!NOTE]
> Adds lifecycle-aware filtering/guarding of OLAP changes, canonical table handling and stable table IDs, unifies plan-to-operations flow, and expands E2E coverage; also bumps ClickHouse images and standardizes test env/timeouts.
> 
> - **Core/Planning**:
>   - Enforce lifecycle policies (ExternallyManaged/DeletionProtected) via new lifecycle filter and execution guard; surface blocked changes in `filtered_olap_changes` and display them.
>   - Add `Table::canonicalize` and apply across loads/reconciliation for backward compatibility (ORDER BY fallback, non-null arrays, PK clearing on unsupported engines).
>   - Stabilize table identity: `Table::id` always includes database; remove legacy dotted-name handling.
>   - Refactor planning: add `infra_changes_to_operations` and `load_target_infrastructure`; remote plan/migration now use shared loaders and ClickHouse diff strategy.
>   - Improve reconciliation: reconcile current map once, adopt unmapped targets safely; adjust admin routes (plan deprecated) and reconciliation path.
> - **CLI/Display & Server**:
>   - Show filtered (blocked) changes; tweak logs/messages; improve server admin reconciliation.
> - **E2E/Tests**:
>   - New lifecycle E2E suite and `plan-utils`; broaden utils (process handling), increase timeouts, standardize `MOOSE_DEV__SUPPRESS_DEV_SETUP_PROMPT` in env.
>   - Update defaults tests to expect UInt64 numbers; remove cloud-only replicated table cases.
> - **Templates/Config**:
>   - Add lifecycle test models and admin API key to TS template; remove `ReplicatedMergeTreeCloudTest`.
> - **Infra/Deps**:
>   - Bump ClickHouse images to `25.8` in docker-compose templates.
>   - Introduce pnpm workspace catalogs; lockfile updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1ec0a2b83dcafb6b85e56732b92fd11253882fc2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/framework/core/infrastructure/table.rs
+++ b/apps/framework-cli/src/framework/core/infrastructure/table.rs
@@ -1863,55 +1863,23 @@ mod tests {
         };
         assert_eq!(table2b.id(DEFAULT_DATABASE_NAME), "analytics_users");
 
-        // Test 3: Legacy format - table name contains database prefix (backward compatibility)
+        // Test 3: With version - database should be included
         let table3 = Table {
-            name: "analytics.users".to_string(),
-            database: None,
+            name: "users".to_string(),
+            version: Some(Version::from_string("1.0".to_string())),
+            database: Some("analytics".to_string()),
             ..table1.clone()
         };
-        assert_eq!(table3.id(DEFAULT_DATABASE_NAME), "analytics.users");
+        assert_eq!(table3.id(DEFAULT_DATABASE_NAME), "analytics_users_1_0");
 
-        // Test 4: CRITICAL - Adding database field to legacy format should NOT change ID
+        // Test 4: With version and default database
         let table4 = Table {
-            name: "analytics.users".to_string(),
-            database: Some("analytics".to_string()),
-            ..table1.clone()
-        };
-        assert_eq!(
-            table4.id(DEFAULT_DATABASE_NAME),
-            "analytics.users",
-            "ID should remain stable when database field is added to legacy table name format"
-        );
-
-        // Test 5: Even with mismatched database, ID should remain stable (name takes precedence)
-        let table5 = Table {
-            name: "analytics.users".to_string(),
-            database: Some("other_db".to_string()),
-            ..table1.clone()
-        };
-        assert_eq!(
-            table5.id(DEFAULT_DATABASE_NAME),
-            "analytics.users",
-            "ID should use name (which contains dot) even if database field differs"
-        );
-
-        // Test 6: With version - database should be included
-        let table6 = Table {
-            name: "users".to_string(),
-            version: Some(Version::from_string("1.0".to_string())),
-            database: Some("analytics".to_string()),
-            ..table1.clone()
-        };
-        assert_eq!(table6.id(DEFAULT_DATABASE_NAME), "analytics_users_1_0");
-
-        // Test 7: With version and default database
-        let table7 = Table {
             name: "users".to_string(),
             version: Some(Version::from_string("1.0".to_string())),
             database: None,
             ..table1.clone()
         };
-        assert_eq!(table7.id(DEFAULT_DATABASE_NAME), "local_users_1_0");
+        assert_eq!(table4.id(DEFAULT_DATABASE_NAME), "local_users_1_0");
     }
 
     #[test]


apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -3292,41 +3294,46 @@ async fn get_admin_reconciled_inframap(
...
-        Ok(None) => InfrastructureMap::empty_from_project(project),
-        Err(e) => {
-            return Err(crate::framework::core::plan::PlanningError::Other(
-                anyhow::anyhow!("Failed to load infrastructure map from state storage: {e}"),
-            ));
-        }
-    };
-
     if !project.features.olap {
+        // If OLAP is disabled, we can't reconcile, just return stored state
```
callicles: That's not right - the infraamp also has kafka etc... and if it is not already part of reconciliation it should eventually be part of it - we should trigger it even though it would not do anything

apps/framework-cli/src/framework/core/lifecycle_filter.rs
```diff
@@ -0,0 +1,200 @@
...
+pub fn apply_lifecycle_filter(changes: Vec<OlapChange>, target_table: &Table) -> FilterResult {
+    let mut applied = Vec::new();
+    let mut filtered = Vec::new();
+    let mut blocked_tables = HashSet::new();
+
+    for change in changes {
+        match change {
+            OlapChange::Table(TableChange::Removed(removed_table)) => {
+                if should_block_table_removal(&removed_table, target_table) {
+                    blocked_tables.insert(removed_table.name.clone());
```
callicles: You should use the id intead of the name here. In a multi database scenario we might have duplicate tables across databases

apps/framework-cli-e2e/test/utils/schema-definitions.ts
```diff
@@ -146,17 +146,6 @@ export const TYPESCRIPT_TEST_SCHEMAS: ExpectedTableSchema[] = [
       { name: "isDeleted", type: "Bool" },
     ],
   },
-  {
-    tableName: "ReplicatedMergeTreeCloudTest",
```
callicles, line 150: This is not supposed to work locally so removed it

apps/framework-cli/src/framework/core/plan.rs
```diff
@@ -332,23 +342,72 @@ pub struct InfraPlan {
...
+///
+/// # Arguments
+/// * `project` - The project configuration
+///
+/// # Returns
+/// * `Result<InfrastructureMap, PlanningError>` - The target infrastructure map with resolved credentials
+pub async fn load_target_infrastructure(
     project: &Project,
-) -> Result<(InfrastructureMap, InfraPlan), PlanningError> {
+) -> Result<InfrastructureMap, PlanningError> {
```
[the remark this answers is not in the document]
georgevanderson: I don't think this is a real worry with how we write the JSON since the newest version is what writes it, but maybe we want to add anyways?

apps/framework-cli/src/framework/core/plan.rs
```diff
@@ -379,7 +438,51 @@ pub async fn plan_changes(
             ))
         })?;
 
-    let current_infra_map = state_storage.load_infrastructure_map().await?;
+    Ok(target_infra_map)
```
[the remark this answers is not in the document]
callicles: Those are not used anymore

apps/framework-cli/src/framework/core/lifecycle_filter.rs
```diff
@@ -0,0 +1,1257 @@
...
+        }
+        _ => applied.push(change),
+    }
+}
+
+/// Filters a table removal operation
+fn filter_table_removal(
+    removed_table: Table,
+    target_table: &Table,
+    default_database: &str,
```
LucioFranco, line 147: This is a leaky abstraction, right?

apps/framework-cli/src/framework/core/lifecycle_filter.rs
```diff
@@ -0,0 +1,1257 @@
...
+/// Filters a table addition operation (blocks orphan creates)
+fn filter_table_addition(
+    added_table: Table,
+    default_database: &str,
+    blocked_table_ids: &HashSet<String>,
+    applied: &mut Vec<OlapChange>,
+    filtered: &mut Vec<FilteredChange>,
+) {
+    if blocked_table_ids.contains(&added_table.id(default_database)) {
+        tracing::debug!(
```
LucioFranco, line 172: should we include debugs for the rest? I could see how this might be useful, can be done as a follow up easily

apps/framework-cli/src/framework/core/lifecycle_filter.rs
```diff
@@ -0,0 +1,1257 @@
...
+            &name,
+            blocked_columns,
+            order_by_change.clone(),
+            partition_by_change.clone(),
+            before.clone(),
+            after.clone(),
+        ));
+    }
+
+    // Always pass through the Updated change with filtered column_changes
```
LucioFranco, line 219: could use a why here

apps/framework-cli/src/framework/core/lifecycle_filter.rs
```diff
@@ -0,0 +1,1257 @@
...
+            tracing::error!(
+                "LIFECYCLE GUARD VIOLATION: {} (table: {}, lifecycle: {:?}, type: {:?})",
+                violation.message,
+                table_display,
+                violation.life_cycle,
+                violation.violation_type
+            );
+        }
+        Err(violations)
+    }
```
LucioFranco: style nit: I would swap this make `if !violations.is_empty()` and then have the function always return Ok(()) and have the if branch return a `return Err(..)`

apps/framework-cli/src/framework/core/lifecycle_filter.rs
```diff
@@ -0,0 +1,1257 @@
...
+        ColumnChange::Updated { after: col, .. } => {
+            (col.name.clone(), ViolationType::ColumnModification)
+        }
+    };
+
+    LifecycleViolation {
+        message: format!(
+            "Attempted to modify column '{}' on table '{}' which has \
+            {:?} lifecycle. This is a bug - the diff/filter \
+            pipeline should have blocked this.",
```
LucioFranco, line 620: we can add a `please open an issue in - https://github.com/514-labs/moosestack/issues/new` to get users to actually create a bug report easily?

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -1988,10 +2012,12 @@ impl InfrastructureMap {
         default_database: &str,
     ) {
         let default_strategy = DefaultTableDiffStrategy;
+        let mut filtered = Vec::new(); // Discard filtered changes for backward compatibility
```
LucioFranco: this doesn't explain why and its not clear

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -2077,7 +2103,8 @@ impl InfrastructureMap {
     ///
     /// # Returns
     /// * `Option<TableChange>` - None if identical, Some(change) if different
-    pub fn simple_table_diff_with_lifecycle(
+    #[cfg(test)]
```
LucioFranco, line 2142: should be moved into the test module where its used

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -2174,15 +2201,18 @@ impl InfrastructureMap {
...
     /// # Arguments
     /// * `path` - The path to the JSON file
     ///
     /// # Returns
     /// A Result containing either the loaded map or an IO error
     pub fn load_from_json(path: &Path) -> Result<Self, std::io::Error> {
         let json = fs::read_to_string(path)?;
-        let infra_map = serde_json::from_str(&json)?;
-        Ok(infra_map)
+        let infra_map: InfrastructureMap = serde_json::from_str(&json)?;
```
LucioFranco, line 2250: style nit: I prefer https://turbo.fish/ `from_str::<InfastructureMap>(..)`

apps/framework-cli/src/framework/core/infra_reality_checker.rs
```diff
@@ -124,6 +124,10 @@ impl<T: OlapOperations> InfraRealityChecker<T> {
     /// exist in reality but not in the map, tables that are in the map but don't exist
     /// in reality, and tables that exist in both but have structural differences.
     ///
+    /// TODO add support for kafka reality check
+    /// TODO this is too big of a function, we should split it into smaller functions, there
```
LucioFranco, line 128: Is this related to this change?

apps/framework-cli/src/utilities/docker-compose.yml.hbs
```diff
@@ -127,7 +127,7 @@ services:
         exec /entrypoint.sh
 
   clickhousedb:
-    image: docker.io/clickhouse/clickhouse-server:${CLICKHOUSE_VERSION:-25.6}
+    image: docker.io/clickhouse/clickhouse-server:${CLICKHOUSE_VERSION:-25.8}
```
LucioFranco, line 130: I don't think we mean to bump this?

LucioFranco (approved):
I've left a lot of comments but they are not blocking, I would like someone to follow up on them.

LucioFranco (approved):

apps/framework-cli/src/framework/core/plan.rs
```diff
@@ -404,9 +504,47 @@ pub async fn plan_changes(
...
+    // Load target state from project code
+    let target_infra_map = load_target_infrastructure(project).await?;
+
+    // Load and reconcile current state
+    let olap_client = clickhouse::create_client(project.clickhouse_config.clone());
+    let target_table_ids: HashSet<String> = target_infra_map
+        .tables
+        .values()
+        .map(|t| t.id(&target_infra_map.default_database))
+        .collect();
```
phiSgr, line 536: we don't need this defensiveness because `load_target_infrastructure` calls `fixup_default_db`

LucioFranco: We agreed though that we need to fix this code in a follow up so this shouldn't block us here.

apps/framework-cli/src/framework/core/infrastructure_map.rs
phiSgr: maybe next time :P

apps/framework-cli/src/framework/core/lifecycle_filter.rs
phiSgr: maybe next time

apps/framework-cli/src/framework/core/lifecycle_filter.rs
phiSgr: true :(

PR #3116, opened 2025-12-10
switch to consumer

georgevanderson:


georgevanderson:
Closed because of our patch version of this kafka library. If we have upstream merge issues, we may need to go down this route.

PR #3142, opened 2025-12-16
fix enums_are_equivalent when order doesn't match

DatGuyJonathan:
---

> [!NOTE]
> Make `enums_are_equivalent` compare members by name/value (order-insensitive) and add tests for varied ordering across string/int/metadata cases.
> 
> - **ClickHouse diff strategy (`diff_strategy.rs`)**:
>   - **Enum equivalence**: Update `enums_are_equivalent` to compare members by name/value instead of index (order-insensitive), including string↔int cross-mapping for TypeScript string enums vs ClickHouse.
> - **Tests**:
>   - Add coverage for integer enums with different order.
>   - Add coverage for string enums with different order (cross-mapped to ints).
>   - Add coverage for metadata-present enums with different order.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 148ae59c816f22e7a7d722f75c97b581b14bd964. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>
 
 <div id='description'>
<a href="https://bito.ai#summarystart"></a>
<h3>Summary by Bito</h3>

<ul>

<li>Fixes a bug in the enum equivalence functionality by updating the logic to compare enum members in an order-insensitive manner.</li>

<li>Removes reliance on member indexing and clarifies matching conditions between string and integer enums, including handling cross-mapping scenarios between TypeScript and ClickHouse enums.</li>

<li>Introduces new tests covering different ordering scenarios for both string and integer enums, as well as enums with metadata.</li>

<li>Overall summary: Fixes a bug in enum equivalence functionality, removes reliance on member indexing, clarifies matching conditions for string and integer enums, handles cross-mapping between TypeScript and ClickHouse enums, introduces new tests, and ensures robust and accurate enum comparisons in the system.</li>

</ul>

</div>

### DIFF
--- a/apps/framework-cli/src/infrastructure/olap/clickhouse/diff_strategy.rs
+++ b/apps/framework-cli/src/infrastructure/olap/clickhouse/diff_strategy.rs
@@ -99,8 +99,8 @@ pub fn enums_are_equivalent(actual: &DataEnum, target: &DataEnum) -> bool {
         return false;
     }
 
-    // Check each member
-    for (idx, target_member) in target.values.iter().enumerate() {
+    // Check each member - compare by name/value, not by index (order-insensitive)
+    for target_member in &target.values {
         match &target_member.value {
             EnumValue::String(target_str) => {
                 // For string enums, we have two cases:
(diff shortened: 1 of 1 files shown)


DatGuyJonathan:
@coderabbitai review

DatGuyJonathan:
> BTW: i think the `String` or `Int` should be in the `values` level. it makes no sense to have the list where one is `String` and another is `Int`. if we fix that the comparison logic should be a fair bit simpler.

That's true. Would be a decent sized pr for backward compatibility (we store metadata in column comments), so I created a ticket for it.

PR #3184, opened 2026-01-12
ENG-1184: fix `moose peek` to look up tables/topics correctly

LucioFranco:
> [!NOTE]
> Fixes resource resolution and query building in `peek`.
> 
> - Resolve `tables`/`topics` by their `name` (case-insensitive) via `find_table_by_name` and `find_topic_by_name` instead of map keys
> - Error messages now include the requested name and a list of available resources when not found
> - ClickHouse SELECT uses a table's explicit `database` when provided, otherwise falls back to project default
> - Adds unit tests for lookups and database selection behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f6f52f3adb6b99cf440b33e56b573794cea167ab. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/cli/routines/peek.rs
+++ b/apps/framework-cli/src/cli/routines/peek.rs
@@ -348,7 +348,10 @@ mod tests {
         }
     }
 
-    fn create_test_topic(name: &str, version: Option<crate::framework::versions::Version>) -> Topic {
+    fn create_test_topic(
+        name: &str,
+        version: Option<crate::framework::versions::Version>,
+    ) -> Topic {
         Topic {
             name: name.to_string(),
             version,


apps/framework-cli/src/cli/routines/peek.rs
```diff
@@ -277,3 +282,229 @@ pub async fn peek(
...
+
+    #[test]
+    fn test_table_lookup_by_name() {
+        let infra = create_test_infra();
+
+        // Should find table by name, not by HashMap key
+        let table = infra
+            .tables
+            .values()
+            .find(|table| table.name.to_lowercase() == "users".to_lowercase());
```
DatGuyJonathan: It looks like these tests assume the find logic is implemented the same way. Maybe we need to extract those to functions, then the tests can call that instead.

PR #3186, opened 2026-01-12
[1/n] ENG-1892: add logging infra for moose aware logs

LucioFranco:
> [!NOTE]
> Adds P0 structured logging as an opt-in path alongside existing legacy/modern formats.
> 
> - New `LoggerSettings.structured_logs` (env: `MOOSE_LOGGER__STRUCTURED_LOGS`) enables JSON logs including current span fields for filtering (e.g., `context`, `resource_type`, `resource_name`)
> - Introduces `setup_structured_logs` using `tracing-subscriber` JSON formatter with current-span output for stdout or date-rotated files
> - Exposes constants in `context` and `resource_type` modules for consistent span annotations
> - Defaults `use_tracing_format` to `false` (no longer read from env at default), preserves legacy/modern selection order
> - Adds tests verifying span fields in JSON, valid logs without spans, and exported constants
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a7020689a382e2da9b7b7c1c19eda0429cc11b2d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,29 @@ updates:
     labels:
       - "rust"
       - "toolchain"
+
+  # =============================================================================
+  # Templates - Monitor all dependencies using glob patterns
+  # =============================================================================
+
+  # TypeScript templates (any package.json in templates/)
+  - package-ecosystem: "npm"
+    directories:
+      - "/templates/**"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    labels:
+      - "templates"
+      - "dependencies"
+
+  # Python templates (any requirements.txt in templates/)
+  - package-ecosystem: "pip"
+    directories:
+      - "/templates/**"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    labels:
+      - "templates"
+      - "dependencies"
--- a/apps/framework-cli/src/cli/routines/peek.rs
+++ b/apps/framework-cli/src/cli/routines/peek.rs
@@ -4,6 +4,8 @@
 //! either database tables or streaming topics for debugging and exploration purposes.
 
 use crate::cli::display::Message;
+use crate::framework::core::infrastructure::table::Table;
+use crate::framework::core::infrastructure::topic::Topic;
 use crate::framework::core::infrastructure_map::InfrastructureMap;
 use crate::infrastructure::olap::clickhouse::mapper::std_table_to_clickhouse_table;
 use crate::infrastructure::olap::clickhouse_http_client::create_query_client;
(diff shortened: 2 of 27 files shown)


apps/framework-cli/src/cli/logger.rs
```diff
@@ -96,6 +96,30 @@ use std::sync::atomic::Ordering;
 
 use super::settings::user_directory;
 
+/// P0 logging context constants.
+/// Used in #[instrument(fields(context = ...))]
+#[allow(dead_code)]
+pub mod context {
+    pub const RUNTIME: &str = "runtime";
+    pub const DEPLOY: &str = "deploy";
+    pub const SYSTEM: &str = "system";
```
callicles, line 105: Can you explain the differences between those? and when used?

I am not sure how we will have `Deploy` in moose context for example?

apps/framework-cli/src/cli/logger.rs
```diff
@@ -467,8 +491,15 @@ pub fn setup_logging(settings: &LoggerSettings) {
 
     let session_id = CONTEXT.get(CTX_SESSION_ID).unwrap();
 
-    // Setup logging based on format type
-    if settings.use_tracing_format {
+    // Check for P0 structured logs with span support
+    let use_structured_logs = std::env::var("MOOSE_STRUCTURED_LOGS")
```
callicles: Can we use the logger settings param? which is input from settings and reads from env variables?

apps/framework-cli/src/cli/logger.rs
LucioFranco: Deploy is supposed to capture logs that are related to the startup of moose, so things that should be captured for why my deployment failed. Runtime is for all the logs that run while moose is up and running post migration/deploy. System is supposed to capture all the rest of the logs like redis stuff, etc.

callicles: Can we rename `Deploy` to `Boot`? that feels that expresses better what you want. 
And will work with dev mode

apps/framework-cli/src/cli/logger.rs
```diff
@@ -181,6 +209,13 @@ fn default_no_ansi() -> bool {
     false // ANSI colors enabled by default
 }
 
+fn default_structured_logs() -> bool {
+    env::var("MOOSE_LOGGER__STRUCTURED_LOGS")
```
callicles: I don't think you need this

We already automatically load env variable here through the Config.rs crate 
look at `read_settings`

You just need the default value

LucioFranco: good call haha fixed this and tracing_format

callicles (approved):

PR #3321, opened 2026-01-16
fix: typescript boot times by precompiling

LucioFranco:
> [!NOTE]
> Introduces precompiled TypeScript workflow and dual-module (ESM/CJS) support across the CLI and ts library, with Docker/CI integration.
> 
> - Adds `moose-tspc` CLI and shared `compiler-config` to precompile app code, auto-detect module system, and provide a `loadModule` loader; includes ESM `.js` import rewriting (`rewriteImportExtensions`).
> - Updates runtime (`moose-runner`, blocks/apis/DMv2/streaming runners, serializers) to honor `MOOSE_USE_COMPILED` and load compiled JS via `loadModule`; production paths set `MOOSE_USE_COMPILED=true`.
> - Extends Docker packager to precompile TS during image build (`MOOSE_SOURCE_DIR`, `moose-tspc`) for both monorepo and standalone flows; sets `ENV MOOSE_USE_COMPILED=true`.
> - Adds E2E tests for module compilation (CJS default, ESM with `type: module`, and ESM import extension rewriting) and a new `test-e2e-module-compilation` job wired into final status checks.
> - Exposes `moose-tspc` in `@514labs/moose-lib` bin and minor workflow/lockfile adjustments.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e2c4f097de5a0fdf759d06c332a882b95775bad3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/ts-moose-lib/src/dmv2/internal.ts
+++ b/packages/ts-moose-lib/src/dmv2/internal.ts
@@ -1369,10 +1369,6 @@ export const getStreamingFunctions = async () => {
       transforms.forEach(([_, transform, config]) => {
         const transformFunctionKey = `${stream.name}_${destinationName}${config.version ? `_${config.version}` : ""}`;
         compilerLog(`getStreamingFunctions: ${transformFunctionKey}`);
-        // DEBUG: Log the function type to verify it's not undefined after compilation
-        compilerLog(
-          `[DEBUG] Function type for ${transformFunctionKey}: type=${typeof transform}, isFunction=${typeof transform === "function"}`,
-        );
         transformFunctions.set(transformFunctionKey, [
           transform,
           config,
--- a/packages/ts-moose-lib/src/streaming-functions/runner.ts
+++ b/packages/ts-moose-lib/src/streaming-functions/runner.ts
@@ -289,12 +289,7 @@ const handleMessage = async (
     }
 
     const transformedData = await Promise.all(
-      streamingFunctionWithConfigList.map(async ([fn, config], idx) => {
-        // DEBUG: Log before invoking the transform function
-        logger.log(
-          `[DEBUG] Calling transform function ${idx}: type=${typeof fn}, isFunction=${typeof fn === "function"}`,
-        );
-
+      streamingFunctionWithConfigList.map(async ([fn, config]) => {
         try {
           return await fn(parsedData);
         } catch (e) {
@@ -728,16 +723,6 @@ const startConsumer = async (
     );
     streamingFunctions = result.functions;
     fieldMutations = result.fieldMutations;
-
-    // DEBUG: Log loaded functions to verify they're valid after compilation
-    logger.log(
-      `[DEBUG] Loaded ${streamingFunctions.length} streaming function(s)`,
-    );
-    streamingFunctions.forEach(([fn, config], i) => {
-      logger.log(
-        `[DEBUG] Function ${i}: type=${typeof fn}, isFunction=${typeof fn === "function"}`,
-      );
-    });
   } else {
     streamingFunctions = [
       [await loadStreamingFunction(args.functionFilePath), {}],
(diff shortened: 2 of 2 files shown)


packages/ts-moose-proto/scripts/release-lib.sh
```diff
@@ -6,6 +6,9 @@ set -eo pipefail
 
 version=$1
 
+# Wait 5 seconds before publishing
+sleep 5
+
```
georgevanderson: can you add a comment for why

LucioFranco: https://github.com/514-labs/moosestack/actions/runs/21051970301/job/60540126116?pr=3321 this is why

packages/ts-moose-lib/src/moose-tspc.ts
```diff
@@ -0,0 +1,110 @@
...
+  writeFileSync(tempTsconfigPath, JSON.stringify(buildTsconfig, null, 2));
+  console.log("Created temporary tsconfig with moose plugins...");
+
+  // Use tspc (ts-patch compiler CLI) with the temporary tsconfig
+  // Include source maps for better error messages in production
+  // --rootDir . preserves directory structure (e.g., app/index.ts -> outDir/app/index.js)
+  // Note: skipLibCheck and noEmitOnError are in compilerOptions
+  try {
+    execSync(
+      `npx tspc -p ${tempTsconfigPath} --outDir ${outDir} --rootDir . --sourceMap --inlineSources`,
```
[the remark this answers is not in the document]
LucioFranco: fixed

packages/ts-moose-lib/src/commons.ts
```diff
@@ -109,6 +117,138 @@ export function mapTstoJs(filePath: string): string {
...
+        }
+      } else if (entry.isFile()) {
+        const ext = nodePath.extname(entry.name);
+        if (extensions.includes(ext)) {
+          results.push(fullPath);
+        }
+      }
+    }
+  } catch {
+    // Silently skip directories we can't read
```
onelesd: log the error in case it is something the user can/should act on

packages/ts-moose-lib/src/commons.ts
```diff
@@ -109,6 +117,138 @@ export function mapTstoJs(filePath: string): string {
...
+/**
+ * Rewrites a single import path to add .js extension if needed.
+ *
+ * @param match - The full matched string
+ * @param prefix - The prefix (from ' or import(')
+ * @param importPath - The import path
+ * @param quote - The closing quote
+ * @returns The rewritten import or original if no change needed
+ */
+function rewriteImportPath(
```
onelesd, line 219: hard for me to keep the whole context here but i think this would break:

```
import { helper } from "./utils"
```

when the file imported is an index:
./utils/index.js

because the rewritten import would be:
./utils.js

packages/ts-moose-lib/src/moose-tspc.ts
```diff
@@ -0,0 +1,132 @@
...
+      // Skip type checking of declaration files to avoid dual-package conflicts
+      // This must be in compilerOptions (not just CLI flag) to fully work
+      skipLibCheck: true,
+      skipDefaultLibCheck: true,
+      // Additional settings to handle module resolution conflicts
+      allowSyntheticDefaultImports: true,
+      // CRITICAL: Emit JavaScript even when there are type errors
+      // This is essential for Docker builds where we need compilation to succeed
+      // Type errors are acceptable here since the code works at runtime
+      noEmitOnError: false,
```
onelesd, line 67: already called out but this is risky

is it potentially doing harm by letting code likely to crash at runtime pass build?

LucioFranco: We have a npx moose check that happens before so I think this is less risky.

LucioFranco: Updated adding a warning as well

apps/framework-cli/src/cli/routines/docker_packager.rs
```diff
@@ -1407,7 +1436,12 @@ fn create_standard_typescript_dockerfile_content(
             "COPY_PACKAGE_FILE",
             &format!("\n                    {copy_section}"),
         )
-        .replace("INSTALL_COMMAND", &install_command);
+        .replace("INSTALL_COMMAND", &install_command)
+        // Pre-compile TypeScript with moose plugins for faster worker startup
+        .replace(
+            "TYPESCRIPT_COMPILE_STEP",
+            &generate_typescript_compile_step(&project.source_dir),
+        );
```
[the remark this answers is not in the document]
LucioFranco: already have a test and these env vars are not for end user use

LucioFranco:
I've tested this against my hosting telemetry preview branch all seems to work including streaming functions 👍

PR #3403, opened 2026-01-23
Fix precompilation for workflows

DatGuyJonathan:
> [!NOTE]
> **Fix precompilation behavior**
> 
> - Updates `loadIndex` to clear `moose_internal` registries and `require.cache` only when not using compiled artifacts, preserving state in compiled (production) mode
> - Maintains hot-reload behavior in development while allowing precompiled workflows to load without cache invalidation
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cf2cb7384cb3a27e92f34abea4d14a291145cbbd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/ts-moose-lib/src/dmv2/internal.ts
+++ b/packages/ts-moose-lib/src/dmv2/internal.ts
@@ -1277,11 +1277,10 @@ const loadIndex = async () => {
   // providing automatic fallback to ts-node if compilation wasn't run.
   const useCompiled = shouldUseCompiled();
 
-  const registry = getMooseInternal();
-
   // In dev mode, clear registry and require.cache to support hot reloading.
   // In production (compiled mode), skip clearing - code doesn't change.
   if (!useCompiled) {
+    const registry = getMooseInternal();
     registry.tables.clear();
     registry.streams.clear();
     registry.ingestApis.clear();


packages/ts-moose-lib/src/dmv2/internal.ts
```diff
@@ -1277,20 +1277,20 @@ const loadIndex = async () => {
   // providing automatic fallback to ts-node if compilation wasn't run.
   const useCompiled = shouldUseCompiled();
 
-  // Clear the registry before loading to support hot reloading
```
DatGuyJonathan, line 1280: Seeing workflow can't be found errors in prod. Workflows need to get the functions repeatedly due to the way temporal isolates it https://github.com/514-labs/moosestack/blob/main/packages/ts-moose-lib/src/scripts/activity.ts#L110. Whereas streaming & apis can just get it once when the runner starts up.

I think this was clearing the registry on subsequent calls, but in prod modules are cached after first load and don't re-execute, so workflows weren't re-registered. Tested the ci build and streaming, consumption, and workflows are working.

PR #3406, opened 2026-01-26
simplify ts compiler plugin to fix incremental compilation

phiSgr:
old behaviour was injecting `typia.something<Foo>()` into the ts file, then let typia plugin handle it afterwards.
problem with that it messed up the incremental compilation.
now we directly invoke the code generation in typia.

also, cleaned up some DMV1 code (the per file query API)

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it rewrites the TypeScript compiler plugin and schema/validation generation path (typia integration, import injection, and transformer wiring), which can affect build behavior and generated OpenAPI/validation at runtime across all TS projects.
> 
> **Overview**
> Switches the Moose TypeScript compiler plugin from a `transformProgram`/program-replacement approach to a regular transformer that **directly invokes typia’s internal code generators** (assert/validate/is, HTTP query parsing, and JSON schema generation) and injects required typia imports per-file.
> 
> Updates config and templates to remove `transformProgram: true` (and related `isolatedModules` usage), adds post-processing of generated JSON schemas to emit proper `Date` as `string`/`date-time` and strip internal ClickHouse tag properties, and removes legacy DMV1 `createApi` exports/implementation and unused framework CLI streaming/template generation code.
> 
> Enhances CLI e2e tests by adding OpenAPI YAML sanity checks (TS + Python) via a new `managementUrl`, and adds `js-yaml` dependencies to parse the spec.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4dca74c6c5e8d33d5cf68186ecdf28be9b7c8dd4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/ts-moose-lib/src/typiaDirectIntegration.ts
+++ b/packages/ts-moose-lib/src/typiaDirectIntegration.ts
@@ -63,9 +63,7 @@ export interface TypiaDirectContext {
  * the node to be attached to a source file. We patch the method directly.
  */
 const createSyntheticModulo = (): ts.LeftHandSideExpression => {
-  const identifier = ts.factory.createIdentifier("typia") as ts.Identifier & {
-    getText: () => string;
-  };
+  const identifier = ts.factory.createIdentifier("typia");
 
   // Monkey-patch getText to return "typia" directly
   identifier.getText = () => "typia";


packages/ts-moose-lib/src/compiler-config.ts
```diff
@@ -13,9 +13,11 @@ import path from "path";
 export const MOOSE_COMPILER_PLUGINS = [
   {
     transform: "./node_modules/@514labs/moose-lib/dist/compilerPlugin.js",
-    transformProgram: true,
+    // No longer using transformProgram - direct typia integration eliminates
+    // the need for program replacement and the associated incremental compilation issues
   },
   {
+    // Keep typia plugin for users who use typia directly (not through Moose resources)
     transform: "typia/lib/transform",
```
onelesd, line 21: is this for backwards compatibility?

packages/ts-moose-lib/src/compilerPluginHelper.ts
```diff
@@ -43,98 +16,95 @@ export const isMooseFile = (sourceFile: ts.SourceFile): boolean => {
...
+    _configOrHost: PluginConfig | ts.CompilerHost | undefined,
+    _extrasOrConfig: TransformerExtras | PluginConfig,
+    maybeProgramExtras?: unknown,
+  ): ts.TransformerFactory<ts.SourceFile> => {
+    // Detect if called with transformProgram: true (4 args) vs regular transformer (3 args)
+    // transformProgram signature: (program, host, config, extras) => Program
+    // regular signature: (program, config, extras) => TransformerFactory
+    if (maybeProgramExtras !== undefined) {
+      throw new Error(
+        `[moose] Your tsconfig.json has "transformProgram": true for the moose plugin, ` +
```
onelesd, line 56: a nice DX improvement would be to publish our own tsconfig.json the user extends from.

this would make it more evident what moose expects vs what the user adds.

packages/ts-moose-lib/src/compilerPluginHelper.ts
```diff
@@ -43,98 +16,95 @@ export const isMooseFile = (sourceFile: ts.SourceFile): boolean => {
...
+        if (
+          sourceFile.isDeclarationFile ||
+          sourceFile.fileName.includes("/node_modules/")
+        ) {
+          return sourceFile;
+        }
+
+        // Only transform files in the current project
+        if (
+          sourceFile.fileName.startsWith("/") &&
```
onelesd, line 83: do we not support windows?

packages/ts-moose-lib/src/typiaDirectIntegration.ts
```diff
@@ -0,0 +1,406 @@
...
+  modulo: ts.LeftHandSideExpression;
+  sourceFile: ts.SourceFile;
+}
+
+/**
+ * Creates a synthetic identifier with a patched getText method.
+ * Typia's programmers call modulo.getText() which normally requires
+ * the node to be attached to a source file. We patch the method directly.
+ */
+const createSyntheticModulo = (): ts.LeftHandSideExpression => {
```
onelesd, line 65: nit; this patches getText() on a synthetic TS node. wdyt about something like:

```typescript
const createSyntheticModulo = (): ts.Identifier => {
  const id = ts.factory.createIdentifier("typia");
  (id as any).getText = () => "typia";
  return id;
};
```

casting to any at assignment would be a bit clearer than using an intersection type, since this is an escape hatch rather than a real extension of ts.Identifier

callicles (approved):

packages/ts-moose-lib/src/compilerPluginHelper.ts
phiSgr: no. this isn't the only place where we assume unix-like. i mean it's good to have, just don't want to increase the scope.

PR #3425, opened 2026-01-31
ENG-2074: update cliLog to use structured logging

LucioFranco:
> [!NOTE]
> **Medium Risk**
> Changes how runtime logs are transported and surfaced in the CLI (including prod/dev UI suppression), which could affect log visibility or formatting across Python/TypeScript runners.
> 
> **Overview**
> **CLI/runtime logging is reworked to use structured stderr logs instead of an HTTP management endpoint.** The `/logs` route and the `CliMessage` JSON payload are removed, and `cliLog` in both `py-moose-lib` and `ts-moose-lib` now writes `__moose_structured_log__` JSON (with `cli_action`/`cli_message_type`, plus new `Warning` support) directly to stderr.
> 
> **Child-process stderr handling is centralized and made prod-aware.** `apps/framework-cli` extends `parse_structured_log`/`process_stderr_lines` to recognize CLI-style structured logs and route them to terminal UI in dev while always emitting them to `tracing`; `is_prod` is threaded through streaming, consumption APIs, and workflow worker spawns to suppress terminal UI in production.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5e5edc91c5d2ba0ed8f0b7eac5ed9fa1e4aec38e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/framework/python/scripts_worker.rs
+++ b/apps/framework-cli/src/framework/python/scripts_worker.rs
@@ -18,7 +18,7 @@ pub enum WorkerProcessError {
     ProjectFileError(#[from] ProjectFileError),
 }
 
-pub fn start_worker(project: &Project) -> Result<Child, WorkerProcessError> {
+pub fn start_worker(project: &Project, is_prod: bool) -> Result<Child, WorkerProcessError> {
     // Create the wrapper lib files inside the .moose directory
     let internal_dir = project.internal_dir()?;
     let python_worker_lib_dir = internal_dir.join(PYTHON_WORKER_WRAPPER_PACKAGE_NAME);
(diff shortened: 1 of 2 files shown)


apps/framework-cli/src/cli/logger.rs
```diff
@@ -93,10 +93,25 @@ use tracing_subscriber::{EnvFilter, Layer};
...
 use crate::utilities::constants::{CONTEXT, CTX_SESSION_ID, NO_ANSI};
 use std::sync::atomic::Ordering;
+use std::sync::OnceLock;
 
 use super::otlp;
 use super::settings::user_directory;
 
+/// Global flag to track if the CLI is running in production mode.
+/// Used to conditionally display CLI logs in the terminal (dev mode only).
+static IS_PROD: OnceLock<bool> = OnceLock::new();
```
callicles: Can we not do that and just pass through the flag?

That makes it more testable than a global flag

LucioFranco: Yeah, I had gone this way since usually logging isn't in the testing sphere and this is only used for the logger. That said, I don't care that much and preventing possible future abuse is good.

LucioFranco: https://github.com/514-labs/moosestack/pull/3425/commits/8a877a3de09803a7adf58b6c724fe4eb531b7aea

PR #3572, opened 2026-02-14
ENG-2230: added github action for building and publishing docker images for moose templates

jmsuzuki:
to be used for quick template deployments in boreal, skipping the build at request time.

---

> [!NOTE]
> **Medium Risk**
> Introduces CI that builds and publishes Docker images to GCP, so misconfiguration could push incorrect images/tags or run unexpectedly on template changes, but it’s isolated to a new workflow file.
> 
> **Overview**
> Adds a new GitHub Actions workflow (`.github/workflows/build-template-images.yaml`) that automatically detects which templates changed (or allows manual selection) and then builds Docker images for each affected template using `moose init` + `moose build --docker`.
> 
> The workflow authenticates to GCP via OIDC, tags images with `moose` version + short commit SHA (and `latest` on `main`), and pushes to Artifact Registry; it also supports monorepo templates by building/tagging each discovered `moose.config.toml` project separately.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 77bf1de2f009576662c45234c885782144d98493. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/build-template-images.yaml
+++ b/.github/workflows/build-template-images.yaml
@@ -1,0 +1,182 @@
+name: Build Template Docker Images
+
+on:
+  push:
+    branches:
+      - "**"
+    paths:
+      - "templates/**"
+
+  workflow_dispatch:
+    inputs:
+      template:
+        type: string
+        description: "Specific template to rebuild (e.g. 'python-fastapi'), or 'all' to rebuild everything. Leave empty to auto-detect from latest commit."
+        required: false
+        default: ""
+      build_all:
+        type: boolean
+        description: "Build all templates"
+        required: false
+        default: false
+
+jobs:
+  detect-changes:
+    name: Detect Changed Templates
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      templates: ${{ steps.set-matrix.outputs.templates }}
+      has_templates: ${{ steps.set-matrix.outputs.has_templates }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Determine changed templates
+        id: set-matrix
+        run: |
+          # All templates that support Docker builds
+          ALL_TEMPLATES=(
+            "typescript"
+            "typescript-empty"
+            "typescript-fastify"
+            "typescript-express"
+            "typescript-mcp"
+            "typescript-migrate-test"
+            "ads-b"
+            "ads-b-frontend"
+            "python"
+            "python-tests"
+            "python-empty"
+            "python-fastapi"
+            "python-fastapi-client-only"
+            "python-experimental"
+            "python-cluster"
+          )
+
+          MANUAL_INPUT="${{ inputs.template }}"
+          BUILD_ALL="${{ inputs.build_all }}"
+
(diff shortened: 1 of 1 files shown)


jmsuzuki:
* **#3572** <a href="https://app.graphite.com/github/pr/514-labs/moosestack/3572?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/514-labs/moosestack/3572?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

.github/workflows/release-cli.yaml
```diff
@@ -667,6 +667,134 @@ jobs:
...
+          # - typescript-tests
+          - typescript-empty
+          - typescript-fastify
+          - typescript-express
+          # - typescript-cluster
+          - typescript-mcp
+          - typescript-migrate-test
+          # - next-app-empty
+          # - github-dev-trends
+          # - brainwaves
```
jmsuzuki: commented out templates are failing...

.github/workflows/release-cli.yaml
```diff
@@ -667,6 +667,135 @@ jobs:
           build-args: |
             FRAMEWORK_VERSION=${{ needs.version.outputs.version }}
 
+  build-template-docker-images:
```
callicles: I believe this should be a seperatte workflow and it should trigger a build only for the template that gets changed. We don't udpate templates systematically when we update the CLI and this will run a matrix of builds for every template for every CLI release.

callicles (changes requested):
I believe we need to change the structure to make it non blocking to CLI release and more efficient and accurate

.github/workflows/release-cli.yaml
jmsuzuki: that's good feedback.  
I'll move it.

.github/workflows/build-template-images.yaml
```diff
@@ -0,0 +1,206 @@
...
+
+      - name: Install Moose CLI
+        run: |
+          TEMPLATE="${{ matrix.template }}"
+
+          # Detect the CLI version the template expects
+          # Check package.json (TypeScript) then requirements.txt (Python)
+          CLI_VERSION=""
+          while IFS= read -r -d '' pkg; do
+            CLI_VERSION=$(python3 -c "import json; print(json.load(open('$pkg')).get('devDependencies',{}).get('@514labs/moose-cli',''))" 2>/dev/null)
```
callicles: We should be looking at moose-lib for the version since we want to enforce sync between moose lib and cli.   
  
instead of re-building the detection flow which is embedded in the install script you can just curl the install script and when you run the command inside a template directory it will setup the right version  
  
If you are thinking well then it downloads moose twice - true but then I would add the when installing moose from a directy that's a moose project detect the version that should be installed in the install.sh

callicles (changes requested):

PR #3620, opened 2026-02-20
preemptively fail if lib version mismatch

phiSgr:
> [!NOTE]
> **Medium Risk**
> Introduces a new preflight that can block all runner invocations if version detection fails or `moose-runner` isn’t available, potentially affecting local/dev environments.
> 
> **Overview**
> Adds a **one-time runtime check** before the CLI spawns `moose-runner`, verifying the installed `@514labs/moose-lib` version matches `CLI_VERSION` and returning a clear error (with a specific message for older libs that lack version support).
> 
> Updates `moose-runner` to report its package version dynamically and introduces a new `print-version` command used by the CLI for this validation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a7813b7acb312bd3539bd025f567f2fa4f4c2b0b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/framework/typescript/bin.rs
+++ b/apps/framework-cli/src/framework/typescript/bin.rs
@@ -37,11 +37,15 @@ fn check_moose_lib_version(project_path: &Path) -> Result<(), String> {
         .map_err(|e| format!("Failed to run moose-runner print-version: {e}"))?;
 
     if !output.status.success() {
-        return Err(format!(
-            "Version mismatch: installed @514labs/moose-lib does not support version checking \
-             and is older than the Moose CLI ({CLI_VERSION}). \
-             Please run `npm install @514labs/moose-lib@{CLI_VERSION}` to update."
-        ));
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        if stderr.contains("unknown command") {
+            return Err(format!(
+                "Version mismatch: installed @514labs/moose-lib does not support version checking \
+                 and is older than the Moose CLI ({CLI_VERSION}). \
+                 Please install @514labs/moose-lib@{CLI_VERSION}."
+            ));
+        }
+        return Err(format!("Failed to check moose-lib version: {stderr}"));
     }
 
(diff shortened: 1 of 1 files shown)


apps/framework-cli/src/framework/typescript/bin.rs
```diff
@@ -1,32 +1,81 @@
...
 
 use crate::project::Project;
 use tokio::process::{Child, Command};
 
 const RUNNER_COMMAND: &str = "moose-runner";
 
+static VERSION_CHECK_RESULT: OnceLock<Result<(), String>> = OnceLock::new();
+
+fn bin_path(project_path: &Path) -> String {
+    let path = env::var("PATH").unwrap_or_else(|_| "/usr/local/bin".to_string());
```
onelesd, line 14: why is the fallback path /usr/local/bin?   
we don't put moose there

apps/framework-cli/src/framework/typescript/bin.rs
```diff
@@ -1,32 +1,81 @@
...
+        .stdout(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::piped())
+        .output()
+        .map_err(|e| format!("Failed to run moose-runner print-version: {e}"))?;
+
+    if !output.status.success() {
+        return Err(format!(
+            "Version mismatch: installed @514labs/moose-lib does not support version checking \
+             and is older than the Moose CLI ({CLI_VERSION}). \
+             Please run `npm install @514labs/moose-lib@{CLI_VERSION}` to update."
```
onelesd: npm may be the wrong package manager. we should just say which version to install.

apps/framework-cli/src/framework/typescript/bin.rs
phiSgr: this is just existing code moved to a function. i think it's just any path so that we can prepend `{}/node_modules/.bin:` to

onelesd (approved):

PR #3634, opened 2026-02-24
use sqlglotc instead of sqlglotrs

phiSgr:
> [!NOTE]
> **Medium Risk**
> Changes CLI infra loading behavior to ignore non-fatal stderr when stdout is valid, which could mask some real errors if output parsing is overly permissive; dependency updates are otherwise straightforward.
> 
> **Overview**
> Updates Python dependencies across `py-moose-lib` and Python templates to use `sqlglot[c]>=29.0.1` instead of `sqlglot[rs]`.
> 
> Improves CLI infra-map loading (`PartialInfrastructureMap::from_subprocess`) to parse the JSON payload from stdout even when stderr contains output, logging stderr as warnings only when stdout is valid; stderr now only fails loading when stdout lacks the expected payload.
> 
> Adjusts the backward-compatibility E2E test to uninstall `sqlglotrs` after `pip install` to avoid deprecation warnings that cause older published CLIs to treat stderr as fatal.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 73af6d1e29ffba1fa6d050c788b71716958b5a81. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli-e2e/test/backward-compatibility.test.ts
+++ b/apps/framework-cli-e2e/test/backward-compatibility.test.ts
@@ -291,7 +291,27 @@ async function setupPythonProjectWithLatestPypi(
           );
           return;
         }
-        resolve();
+
+        // TODO(sqlglotc): Remove once published moose-lib no longer depends on sqlglot[rs].
+        // The published moose-lib transitively installs sqlglotrs, which emits a
+        // deprecation warning to stderr on import. The old (published) CLI treats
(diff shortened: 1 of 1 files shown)


callicles (approved):

phiSgr:
lol

```
Run arduino/setup-protoc@v3
Error: Error: socket hang up
```

PR #3662, opened 2026-02-27
fix -Merge suffix position

phiSgr:
no relevant bug fix in py-moose-lib because no relevant funcationality there

---

> [!NOTE]
> **Low Risk**
> Low-risk, localized change to SQL string generation for aggregated columns; main risk is altering generated ClickHouse function names for any existing `aggregationFunction` annotations that include parentheses.
> 
> **Overview**
> Fixes how `Sql` renders aggregated `Column`s by placing the `Merge` suffix **before** any existing parameter list in the annotated aggregation function name (e.g. `foo(…)` -> `fooMerge(…)`), instead of always appending `Merge` at the end.
> 
> This corrects generated ClickHouse SQL for *parametric* aggregate functions while preserving the previous behavior for non-parametric function names.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d6077c44aca9c29db05f897d5f8de33b11261753. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/ts-moose-lib/src/sqlHelpers.ts
+++ b/packages/ts-moose-lib/src/sqlHelpers.ts
@@ -159,8 +159,14 @@ export class Sql {
           ([k, _]) => k === "aggregationFunction",
         );
         if (aggregationFunction !== undefined) {
-          this.strings[pos] +=
-            `${(aggregationFunction[1] as AggregationFunction).functionName}Merge(\`${child.name}\`)`;
+          const funcName = (aggregationFunction[1] as AggregationFunction)
+            .functionName;
+          const parenIdx = funcName.indexOf("(");
+          const mergedName =
+            parenIdx !== -1 ?
+              `${funcName.slice(0, parenIdx)}Merge${funcName.slice(parenIdx)}`
+            : `${funcName}Merge`;
+          this.strings[pos] += `${mergedName}(\`${child.name}\`)`;
         } else {
           this.strings[pos] += `\`${child.name}\``;
         }


groy-514 (approved):
Yes, that fix addresses exactly the issue. Let me trace through why.
The problem was: moose-lib generated quantile(0.95)Merge(\col\) (appending "Merge" after the full function name), but ClickHouse requires quantileMerge(0.95)(\col\) ("Merge" inserted between the function base name and the parameters).
What the fix does: It checks if functionName contains a parenthesis. If it does (parametric function), it inserts "Merge" before the ( instead of after the full string:
funcName = "quantile(0.95)"parenIdx = 8  (index of "(")mergedName = "quantile" + "Merge" + "(0.95)" = "quantileMerge(0.95)"→ quantileMerge(0.95)(`p95_ms`)  ✓
For non-parametric functions it falls through to the existing behavior:
funcName = "count"parenIdx = -1  (no parenthesis)mergedName = "count" + "Merge" = "countMerge"→ countMerge(`query_count`)  ✓
What this means for our code: If this change lands in moose-lib, we'd type the columns with the parameter embedded in the function name:
p95_ms: number & Aggregated<"quantile(0.95)", [number]>;
instead of our current:
p95_ms: number & Aggregated<"quantile", [number]>;
And then the API could use moose's auto-finalization -- just reference ${cols.p95_ms} in the sql template and moose generates the correct quantileMerge(0.95)(\p95_ms\) automatically. No more explicit raw SQL for the -Merge calls.

PR #3693, opened 2026-03-05
prompt user when destructive operation is detected during dev

phiSgr:
> [!NOTE]
> **Medium Risk**
> Changes the dev hot-reload migration flow by adding interactive confirmation gates and a new column-rename operation, which affects schema execution and terminal I/O behavior. Risk is moderated by being primarily dev-mode/interactive, but incorrect gating or rename detection could block or apply unintended schema changes.
> 
> **Overview**
> Adds *migration safety gates* to `moose dev`: plans are now scanned for destructive operations (table/column/view drops and drop+recreate) and execution is blocked until the user confirms, with overrides via new flags/env vars (`--yes-all`, `--yes-destructive`, `--yes-rename`, `MOOSE_ACCEPT_*`).
> 
> Introduces heuristic *column rename detection* in table diffs, a `ColumnChange::Renamed` variant, and a corresponding ClickHouse `RenameTableColumn` operation; detected renames are prompted for confirmation and then rewritten from `Removed`+`Added` into `Renamed` to avoid treating renames as destructive.
> 
> Improves terminal robustness by adding a global `terminal_lock`, making spinners pause/resume via `SpinnerHandle`, and adding a non-blocking `prompt_user_async` so interactive prompts don’t corrupt output while watchers/spinners are running.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0ba368b7c87e2467da88eac6f898b69c95ac7478. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/framework-cli/src/cli.rs
+++ b/apps/framework-cli/src/cli.rs
@@ -1522,7 +1522,11 @@ pub async fn top_command_handler(
 
             let template_cmd = template_args.command.as_ref().unwrap();
             match template_cmd {
-                TemplateSubCommands::List {} => {
+                TemplateSubCommands::List { json } => {
+                    if *json {
+                        QUIET_STDOUT.store(true, Ordering::Relaxed);
+                    }
+
                     let capture_handle = crate::utilities::capture::capture_usage(
                         ActivityType::TemplateListCommand,
                         None,
@@ -1531,7 +1535,7 @@ pub async fn top_command_handler(
                         HashMap::new(),
                     );
 
-                    let result = list_available_templates(CLI_VERSION).await;
+                    let result = list_available_templates(CLI_VERSION, *json).await;
 
                     wait_for_usage_capture(capture_handle).await;
 
@@ -1758,6 +1762,7 @@ pub async fn top_command_handler(
             let component_name = match &component {
                 AddComponent::McpServer(_) => "mcp-server",
                 AddComponent::Chat(_) => "chat",
+                AddComponent::Benchmark(_) => "benchmark",
             };
 
             let capture_handle = crate::utilities::capture::capture_usage(
--- a/apps/framework-cli/src/cli/commands.rs
+++ b/apps/framework-cli/src/cli/commands.rs
@@ -296,7 +296,7 @@ pub enum Commands {
         visible_alias = "a",
         about = "[EXPERIMENTAL] Add a component to your project",
         long_about = "Add a component to your project\n\n[EXPERIMENTAL] Component APIs and available components may change in future releases.",
-        after_help = "Examples:\n  moose add mcp-server --dir packages/moosestack-service\n  moose add chat --dir packages/web-app"
+        after_help = "Examples:\n  moose add mcp-server --dir packages/moosestack-service\n  moose add chat --dir packages/web-app\n  moose add benchmark --dir moose"
     )]
     Add {
         #[command(subcommand)]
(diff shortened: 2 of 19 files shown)


ghost:
No deployment triggered for this PR.

No relevant changes detected for:
- **moosestack** (`/examples/fastify-moose/moose`)

Projects rebuild when files change in their root path, `packages/`, or root-level lock files.

ghost:
> **Environment:** `preview branch: (cross-repo-vercel)`

No deployment triggered for this PR.

No relevant changes detected for:
- **moosestack** (`/examples/fastify-moose/moose`)

Projects rebuild when files change in their root path, `packages/`, or root-level lock files.

apps/framework-cli/src/cli/watcher.rs
```diff
@@ -260,30 +264,52 @@ async fn watch(
...
+                                framework::core::plan::plan_changes(&**state_storage, &project_inner).await
                             })
                             .await;
 
                             match plan_result {
                                 Ok((_, plan_result)) => {
                                     with_timing_async("Validation", async {
-                                        framework::core::plan_validator::validate(&project, &plan_result)
+                                        framework::core::plan_validator::validate(&project_inner, &plan_result)
                                     })
```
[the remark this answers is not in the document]
phiSgr: not introduced in this branch. i don't want to modify it now.

apps/framework-cli/src/framework/core/plan_risk.rs
```diff
@@ -0,0 +1,663 @@
...
+/// Walks the OLAP changes and collects every operation that may cause data loss.
+///
+/// A `TableChange::Removed` followed by a `TableChange::Added` with the same
+/// name is treated as a recreate rather than two independent operations.
+/// `ColumnChange::Renamed` is non-destructive and is intentionally skipped.
+pub fn classify_plan_risk(changes: &InfraChanges) -> PlanRisk {
+    let mut destructive_changes = Vec::new();
+
+    // Collect (database, name) pairs for tables that are both removed and added (recreates).
+    let removed_table_keys: std::collections::HashSet<(Option<&str>, &str)> = changes
```
LucioFranco: The llm should def import HashSet its making this line so long lol

apps/framework-cli/src/cli/display/terminal_lock.rs
```diff
@@ -0,0 +1,51 @@
...
+//! a pipe, so contention is negligible.
+//!
+//! When a pinned prompt is active, [`scroll_region_bottom`] returns the
+//! row that log output should target so it scrolls inside the region
+//! instead of overwriting the prompt area.
+
+use std::sync::atomic::{AtomicU16, Ordering};
+use std::sync::{Mutex, MutexGuard};
+
+static LOCK: Mutex<()> = Mutex::new(());
```
LucioFranco, line 24: we need to figure out how to not enable this lock for production and actually we should be shoving all logs in prod into tracing anyways. This sort of thing is fine for the dev server

apps/framework-cli/src/framework/core/infrastructure_map.rs
```diff
@@ -3739,6 +3744,534 @@ pub fn compute_table_columns_diff(
...
+///
+/// When a user renames a column, the name-based diff produces a `Removed` +
+/// `Added` pair. This function inspects those pairs and scores them using
+/// several heuristics (type equality, positional proximity, property
+/// similarity, name similarity, and set cardinality) to decide whether a
+/// remove/add pair is actually a rename.
+///
+/// Only pairs that exceed [`RENAME_CONFIDENCE_THRESHOLD`] are returned, and
+/// each removed/added column appears in at most one rename (greedy best-match).
+pub fn detect_column_renames(
```
LucioFranco, line 3921: I dont feel like this is the right approach for detecting renames due to the false positives here. I wonder if all we need is to just check that a col that was used in a prev order_by wasn't changed? I am worried here about the false positives that this would create. I am curious to hear what @callicles thinks about this.

apps/framework-cli/src/framework/core/plan_risk.rs
phiSgr: agreed. i cleaned up some of the worst ones already.

apps/framework-cli/src/framework/core/infrastructure_map.rs
callicles: Overall stance is: if there is a risk of bad inference, don't assume anything - ask the user. 
For the detection not sure if we can have an algorithm that's reliable - probably something to hash out as a team

LucioFranco: @phiSgr I think maybe we should remove this for now and just detect the stuff we can with 100% certainty like changing a col in an order_by that would cause a drop table?

PR #3917, opened 2026-04-02
fix: return 404 instead of 500 for unknown API routes

03cranec:
This is a claude vibe coded attempt at fixing ENG-2653 https://linear.app/514/issue/ENG-2653/apiname-returns-500-on-unknown-route

https://claude.ai/code/session_014UXE2JgQnj7Nm9M7x932To

sending to eng in case it is good / useful / can just be approved. it not, feel free to take over the branch/PR, or just implement the ticket yourself from scratch.

### DIFF
--- a/packages/ts-moose-lib/src/consumption-apis/runner.ts
+++ b/packages/ts-moose-lib/src/consumption-apis/runner.ts
@@ -69,9 +69,10 @@ const send404 = (
   req: http.IncomingMessage,
   res: http.ServerResponse,
   startMs: number,
+  message: string = "Not Found",
 ) => {
   res.writeHead(404, { "Content-Type": "application/json" });
-  res.end(JSON.stringify({ error: "Not Found" }));
+  res.end(JSON.stringify({ error: "Not Found", message }));
   httpLogger(req, res, startMs);
 };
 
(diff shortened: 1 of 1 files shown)


packages/ts-moose-lib/src/consumption-apis/runner.ts
```diff
@@ -374,6 +368,10 @@ const apiHandler = async (
         res.writeHead(400, { "Content-Type": "application/json" });
         res.end(JSON.stringify(error.toJSON?.() ?? { error: error.message }));
         httpLogger(req, res, start, matchedApiName);
+      } else if (error?.name === "NotFoundError") {
```
onelesd:
```suggestion
      } else if (error instanceof NotFoundError) {
```

onelesd (approved):

packages/ts-moose-lib/src/consumption-apis/runner.ts
phiSgr: this is actually tricky stuff.

we might load two versions of the same class and `instanceof` will fail to check it.
this is what `same workaround as ResultSet` comment above meant.

PR #4024, opened 2026-04-17
fix: startup uneasyness

callicles:
> [!NOTE]
> **Medium Risk**
> Touches dev infrastructure lifecycle and proxying behavior (port preflight, retries, process restart logic), which could affect local startup reliability and error surfacing, but is scoped to dev/docklerless flows.
> 
> **Overview**
> Improves **dockerless/dev startup stability** by adding a preflight port conflict check and better shutdown semantics, reducing EADDRINUSE restart storms and flaky reruns.
> 
> The CLI consumption proxy now **retries transient connect failures** and returns a structured `503` with `Retry-After` when the consumption API is briefly unavailable. Consumption API startup also waits for `proxy_port` to become bindable, the Node runner binds explicitly to `127.0.0.1` and exits cleanly on `EADDRINUSE`, and the Rust `RestartingProcess` adds a circuit breaker to stop infinite rapid respawns.
> 
> ClickHouse native infra handling is tightened by persisting the correct watchdog `comm` in PID files and waiting/escalating on shutdown so ports/files are actually released; E2E suites now pass their offset port sets to cleanup helpers to reliably kill leftover processes. Adds a `Cross.toml` convenience config for aarch64 cross-builds and updates `pnpm-lock.yaml` platform `libc` metadata.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9427e836490b32f88ad39d9d2b3d351d79d53224. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cross.toml
+++ b/Cross.toml
@@ -1,0 +1,32 @@
+# Local cross-compile config for dec-bench harness tests.
+# CI uses manylinux_2_28 directly; this is the convenience path for
+# `cross build --target aarch64-unknown-linux-gnu -p moose-cli`.
+[build.env]
+# Tell openssl-sys to link the system OpenSSL 1.1 inside the container
+# instead of vendoring OpenSSL 3.5.5. The vendored build strips deprecated
+# symbols (SSL_get_peer_certificate), which librdkafka still calls — so the
+# link fails. Using the system lib keeps both old and new names available.
+passthrough = ["OPENSSL_LIB_DIR", "OPENSSL_INCLUDE_DIR", "OPENSSL_STATIC"]
+
+[target.aarch64-unknown-linux-gnu]
+# The pinned 0.2.5 image ships libssl 1.0.0, which lacks SSL_get_peer_certificate
+# as an exported symbol and fails librdkafka's final link. `:main` is Ubuntu
+# 20.04 with libssl 1.1.1, which has the symbol.
+image = "ghcr.io/cross-rs/aarch64-unknown-linux-gnu:main"
+# cross's /opt/toolchain.cmake pins CMAKE_FIND_ROOT_PATH to /usr/aarch64-linux-gnu
+# (sysroot layout), but apt multiarch installs arm64 libs to
+# /usr/lib/aarch64-linux-gnu (Debian layout). rdkafka-sys's cmake build of
+# librdkafka does FIND_PACKAGE(ZLIB) / FIND_PACKAGE(OpenSSL) and can't see
+# across that gap — so we symlink the multiarch lib/headers into the sysroot.
+pre-build = [
+  "dpkg --add-architecture arm64",
+  "apt-get update",
+  "apt-get install -y --no-install-recommends protobuf-compiler libprotobuf-dev cmake pkg-config zlib1g-dev:arm64 libssl-dev:arm64",
+  "mkdir -p /usr/aarch64-linux-gnu/lib /usr/aarch64-linux-gnu/include",
+  "ln -sf /usr/lib/aarch64-linux-gnu/libz.so /usr/aarch64-linux-gnu/lib/libz.so",
+  "ln -sf /usr/lib/aarch64-linux-gnu/libz.a /usr/aarch64-linux-gnu/lib/libz.a",
+  "ln -sf /usr/include/zlib.h /usr/aarch64-linux-gnu/include/zlib.h",
+  "ln -sf /usr/include/zconf.h /usr/aarch64-linux-gnu/include/zconf.h",
+  "for f in libssl.so libssl.a libcrypto.so libcrypto.a; do ln -sf /usr/lib/aarch64-linux-gnu/$f /usr/aarch64-linux-gnu/lib/$f; done",
+  "ln -sfn /usr/include/openssl /usr/aarch64-linux-gnu/include/openssl",
+]
--- a/Cross.toml
+++ b/Cross.toml
@@ -7,11 +7,23 @@
 # system OpenSSL 1.1 inside the container instead. OPENSSL_NO_VENDOR is the
 # documented opt-out for the `vendored` feature — it keeps moose-cli's
 # Cargo.toml unchanged so regular native builds keep vendoring.
+#
+# CC / CXX / AR are passthrough because rdkafka-sys runs librdkafka's
+# autoconf `./configure` as a subprocess, which reads bare `$CC` from env
+# rather than cargo's target-qualified `CC_aarch64_unknown_linux_gnu`.
+# Without them configure defaults to the host `gcc` (x86_64) and ld fails
+# with `cannot find -lcrypto` because only arm64 libs are in the sysroot.
+# Invoke like:
+#   CC=aarch64-linux-gnu-gcc CXX=aarch64-linux-gnu-g++ AR=aarch64-linux-gnu-ar \
+#     OPENSSL_NO_VENDOR=1 cross build --target aarch64-unknown-linux-gnu -p moose-cli
 passthrough = [
   "OPENSSL_NO_VENDOR",
   "OPENSSL_LIB_DIR",
   "OPENSSL_INCLUDE_DIR",
   "OPENSSL_STATIC",
+  "CC",
+  "CXX",
+  "AR",
 ]
 
 [target.aarch64-unknown-linux-gnu]
--- a/Cross.toml
+++ b/Cross.toml
@@ -2,11 +2,17 @@
 # CI uses manylinux_2_28 directly; this is the convenience path for
 # `cross build --target aarch64-unknown-linux-gnu -p moose-cli`.
 [build.env]
-# Tell openssl-sys to link the system OpenSSL 1.1 inside the container
-# instead of vendoring OpenSSL 3.5.5. The vendored build strips deprecated
-# symbols (SSL_get_peer_certificate), which librdkafka still calls — so the
-# link fails. Using the system lib keeps both old and new names available.
-passthrough = ["OPENSSL_LIB_DIR", "OPENSSL_INCLUDE_DIR", "OPENSSL_STATIC"]
+# Tell openssl-sys to skip the vendored build (it strips deprecated symbols
+# like SSL_get_peer_certificate that librdkafka still calls) and link the
+# system OpenSSL 1.1 inside the container instead. OPENSSL_NO_VENDOR is the
+# documented opt-out for the `vendored` feature — it keeps moose-cli's
+# Cargo.toml unchanged so regular native builds keep vendoring.
+passthrough = [
+  "OPENSSL_NO_VENDOR",
+  "OPENSSL_LIB_DIR",
+  "OPENSSL_INCLUDE_DIR",
+  "OPENSSL_STATIC",
+]
 
 [target.aarch64-unknown-linux-gnu]
 # The pinned 0.2.5 image ships libssl 1.0.0, which lacks SSL_get_peer_certificate
--- a/apps/framework-cli/src/cli/local_webserver.rs
+++ b/apps/framework-cli/src/cli/local_webserver.rs
@@ -2698,7 +2698,11 @@ impl Webserver {
             .await
             .unwrap_or_else(|e| handle_listener_err(management_socket.port(), e));
 
-        // Check if proxy port is available
+        // Defense in depth: quick bind-and-drop on proxy_port so Docker-mode
+        // users get a clear error before the Node worker spawns. The
+        // --dockerless path also runs a structured preflight in
+        // NativeInfraProvider::start that covers this port alongside the
+        // embedded-infra ports.
         let proxy_socket = self.get_socket(project.http_server_config.proxy_port).await;
         TcpListener::bind(proxy_socket)
             .await
--- a/apps/framework-cli/src/cli/local_webserver.rs
+++ b/apps/framework-cli/src/cli/local_webserver.rs
@@ -583,9 +583,11 @@ async fn get_consumption_api_res(
         .collect();
     let url_parsed: reqwest::Url = url.parse()?;
 
-    // Retry only on genuine connect errors — these are the hot-reload window
-    // (consumption-api primary restarting → :proxy_port momentarily closed).
-    // Other failures (timeouts, TLS errors, 5xx from upstream) surface once.
+    // Retry only on genuine connect errors — a transient "socket not
+    // accepting" window where the consumption-api is momentarily
+    // unavailable. In dev this is most often the hot-reload restart of the
+    // primary worker; in prod it can happen on any brief restart / socket
+    // reset. Other failures (timeouts, TLS, 5xx from upstream) surface once.
     const MAX_ATTEMPTS: usize = 3;
     const BACKOFF_MS: [u64; 2] = [150, 400]; // applied between attempts 1→2 and 2→3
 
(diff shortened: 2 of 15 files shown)


apps/framework-cli/src/utilities/native_infra/preflight.rs
```diff
@@ -0,0 +1,365 @@
...
+                };
+                let owner_pid = owner_name
+                    .and_then(|name| read_live_pid(&native_dir.join(format!("{name}.pid")), name));
+                conflicts.push(PortConflict {
+                    spec: *spec,
+                    owner_pid,
+                    owner_name: owner_name.filter(|_| owner_pid.is_some()),
+                });
+            }
+        }
```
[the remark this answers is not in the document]
callicles: Skipping — the service→owner match lives one scroll away from the producer in `port_specs_for`, both in the same file. For what is explicitly a best-effort error enrichment, adding a `pid_owner` field to `PortSpec` (or promoting service to an enum) is more API surface than the coupling justifies. Happy to revisit if this file grows.

Cross.toml
```diff
@@ -0,0 +1,32 @@
+# Local cross-compile config for dec-bench harness tests.
+# CI uses manylinux_2_28 directly; this is the convenience path for
+# `cross build --target aarch64-unknown-linux-gnu -p moose-cli`.
+[build.env]
+# Tell openssl-sys to link the system OpenSSL 1.1 inside the container
+# instead of vendoring OpenSSL 3.5.5. The vendored build strips deprecated
+# symbols (SSL_get_peer_certificate), which librdkafka still calls — so the
+# link fails. Using the system lib keeps both old and new names available.
+passthrough = ["OPENSSL_LIB_DIR", "OPENSSL_INCLUDE_DIR", "OPENSSL_STATIC"]
```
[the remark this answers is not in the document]
callicles: Skipping — GitHub flagged this comment as outdated. Functionally the `pre-build` steps below install and symlink the arm64 OpenSSL into the sysroot, so a stray host-set `OPENSSL_LIB_DIR`/`OPENSSL_INCLUDE_DIR` wouldn't be load-bearing. `OPENSSL_NO_VENDOR` is the only variable the build actually depends on.

Cross.toml
```diff
@@ -0,0 +1,32 @@
...
+
+[target.aarch64-unknown-linux-gnu]
+# The pinned 0.2.5 image ships libssl 1.0.0, which lacks SSL_get_peer_certificate
+# as an exported symbol and fails librdkafka's final link. `:main` is Ubuntu
+# 20.04 with libssl 1.1.1, which has the symbol.
+image = "ghcr.io/cross-rs/aarch64-unknown-linux-gnu:main"
```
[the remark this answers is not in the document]
callicles: Skipping — the `:main` tag is deliberate (documented at L18–20 of this file): the pinned `0.2.5` image ships libssl 1.0.0 which is missing `SSL_get_peer_certificate` and breaks the rdkafka link. This `Cross.toml` is a local dev convenience path, not CI, so the reproducibility argument for digest-pinning doesn't buy much here.

packages/ts-moose-lib/src/consumption-apis/runner.ts
```diff
@@ -611,7 +611,35 @@ export const runApis = async (config: ApisConfig) => {
...
+          console.error(
+            `[consumption-api] unexpected server error on port ${port}:`,
+            err,
+          );
+        }
+        // Surrender gracefully. The Cluster's workerStop isn't appropriate
+        // here (we never reached a healthy state), so exit non-zero and let
+        // the primary schedule a retry.
+        process.exit(1);
+      });
```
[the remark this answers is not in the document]
callicles: Intentional — the PR's goal for ENG-2768 is to make the worker exit non-zero on any server error so the Rust `RestartingProcess` circuit breaker can count rapid failures. A worker that only handles HTTP requests can't recover from a post-listen socket error, so hard-exit and let the primary respawn. Scoping the handler to the listen phase would silently swallow those failures and reintroduce the spam-the-log behavior we just fixed.

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -571,26 +571,111 @@ async fn get_consumption_api_res(
...
+        match http_client.execute(client_req).await {
+            Ok(res) => {
+                let status = res.status();
+                let body = res.bytes().await?;
+                return Ok(add_cors_headers(Response::builder())
+                    .status(status)
+                    .header("Content-Type", "application/json")
+                    .body(Full::new(body))
+                    .unwrap());
+            }
```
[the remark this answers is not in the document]
callicles: Skipping — the retry path is intentionally scoped to *connect* errors (see `is_connect_error` at L658), which targets the hot-reload window where the worker isn't listening yet. If `execute()` succeeds, the TCP connection was made and the upstream committed to a response status; re-issuing after a body-read failure would be unsafe (the retried response could differ). Body-read failures after headers arrive are a different failure class that shouldn't be auto-retried.

apps/framework-cli/src/utilities/native_infra/mod.rs
```diff
@@ -489,12 +511,43 @@ pub fn kill_pid_file(pid_path: &Path) {
...
         }
+        sleep(POLL_INTERVAL);
     }
 
+    warn!("PID {pid} did not exit within {WAIT_TIMEOUT:?}; sending SIGKILL");
+    let _ = std::process::Command::new("kill")
+        .args(["-KILL", &pid.to_string()])
+        .output();
+    // Brief grace for the kernel to release ports held by the killed process.
+    sleep(Duration::from_millis(200));
```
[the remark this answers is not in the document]
callicles: Skipping — zombie state isn't reachable in this architecture. ClickHouse and Temporal are spawned with `kill_on_drop(false)` and we don't retain the `Child` handles across CLI invocations. When a subsequent `moose dev`/`moose clean` kills by PID, the current CLI process isn't their parent, so the kernel has already reparented them to PID 1 (launchd/init), which reaps zombies immediately. `kill -0` is correct and portable here; `ps -o stat=` would add parsing overhead for a scenario that can't occur.

apps/framework-cli/src/utilities/native_infra/preflight.rs
```diff
@@ -0,0 +1,365 @@
...
+        } else {
+            writeln!(
+                f,
+                "If this is another moose project's dev server, run `moose clean` in \
+                 that project. Otherwise, stop the process holding these ports."
+            )?;
+        }
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
callicles: Follow-up: reverted this change after testing. Switching to .all() made same-project conflicts always route to the cleanup message, because devredis/devkafka/http/management/proxy_port don't write PID files — so only clickhouse/temporal are ever attributed, and .all() was effectively never true in practice. The resulting message was self-contradictory: it listed attributed PIDs from a prior moose dev and then told the user it was probably another moose project.

Kept .any() with the trade-off that the rare case of [our clickhouse + unrelated process on some port] still routes to "keep editing" — but every conflicting port is still listed above, so the user can intervene on a straggler if they see one. Added a code comment explaining the choice.

Cross.toml
```diff
@@ -0,0 +1,32 @@
...
+  "dpkg --add-architecture arm64",
+  "apt-get update",
+  "apt-get install -y --no-install-recommends protobuf-compiler libprotobuf-dev cmake pkg-config zlib1g-dev:arm64 libssl-dev:arm64",
+  "mkdir -p /usr/aarch64-linux-gnu/lib /usr/aarch64-linux-gnu/include",
+  "ln -sf /usr/lib/aarch64-linux-gnu/libz.so /usr/aarch64-linux-gnu/lib/libz.so",
+  "ln -sf /usr/lib/aarch64-linux-gnu/libz.a /usr/aarch64-linux-gnu/lib/libz.a",
+  "ln -sf /usr/include/zlib.h /usr/aarch64-linux-gnu/include/zlib.h",
+  "ln -sf /usr/include/zconf.h /usr/aarch64-linux-gnu/include/zconf.h",
+  "for f in libssl.so libssl.a libcrypto.so libcrypto.a; do ln -sf /usr/lib/aarch64-linux-gnu/$f /usr/aarch64-linux-gnu/lib/$f; done",
+  "ln -sfn /usr/include/openssl /usr/aarch64-linux-gnu/include/openssl",
```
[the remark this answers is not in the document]
callicles: Follow-up after end-to-end testing: I validated `cross build --target aarch64-unknown-linux-gnu -p moose-cli` to produce an ARM aarch64 ELF. The actual blocker was earlier than opensslconf.h — librdkafka's autoconf `./configure` was running the host `gcc` (x86_64) and failing at `cannot find -lcrypto` before any aarch64 compile ever reached the sysroot. Fixed by passing `CC=aarch64-linux-gnu-gcc CXX=aarch64-linux-gnu-g++ AR=aarch64-linux-gnu-ar` on the host invocation and adding those to the Cross.toml passthrough list.

With CC set correctly, I also re-tested whether the multiarch opensslconf.h layering was needed. It was **not** — the plain `ln -sfn /usr/include/openssl ...` symlink is sufficient because the header set rdkafka pulls in from `/usr/include/openssl/` (ssl.h, evp.h, err.h, etc.) is architecture-independent, and nothing in its compile-test path transitively includes `<openssl/opensslconf.h>`. Your multiarch analysis is correct per Debian spec; it just doesn't trigger for this particular rdkafka+openssl usage. Dropped the `cp -a` + layered symlink to keep the diff minimal per CLAUDE.md's 'only the code the task needs' guideline.

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -571,26 +571,111 @@ async fn get_consumption_api_res(
         let _ = execute!(std::io::stdout(), Print(msg + "\n"));
     }
 
-    let mut client_req = reqwest::Request::new(req.method().clone(), url.parse()?);
+    // Capture method + headers up front so we can rebuild the reqwest::Request
```
LucioFranco, line 574: If this works this is fine, but I think we need to switch to actual proxy code here use tower + hyper instead of reqwest. We can make it so we can just use clone then. If you feel like this code is tested them I am okay merging this now though this has smell to me. Thinking its like maybe ~1-2 days max of effort to do the conversion.

callicles: Agreed and I think that's outside of the scope here

LucioFranco: Agreed 100%

apps/framework-cli/src/cli/local_webserver.rs
```diff
@@ -571,26 +571,111 @@ async fn get_consumption_api_res(
...
-        .status(status)
+    // All retries exhausted with connect errors — return a structured 503 so
+    // agents can reason about this as a transient hot-reload window instead
+    // of an ambiguous 500. Retry-After tells well-behaved clients to try
+    // again shortly.
+    let body = serde_json::json!({
+        "error": "consumption_api_unavailable",
+        "retryable": true,
+        "message": format!(
+            "Consumption API is temporarily unavailable (restarting on hot reload). \
```
LucioFranco: does this only run in dev?

callicles: Hum, actually, no, but a transient socket error in prod is also retriable. 

Do you think I should make this less specific to hot reload on the error message?

apps/framework-cli/src/infrastructure/processes/consumption_registry.rs
```diff
@@ -132,3 +149,32 @@ impl ConsumptionProcessRegistry {
...
+    const INTERVAL: Duration = Duration::from_millis(100);
+    const TIMEOUT: Duration = Duration::from_secs(5);
+
+    let deadline = Instant::now() + TIMEOUT;
+    loop {
+        // A successful bind here proves the port is free. The TcpListener
+        // is dropped at the end of this scope, releasing the fd. Since we
+        // never accept a connection on it, no TIME_WAIT is created, so the
+        // Node worker can bind the same port a few ms later.
+        match TcpListener::bind((host, port)) {
```
LucioFranco, line 168: this doesn't feel ideal but for now is fine. Thinking maybe we can just use like the `nix` library to get this info in a better way but can do that later.

apps/framework-cli/src/utilities/native_infra/mod.rs
```diff
@@ -126,6 +127,21 @@ impl InfraProvider for NativeInfraProvider {
...
 
     fn start(&self, project: &Project) -> Result<(), RoutineFailure> {
+        // Preflight: surface EADDRINUSE in a single actionable message before
+        // anything starts. Prevents the Node consumption worker from entering
+        // an unbounded restart loop when a prior `moose dev --dockerless` is
+        // still holding ports 4001 / 6379 / 19092.
+        let specs = preflight::port_specs_for(
+            project,
+            self.scripts_enabled,
+            /* include_webserver = */ true,
```
LucioFranco: lol what is this comment would just remove it...

apps/framework-cli/src/utilities/native_infra/mod.rs
```diff
@@ -489,12 +513,43 @@ pub fn kill_pid_file(pid_path: &Path) {
...
+    }
+
+    // ClickHouse takes several seconds to flush and release its listen sockets;
+    // without waiting here the next `moose dev --dockerless` preflight sees the
+    // ports still bound and aborts with a false "another moose dev is running"
+    // error. Poll `kill -0` until the PID is gone, then escalate to SIGKILL.
+    const WAIT_TIMEOUT: Duration = Duration::from_secs(10);
+    const POLL_INTERVAL: Duration = Duration::from_millis(100);
+    let deadline = Instant::now() + WAIT_TIMEOUT;
+    while Instant::now() < deadline {
```
LucioFranco, line 529: no indication to user that we are waiting here?

LucioFranco (approved):

apps/framework-cli/src/cli/local_webserver.rs
LucioFranco: Yeah the error message (and maybe) the comments make it seem like its only for dev so would be good to tighten that up

apps/framework-cli/src/utilities/native_infra/mod.rs
callicles: Hum in moose dev normal during boot, there is a loading that shows we are checking if ClickHouse is up. do we have the same? I did not add anything here for this purpose

apps/framework-cli/src/cli/local_webserver.rs
callicles: Tightened the comment and the user-facing message to not frame the retry as dev-only. The retry still only fires on connect-class errors (transient socket unavailability), but now explicitly acknowledges it applies to any brief restart/reset — not just hot reload.

apps/framework-cli/src/utilities/native_infra/mod.rs
callicles: Removed the /* include_webserver = */ inline marker — agreed it's clutter.

apps/framework-cli/src/utilities/native_infra/mod.rs
callicles: Added a user-visible 'Stopping native infrastructure (up to ~10s per service)...' message in local_webserver.rs around stop_native_infra, plus a 'Stopped' follow-up once it returns. The per-PID poll loop still logs at info!; the outer wrapper makes it obvious the process isn't hung.

PR #4082, opened 2026-05-13
fix(ddl): skip dependent index/projection rebuild for comment-only column changes

hh2110:
ModifyTableColumn previously dropped and re-added every dependent skip-index and projection on any ColumnChange::Updated, regardless of which field diverged. ClickHouse only requires that drop when the modification changes on-disk layout or evaluated values (data_type, default, codec, materialized); a pure comment change is accepted in place. The over-conservative drop forced ClickHouse to rebuild skip-indexes lazily on merges, degrading data-skipping on existing parts -- hours of slower time-windowed reads for a metadata-only migration.

Gate drop_column_dependents / readd_column_dependents on a new column_modify_requires_dependent_drop helper that compares before/after with the comment field normalised. Any other field difference (including annotations) keeps the existing PR #3813 drop path, so new Column fields default to the safe behaviour.

Adds regression tests for comment-only changes against both minmax indexes and projections; existing PR #3813 type-change tests still cover the drop path.

### DIFF
--- a/apps/framework-cli/src/infrastructure/olap/ddl_ordering.rs
+++ b/apps/framework-cli/src/infrastructure/olap/ddl_ordering.rs
@@ -1348,21 +1348,58 @@ struct HandledDependencies {
     readded_projections: HashSet<String>,
 }
 
-/// Returns true if the diff between `before` and `after` requires dependent
-/// indexes/projections to be dropped before `ALTER TABLE ... MODIFY COLUMN`.
+/// Returns `true` if `before` and `after` differ in any field other than `comment`.
 ///
-/// ClickHouse rejects `MODIFY COLUMN` on an indexed column only when the
-/// modification changes on-disk layout or evaluated values (e.g. `data_type`,
-/// `default`, `codec`, `materialized`). A pure `comment` change is accepted
-/// without dropping the dependent, and re-creating the index would force
-/// ClickHouse to rebuild it lazily on merges, degrading data-skipping until then.
+/// `ALTER TABLE ... MODIFY COLUMN` is rejected by ClickHouse on an indexed
+/// column whenever the modification changes on-disk layout or evaluated values
+/// (`data_type`, `default`, `codec`, `materialized`, etc.). A pure `comment`
+/// change is the one case ClickHouse accepts without dropping dependent
+/// indexes/projections — and re-creating those would force a lazy index
+/// rebuild on merges, degrading data-skipping on existing parts until then.
 ///
-/// Conservatively, any field other than `comment` differing forces the drop,
-/// so new `Column` fields default to the safe behaviour.
+/// Implemented via exhaustive destructuring so that adding a new `Column`
+/// field is a compile error: the author must explicitly decide whether the
+/// field belongs in the "requires drop" set.
 fn column_modify_requires_dependent_drop(before: &Column, after: &Column) -> bool {
-    let mut normalised = before.clone();
-    normalised.comment = after.comment.clone();
-    normalised != *after
+    let Column {
+        name: _,
+        comment: _, // metadata-only; ClickHouse accepts the MODIFY without dropping dependents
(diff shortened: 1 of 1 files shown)


apps/framework-cli/src/infrastructure/olap/ddl_ordering.rs
```diff
@@ -1348,6 +1348,23 @@ struct HandledDependencies {
...
+/// modification changes on-disk layout or evaluated values (e.g. `data_type`,
+/// `default`, `codec`, `materialized`). A pure `comment` change is accepted
+/// without dropping the dependent, and re-creating the index would force
+/// ClickHouse to rebuild it lazily on merges, degrading data-skipping until then.
+///
+/// Conservatively, any field other than `comment` differing forces the drop,
+/// so new `Column` fields default to the safe behaviour.
+fn column_modify_requires_dependent_drop(before: &Column, after: &Column) -> bool {
+    let mut normalised = before.clone();
+    normalised.comment = after.comment.clone();
```
callicles: Why is the double clone needed?

callicles: Also the comment doesn't seem to reflect what it is doing. Is the comment the only case? or is is when drop is NOT required

hh2110: It wasn't — that was a sloppy way to mask the `comment` field for a single `!=`. Switched to exhaustive destructuring: zero clones, and a new `Column` field now fails compilation rather than silently joining the no-drop set.

hh2110: Fair — rewrote it. Now leads with the rule (`true` if `before` and `after` differ in any field other than `comment`); ClickHouse rationale moved below. `comment` is the only no-drop case; that's enforced at compile time by the destructure.

callicles (approved):

apps/framework-cli-e2e/test/backward-compatibility.test.ts
hh2110, line 1: @callicles  -- I had to make the following changes for CI to pass
```
The TypeScript backward-compat test set both @514labs/moose-cli and
@514labs/moose-lib to "latest" in the generated project's package.json,
then asserted via verifyTypeScriptVersionsMatch that the resolved
versions matched. That works while npm's two `latest` dist-tags stay in
sync, but breaks permanently when one publish lags or fails -- as
happened around 0.6.532, where moose-lib@latest advanced to 0.6.532
while moose-cli@latest remained at 0.6.531.
```

hh2110: but CI is now passing -- are we ok to merge?

callicles: Hum, no - the diagnosis is right, but the fix seems wrong. it looks like we take a version of the cli to pin it. But I don't think the version of the CLI is necessarily the latest in that case?

hh2110: @callicles  -- what do i need to do to get this merged please?

callicles: I believe we could do the following:
* I made a PR to trigger a new release so that things are not problematic. https://github.com/514-labs/moosestack/pull/4095 -> you should be able to approve it to merge.
* You revert the changes you have done on the test, which renders them moot I believe (there is no previous version anymore)

Then all should be green and we can merge

callicles: Actually - mine is breaking as well. I will do the proper fix for the test and. you can revert the fixes here and then you will just have to rebase on main

callicles: Done - you should be able to rebase with the revert and have everything green

